PR #56, opened 2018-02-26
Start Telemetry API definition

abrooksv:
* Expose a Metric API
* Add a Cognito Identity credential provider
* Add a batching metric publisher


gholt-aws:
closing in favor of a different approach...

PR #238, opened 2018-09-19
Filter codecov PR comment by flag

abrooksv:
Testing to see if we can filter unit test only

## License


kiiadi:
What we doing with this one?

abrooksv:
Never heard back from CodeCov, will revisit if I ever do

Issue #265, opened 2018-10-01
[0.5] Design: Update service/resource icons with new branding


nickkova4:
**Is your feature request related to a problem? Please describe.**
Yes, we need to update the Lambda icons to match new AWS branding. The icons marketing created for us for the overall app (e.g. Lambda) and it's child resources (e.g. apps and functions) need to communicate the proper hierarchical relationship. We also need to gauge the designs for the extension icons as they appear in the IDE panels. Finally we want the other icon instances covered [here](https://quip-amazon.com/ndV3ATiAUmiO) to be addressed. 

**Describe the solution you'd like**
New icons work well with the IDE in light and dark modes and on retina screens. 

**Describe alternatives you've considered**
We have considered having no icons and only using text.

**Additional context**
N/A

abrooksv:
Fixed overlooked New Lambda icon: #293

PR #276, opened 2018-10-04
Update to use new lambda icons

abrooksv:
## Types of changes

## Description
Update new lambda function icon from branding, drop the lambda service icon from explorer

## Related Issue(s)
#265 

screen shot 2018-10-04 at 3 54 06 pm
screen shot 2018-10-04 at 3 54 26 pm

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsExplorerNode.kt
```diff
@@ -49,7 +49,7 @@ class AwsExplorerRootNode(project: Project) :
     }
 }
 
-abstract class AwsExplorerServiceRootNode(project: Project, value: String, awsIcon: Icon) :
+abstract class AwsExplorerServiceRootNode(project: Project, value: String, awsIcon: Icon?) :
```
kiiadi: Was dropping the icon from the service something that UX suggested? I'm not sure this should be optional - we should either have icons on all our service nodes or none of them. Anything else is going to look funny IMO.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/ExplorerNodes.kt
```diff
@@ -19,7 +19,7 @@ import software.aws.toolkits.jetbrains.core.explorer.AwsTruncatedResultNode
 import software.aws.toolkits.resources.message
 
 class LambdaServiceNode(project: Project) :
-    AwsExplorerServiceRootNode(project, message("lambda.service_name"), AwsIcons.Logos.LAMBDA) {
+    AwsExplorerServiceRootNode(project, message("lambda.service_name"), null) {
```
kiiadi: If we are going to keep this interface we might as well give `awsIcon` a default value of `null` rather than having to pass it explicitly.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsExplorerNode.kt
abrooksv: Ya, that came from branding and Nick. To be honest this was a quick job since Nick was sitting next to me :)

This should not be optional, it should be coded to always pass null to parent

abrooksv:
Updated to remove AwsIcon? from interface of a service node

Issue #271, opened 2018-10-04
Lambda handler resolver for Java does not work correctly when using RequestHandler interface


abrooksv:
**To reproduce**
Given:
```
package helloworld;

import com.amazonaws.services.lambda.runtime.Context;
import com.amazonaws.services.lambda.runtime.RequestHandler;

public class App implements RequestHandler<Object, Object> {

    public Object handleRequest(final Object input, final Context context) {
      return "Hello";
    }
}
```

The `Lambda#findPsiElementsForHandler` says `helloworld.App::handleRequest` is invalid, and requires it to be `helloworld.App` even though both are valid

MariusVolkhart:
This appears to still be an issue in `v1.0`.

Generating a new Hello World sample, when I go to run the function locally, the IDE says "Cannot find handler 'helloworld.App::handleRequest' in the project.

MariusVolkhart:
Nevermind, I forgot to package the project first

PR #296, opened 2018-10-16
Make JavaLambdaHandlerResolver more accurate

abrooksv:
## Types of changes

## Description
When using the Lambda interfaces (e.g. RequestHandler), you could not reference the handler by method name, only class. This was due to multiple methods got returned due to the hierarchy. This ended up exposing gaps related to Lambda overloading resolution behavior. This makes this behavior match the docs for all use cases except undefined behavior.

## Motivation and Context
Bug fix

## Related Issue(s)

#271

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/JavaExtensions.kt
```diff
@@ -189,14 +189,49 @@ class JavaLambdaHandlerResolver : LambdaHandlerResolver {
...
+                var weight = it.parameterList.parametersCount
+                if (it.parameterList.parameters.getOrNull(weight - 1)?.isContextParameter() == true) {
+                    weight++
+                }
+                maxWeight = maxOf(maxWeight, weight)
+                weight
+            }.getOrDefault(maxWeight, emptyList())
+
+        // Empty, or undefined behavior
+        if (methods.isEmpty() || methods.size > 1) {
```
kiiadi: So in the case of "undefined behaviour" we blow up?

abrooksv: We return nothing, so we wont "find" it for Go To Source, and will be deemed invalid in the run configs

Issue #307, opened 2018-10-17
Add tests for JavaLambdaHandlerResolver#determineHandler


abrooksv:
**Describe the bug**
We do not handle all the cases going from PsiElement -> Handler, like we do going from Handler -> PsiElement

**To reproduce**
Given 
```java
package com.example;

import com.amazonaws.services.lambda.runtime.Context;

public abstract class AbstractHandler {
    public void handleRequest(InputStream input, OutputStream output, Context context) { }

    protected abstract void internalHandle();
}

package com.example;

import com.amazonaws.services.lambda.runtime.Context;
import java.io.InputStream;
import java.io.OutputStream;

public class ConcreteHandler extends AbstractHandler {
    @Override
    protected  void internalHandle() { }
}
```

**Expected behavior**
`com.exmaple.ConcreteHandler::handleRequest` should be valid.

PR #319, opened 2018-10-19
Support inheritance for finding valid handlers in Java.

zhangzhx:
Fix issue #307

## Types of changes

## Description

Given code
```java
public abstract class AbstractClass implements RequestHandler<String, String> {

    @Override
    public String handleRequest(String request, Context context) {
        return internalMethod(request, context);
    }

    protected abstract String internalMethod(String request, Context context);

    public String customHandleRequest(String request, Context context) {
        return internalMethod(request, context);
    }
}

public class ConcreteClass extends AbstractClass {

    @Override
    protected String internalMethod(String request, Context context) {
        return "Hello " + request;
    }

    public String concreteClassHandler(String request, Context context) {
        return internalMethod(request, context);
    }
}

```
The handler auto-completion:
inheritedhandlersautocompletion

## Related Issue(s)
#307 

## Checklist

- [x ] I have read the **README** document
- [ x] I have read the **CONTRIBUTING** document
- [ x] Local run of `gradlew check` succeeds
- [ x] My code follows the code style of this project
- [ x] I have added tests to cover my changes
- [ x] All new and existing tests passed

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaHandlerResolver.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaHandlerResolver.kt
@@ -37,5 +37,14 @@ interface LambdaHandlerResolver {
      */
     fun determineHandler(element: PsiElement): String?
 
+    /**
+     * Return a set of valid handlers from the given [PsiElement]. Different from [determineHandler], it returns all the valid
+     * handler strings contained by the [element].
+     *
+     * For Java implementation, if the element is a Java class, it returns all the valid methods either defined by this class or its
+     * super class(es).
+     */
+    fun determineHandlers(element: PsiElement): Set<String>
+
     companion object : RuntimeGroupExtensionPointObject<LambdaHandlerResolver>(ExtensionPointName.create("aws.toolkit.lambda.handlerResolver"))
 }
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaHandlerResolver.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaHandlerResolver.kt
@@ -3,7 +3,6 @@
 
 package software.aws.toolkits.jetbrains.services.lambda.java
 
-import com.intellij.execution.JavaExecutionUtil
 import com.intellij.lang.jvm.JvmModifier
 import com.intellij.openapi.project.Project
 import com.intellij.psi.JavaPsiFacade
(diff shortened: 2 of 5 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaHandlerResolver.kt
```diff
@@ -37,5 +37,14 @@ interface LambdaHandlerResolver {
...
     fun determineHandler(element: PsiElement): String?
 
+    /**
+     * Return a set of valid handlers from the given [PsiElement]. Different from [determineHandler], it returns all the valid
+     * handler strings contained by the [element].
+     *
+     * For Java implementation, if the element is a Java class, it returns all the valid methods either defined by this class or its
+     * super class(es).
+     */
+    fun determineHandlers(element: PsiElement): Set<String>
```
kiiadi: I don't understand how a single `PsiElement` could have more than one handler associated with it. We do the determination based on either a "Class" or a "method" element (or an identifier with a class or method as it's parent) a class's handler will always be the fully-qualified class name and a method will be the fully-qualified class name with "::[methodName]" appended.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaHandlerResolver.kt
```diff
@@ -107,6 +112,25 @@ class JavaLambdaHandlerResolver : LambdaHandlerResolver {
...
+        if (!clz.canBeInstantiatedByLambda()) {
+            return emptySet()
+        }
+
+        val handlers = mutableSetOf<String>()
+        if (clz.implementsLambdaHandlerInterface()) {
+            clz.qualifiedName?.let { handlers.add(it) }
+        }
+
+        handlers.addAll(clz.allMethods
```
kiiadi: Maybe I've mis-understood but as above, I think a class-level (ie. something that implements either `RequestHandler` or `StreamingRequestHandler` - at any-point in its hierarchy) should just be the full-qualified class name; the method-name does not need to be appended to the handler.

Right?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/LambdaHandlerIndexTest.kt
```diff
@@ -74,7 +74,152 @@ class LambdaHandlerIndexTest {
...
+                    return request.toUpperCase();
+                }
+            }
+            """
+        )
+
+        runInEdtAndWait {
+            assertThat(LambdaHandlerIndex.listHandlers(projectRule.project))
+                    .contains("com.example.ConcreteClass")
+                    .contains("com.example.ConcreteClass::customHandleRequest")
```
kiiadi: Agree this was missing - that's the real problem right?

zhangzhx: Yeah, this CR is to solve this missing part.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaHandlerResolver.kt
zhangzhx: Correct. We remove the special method `handleRequest` if the class implements Lambda handler interfaces in `isValidHandler`.

See test case `testHandlerIsReturned_customMethodInRequestHandlerImplementation`

PR #320, opened 2018-10-22
node js support prototype

denofevil:
**This is a prototype, do not merge as is**
This change introduces a prototype of NodeJS lambda run and debug support.

## Types of changes

## Description
It allows you to create a new run configuration for a NodeJS lambda (created via `sam init --runtime nodejs`) and debug it locally.
In order to try it do the following:
1. Create a new project via `sam init --runtime nodejs`
2. Open it in IntelliJ
3. Mark `hello_world` directory as a source root
After that, you should get an icon on the gutter to run or debug the lambda.

There are 3 major improvements that are required to merge this:
1. NodeJS lambda packager needs to detect lambda root properly. Right now it requires a manual setting of source root to work.
2. API for getting a runtime should be refined. Right now plugin uses SDK concept to select runtime, and NodeJS plugin doesn't use SDK. Without that, new configurations are created with empty runtime field.
3. The debugger doesn't use project mappings, breakpoints work but when hit a remote version is opened.

## Checklist

## License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,7 +170,7 @@ apply plugin: 'org.jetbrains.intellij'
 apply plugin: ChangeLogPlugin
 
 intellij {
-    version "IC-" + ideaVersion
+    version "IU-" + ideaVersion
     pluginName 'aws-jetbrains-toolkit'
     updateSinceUntilBuild false
     downloadSources = System.getenv("CI") == null
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSRuntimeGroup.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSRuntimeGroup.kt
@@ -1,0 +1,27 @@
+package software.aws.toolkits.jetbrains.services.lambda.nodejs
+
+import com.intellij.lang.javascript.JavaScriptSupportLoader
+import com.intellij.lang.javascript.JavascriptLanguage
+import com.intellij.openapi.projectRoots.Sdk
+import software.amazon.awssdk.services.lambda.model.Runtime
+import software.aws.toolkits.jetbrains.services.lambda.RuntimeGroupInformation
+
+class NodeJSRuntimeGroup : RuntimeGroupInformation {
+    override val runtimes: Set<Runtime> = setOf(
+            Runtime.NODEJS,
+            Runtime.NODEJS8_10,
+            Runtime.NODEJS6_10,
+            Runtime.NODEJS4_3,
+            Runtime.NODEJS4_3_EDGE
+    )
+
+    override val languageIds: Set<String> = setOf(
+            JavascriptLanguage.INSTANCE.id,
+            JavaScriptSupportLoader.TYPESCRIPT.id,
+            JavaScriptSupportLoader.FLOW_JS.id,
+            JavaScriptSupportLoader.ECMA_SCRIPT_6.id
+    )
+
+    // TODO we need module here, NodeJS doesn't use Sdk concept
+    override fun runtimeForSdk(sdk: Sdk): Runtime? = Runtime.NODEJS
+}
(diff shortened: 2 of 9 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSRuntimeGroup.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    override val languageIds: Set<String> = setOf(
+            JavascriptLanguage.INSTANCE.id,
+            JavaScriptSupportLoader.TYPESCRIPT.id,
+            JavaScriptSupportLoader.FLOW_JS.id,
+            JavaScriptSupportLoader.ECMA_SCRIPT_6.id
+    )
+
+    // TODO we need module here, NodeJS doesn't use Sdk concept
+    override fun runtimeForSdk(sdk: Sdk): Runtime? = Runtime.NODEJS
```
abrooksv: Should be safe to change to `runtimeForSdk(module: Module, sdk: Sdk)`, or is the SDK null completely in WebStorm?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSSamDebugSupport.kt
```diff
@@ -0,0 +1,52 @@
...
+import software.aws.toolkits.jetbrains.services.lambda.execution.sam.SamRunningState
+import java.net.InetSocketAddress
+
+class NodeJSSamDebugSupport : SamDebugSupport {
+    override fun createDebugProcess(environment: ExecutionEnvironment, state: SamRunningState, debugPort: Int): XDebugProcessStarter? {
+        return object : XDebugProcessStarter() {
+            override fun start(session: XDebugSession): XDebugProcess {
+                val executionResult = state.execute(environment.executor, environment.runner)
+                // TODO proper file finder is needed for the debugger to open proper local files
+                val fileFinder = NodeJSFileFinder(session.project)
```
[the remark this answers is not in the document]
denofevil: Yes, it should work. Most likely providing `com.intellij.javascript.debugger.RemoteDebuggingFileFinder` with proper mappings would be enough

build.gradle
```diff
@@ -170,7 +170,7 @@ apply plugin: 'org.jetbrains.intellij'
 apply plugin: ChangeLogPlugin
 
 intellij {
-    version "IC-" + ideaVersion
+    version "IU-" + ideaVersion
```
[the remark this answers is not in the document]
denofevil: Yes, that makes sense as JS and Go are available only in paid products

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSRuntimeGroup.kt
denofevil: SDK is null in WebStorm. In other IDEs SDK may be not null, but it's not required for NodeJS plugin to function.

kiiadi: Why don't we remove `runtimeForSdk` and replace it with `runtimeForModule(Module)` - those languages that do have a module and want the SDK to drive that can just use the `ModuleRootManager`. We may also need to introduce a `runtimeForProject(Project)` method.

There's two usages of the current `runtimeForSdk`:
* [`SamRunSettingsEditor`](https://github.com/aws/aws-toolkit-jetbrains/blob/9d98f62b9cf655b0fdd417666454d8eebe9911b9/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamRunConfiguration.kt#L234) this only has `project` in scope, the `RunConfiguration` is not (currently) tied to a module, but we can probably infer it from the handler element.
* [`LambdaSamRunConfigurationProducer`](https://github.com/aws/aws-toolkit-jetbrains/blob/9d98f62b9cf655b0fdd417666454d8eebe9911b9/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/LambdaSamRunConfigurationProducer.kt#L60) this *does* have module in scope, so we can replace its usage.

kiiadi:
@denofevil did you want to take a pass at updating the PR or do you want us to take ownership of it from here?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJSRuntimeGroup.kt
denofevil: Yep, that would work

denofevil:
@kiiadi it would be awesome if you could take ownership from here

Issue #322, opened 2018-10-22
Run local from SAM template odd Runtime dropdown behaviour


kiiadi:
**Describe the bug**

When creating (or loading a saved) Runtime Configuration that pulls details from a template, the `Runtime` drop-down should be disabled. Disabling only happens when 'function' is (re)selected - however it should be disabled on initial load.

**To reproduce**
![2018-10-22_13-12-33](https://user-images.githubusercontent.com/4661536/47316704-97cbf080-d5fc-11e8-94d3-d0efcc8f65ac.gif)

**Expected behavior**
Runtime dropdown should be disabled on initial load whenever a configuration uses a template.

PR #347, opened 2018-10-26
Fix the bug that runtime dropdown box is still enabled when using tem…

zhangzhx:
…plate for lambda function.

## Types of changes

## Description

The `populateValues` method enables the `ComboBox` when the target element array is not empty. This is not intended in this case. The population happens after we explicitly disable the `ComboBox` as the action is in SAM template.

This also affect the `iamRole` and `bucket source` dropdown box. As these values are required anyway, enabling these two widgets by default.

## Related Issue(s)

#322 

## Testing

fixbug

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaPanel.form
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaPanel.form
@@ -107,7 +107,7 @@
           <grid row="6" column="1" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="1" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties>
-          <enabled value="true"/>
+          <enabled value="false"/>
         </properties>
       </component>
       <component id="2ff11" class="com.intellij.openapi.ui.ComboBox" binding="runtime">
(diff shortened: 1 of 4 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -29,7 +29,6 @@ fun <T> ComboBox<T>.populateValues(selected: T? = null, block: () -> List<T>) {
             model.removeAllElements()
             values.forEach { model.addElement(it) }
             this.selectedItem = selected
-            this.isEnabled = values.isNotEmpty()
```
abrooksv: This is used in several places, is this going to affect other spots negatively?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaPanel.form
```diff
@@ -121,7 +121,7 @@
           <grid row="7" column="1" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="1" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties>
-          <enabled value="false"/>
+          <enabled value="true"/>
```
abrooksv: They are supposed to be disabled due to the values are loaded in from the network, so they should not be enabled till it loads.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
zhangzhx: This affect iamRole and bucket selection in Lambda creation panel. Reworked the way for fixing the bug to avoid affecting the current behavior.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaPanel.form
zhangzhx: Reverted this change.

zhangzhx:
This commit also fixes the bug when no bucket/iamRole in your AWS account/region, creating one still disables the widget.

Issue #273, opened 2018-10-04
[0.5] Create bucket dialog does not report errors


abrooksv:
**Describe the bug**

**To reproduce**
1. Open the Create Lambda -> Create Bucket dialog
2. Enter a bucket that already exists
3. No error, until all dialogs are closed

**Expected behavior**
Dialog should fail to close with the error

```
software.amazon.awssdk.services.s3.model.BucketAlreadyExistsException: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again. (Service: S3Client, Status Code: 409, Request ID: 677985FFF9024A73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleErrorResponse(HandleResponseStage.java:116)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleResponse(HandleResponseStage.java:74)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:59)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:41)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:205)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage$RetryExecutor.doExecute(RetryableStage.java:139)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage$RetryExecutor.execute(RetryableStage.java:109)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:67)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:47)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:205)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:56)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:41)
[...]

PR #365, opened 2018-11-02
Add create S3 Bucket panel.

zhangzhx:
Fix the S3 bucket creation dialog doesn't show error message

## Types of changes

## Related Issue(s)
#273 

s3bucketcreation

## Checklist

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialogTest.kt
```diff
@@ -0,0 +1,98 @@
...
+class CreateS3BucketDialogTest {
+    @Rule
+    @JvmField
+    val projectRule = ProjectRule()
+
+    private val s3Mock = mock<S3Client>(defaultAnswer = DelegateSdkConsumers())
+
+    @Before
+    fun setUp() {
+        reset(s3Mock)
```
richa-li: Not sure if this is required, I thought Mockito/JUnit reset the class class for each test method

zhangzhx: Nice catch. Removed.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private fun bucketName(): String = view.bucketName.text.trim()
+
+    private fun createBucket() {
+        s3Client.createBucket { request -> request.bucket(bucketName()) }
+        bucket = bucketName()
+    }
+
+    @TestOnly
+    fun validateBucketName(): String? = if (bucketName().isEmpty()) message("s3.create.bucket.missing.bucket.name") else null
```
kiiadi: Should we do more validation - s3 publishes a pretty good set of things that it'll accept.

ps - why expose this method? can't we just call `doValidate` directly?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+
+class CreateS3BucketDialog(
+    private val project: Project,
+    private val s3Client: S3Client,
+    private val parent: Component? = null
+) : DialogWrapper(project, parent, false, IdeModalityType.PROJECT) {
+
+    private val view = CreateBucketPanel(project)
+
+    var bucket: String? = null
```
kiiadi: Why do we need the `bucket` instance variable, the `bucketName()` method and the access from the testfield ?

Reckon we can just fetch it straight from the text field.... no need to have two mutable states.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialogTest.kt
kiiadi: The reset will only happen if you're using the `@RunWith(JUnitMockitoRunner)` which we ain't.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialog.kt
zhangzhx: S3 bucket naming has many restrictions and they are changing overtime. https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

The SDKs don't do client side validations and I have the same concerns. Ie. if S3 removes some restriction, the old version of the plugin won't allow customer to create bucket.

zhangzhx: `doValidate` is a `protected` method.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialog.kt
zhangzhx: I was thinking that `bucketName` always returns whatever is in the text field whether is validated or not. `bucket` returns the validated value which is safe to use outside of the Panel.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateBucketPanel.java
```diff
@@ -0,0 +1,20 @@
...
+import javax.swing.JTextField;
+
+public class CreateBucketPanel {
+    private final Project project;
+
+    JTextField bucketName;
+    JPanel component;
+
+    public CreateBucketPanel(Project project) {
+        this.project = project;
```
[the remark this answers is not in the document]
zhangzhx: Removed this.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateBucketPanel.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.intellij.openapi.project.Project;
+
+import javax.swing.JPanel;
+import javax.swing.JTextField;
+
+public class CreateBucketPanel {
+    private final Project project;
+
+    JTextField bucketName;
```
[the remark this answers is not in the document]
zhangzhx: Added.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/CreateS3BucketDialog.kt
zhangzhx: Removed the `bucket` field as it doesn't gain too much. Using `bucketName()` instead

PR #371, opened 2018-11-03
Adding memory to the edit/upload lambda dialog.

### LINKED ISSUE #350: Should be able to set Lambda memory in Edit Lambda UI
Memory is an important setting and we should expose it in the Create/Edit Lambda UI

kiiadi:
Fixes #350

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Description

Adds the ability to specify and update memory settings on a lambda.

## Motivation and Context

Memory is an important setting

## Related Issue(s)

#350

## Testing

Updated tests accordingly.

screen shot 2018-11-02 at 5 34 10 pm

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditLambdaDialog.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditLambdaDialog.kt
@@ -40,6 +40,10 @@ import java.util.function.Function
 import javax.swing.Action
 import javax.swing.JComponent
 
+private const val MIN_MEMORY = 128
+private const val MAX_MEMORY = 3008
+private const val MEMORY_INCREMENT = 64
+private const val DEFAULT_MEMORY = 128
 private val DEFAULT_TIMEOUT = TimeUnit.MINUTES.toSeconds(1).toInt()
 private val MAX_TIMEOUT = TimeUnit.MINUTES.toSeconds(15).toInt()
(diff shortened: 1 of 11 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditLambdaDialog.kt
```diff
@@ -280,6 +288,13 @@ class UploadToLambdaValidator {
                 )
             }
         }
+        view.memorySize.text.toIntOrNull().let {
+            if (it == null || it < MIN_MEMORY || it > MAX_MEMORY || it.rem(MEMORY_INCREMENT) != 0) {
```
zhangzhx: Is it too restrict for the values to be 64ish? It seems Lambda accepts non 64ish values. I guess the actual memory is the upper bound of the specified value.

zhangzhx: Lambda might be using the 64mb under the hood, but I've manually tested to provide a value not 64ish using AWS CLI and Lambda doesn't error out. So my point here is that it would be a bad user experience to restrict the memory while it is allowed to use non-64ish size. I don't think customers want to do the math all the time.

abrooksv: +1 Slider if we can make it look nice

Issue #401, opened 2018-11-06
[2/5] Delete stack error handling 


zhangzhx:
A stack could be failed to delete. Handle the errors properly of deleting the stack.

Issue #400, opened 2018-11-06
[0.5/1] Show progress for deleting stack


zhangzhx:
Deleting stack is an async call. Show progress for the current status of the stack deletion.

PR #413, opened 2018-11-08
Support resource delete status report.

zhangzhx:
## Types of changes

## Description

When deleting resources, use callback to handle the final status of the deletion when it is async operation.

## Related Issue(s)

#400 #401 

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/DeleteResourceAction.kt
```diff
@@ -34,11 +35,18 @@ abstract class DeleteResourceAction<in T : AwsExplorerResourceNode<*>>(text: Str
 
             if (response != null) {
                 ApplicationManager.getApplication().executeOnPooledThread {
+                    val future = CompletableFuture<T>()
                     try {
                         performDelete(selected)
-                        notifyInfo(message("delete_resource.deleted", selected.resourceType, resourceName))
+                        future.complete(selected)
```
abrooksv: Why use the future, if you are always on the same thread anyway?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/applications/ServerlessExplorerNodes.kt
```diff
@@ -47,5 +51,42 @@ class DeleteApplicationAction : DeleteResourceAction<ServerlessApplicationNode>(
...
+                }
+
+                val targetStack = stacks[0]
+                when (targetStack.stackStatus()) {
+                    in deletionSuccessStatuses -> return
+                    in deletionFailedStatuses -> throw RuntimeException("Failed to delete stack $stackName in status ${targetStack.stackStatus()}")
+                    else -> Thread.sleep(delay * 1000L)
+                }
+            }
+            throw TimeoutException("Failed to delete stack $stackName in ${maxAttempts * delay}s. Please go to AWS Console for the latest status of the stack!")
```
kiiadi: We should move this to a `message`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/applications/ServerlessExplorerNodes.kt
```diff
@@ -47,5 +51,42 @@ class DeleteApplicationAction : DeleteResourceAction<ServerlessApplicationNode>(
...
+
+    companion object {
+
+        // TODO Use Java SDK waiter once supported
+        fun waitForStackDeletionComplete(client: CloudFormationClient, stackName: String) {
+            waitForStackDeletionComplete(client, stackName, 120, 30)
+        }
+
+        @TestOnly
+        fun waitForStackDeletionComplete(client: CloudFormationClient, stackName: String, maxAttempts: Int, delay: Int) {
```
kiiadi: This logic should probably be moved into a separate class so that it can be unit tested independently - then we wouldn't need this `@TestOnly` tag.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/DeleteResourceAction.kt
kiiadi: definitely don't see the point of the future here

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/applications/DeleteApplicationActionTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import software.amazon.awssdk.services.cloudformation.model.CloudFormationException
+import software.amazon.awssdk.services.cloudformation.model.DescribeStacksRequest
+import software.amazon.awssdk.services.cloudformation.model.DescribeStacksResponse
+import software.amazon.awssdk.services.cloudformation.model.Stack
+import software.amazon.awssdk.services.cloudformation.model.StackStatus
+import software.aws.toolkits.jetbrains.core.MockClientManagerRule
+import software.aws.toolkits.jetbrains.utils.delegateMock
+import java.util.concurrent.TimeoutException
+
+class DeleteApplicationActionTest {
```
kiiadi: How are we testing the 'waiter' makes multiple calls ?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/DeleteResourceAction.kt
zhangzhx: Reverted to not use Future.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/applications/ServerlessExplorerNodes.kt
zhangzhx: Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/applications/ServerlessExplorerNodes.kt
zhangzhx: I thought we are gonna remove this when Java SDK support waiter. But it's cleaner to move this away. Moved to the Cloudformation.kt file.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/applications/DeleteApplicationActionTest.kt
zhangzhx: Reworked the testing. Adding call times verification.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        val targetStack = stacks[0]
+        when (targetStack.stackStatus()) {
+            in deletionSuccessStatuses -> return
+            in deletionFailedStatuses -> throw RuntimeException(message("cloudformation.delete_stack.failed", stackName, targetStack.stackStatus().toString()))
+            else -> Thread.sleep(delay * 1000L)
+        }
+    }
+    throw TimeoutException(message("cloudformation.delete_stack.timeout", stackName, delay * maxAttempts))
```
kiiadi: Can we make this a `RuntimeException` - `TimeoutException` is checked I believe.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// TODO Use Java SDK waiter once supported
+fun CloudFormationClient.waitForStackDeletionComplete(stackName: String) {
```
kiiadi: Why do we need this? Why can't we just use default parameters on the method above? eg.

```
CloudFormationClient.waitForStackDeletionComplete(stackName: String, maxAttempts: Int = 120, delay: Duration = Duration.ofSeconds(20))
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
```diff
@@ -0,0 +1,43 @@
...
+package software.aws.toolkits.jetbrains.services.cloudformation
+
+import software.amazon.awssdk.services.cloudformation.CloudFormationClient
+import software.amazon.awssdk.services.cloudformation.model.CloudFormationException
+import software.amazon.awssdk.services.cloudformation.model.DescribeStacksRequest
+import software.amazon.awssdk.services.cloudformation.model.StackStatus
+import software.aws.toolkits.resources.message
+import java.util.concurrent.TimeoutException
+
+fun CloudFormationClient.waitForStackDeletionComplete(stackName: String, maxAttempts: Int, delay: Int) {
```
kiiadi: delay should be a `Duration` not an int

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
```diff
@@ -0,0 +1,43 @@
...
+                return
+            }
+            throw e
+        }
+        if (stacks.size != 1) {
+            throw RuntimeException(message("cloudformation.delete_stack.multiple_stacks_with_same_name"))
+        }
+
+        val targetStack = stacks[0]
+        when (targetStack.stackStatus()) {
```
kiiadi: We probably want this whole thing to be a more generic "waiter" function - there's going to be other places where we need to repeat an SDK call until a terminal state is reached. Something like:

```
fun <T> waiter(call: () -> T, successPredicate: (T) -> Boolean, failurePredicate: (T) -> Boolean, attempts: Int, delay: Duration)
```

This would handle all of your `repeat` logic and would match on the state using the predicates.

Take a look at the waiter in the SDK's test utils - `https://github.com/aws/aws-sdk-java-v2/blob/master/test/test-utils/src/main/java/software/amazon/awssdk/testutils/Waiter.java`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: It's for the test. In the test case, I don't want to wait for so long for time out.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: Kinda borrowed the Int from Java SDK. Our current waiter is configuring the delay as an int in second. Do we really need the delay in nano level?

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/waiters/FixedDelayStrategy.java#L45

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: Good point. That's exactly what I'm thinking now as I'm working on the deploy SAM. It has a lot status change handling. Will rework this for generic waiters.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
kiiadi: Java 6 doesn't have the the `Duration` object, it was only introduced in Java 8 - if we're representing a duration (we are) then we should make that obvious through the type system.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
kiiadi: what's your point? we can still collapse them into one method using Kotlin default arguments.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: Done. I created two new types of exceptions extending `RuntimeException` for indicating the type of the exception thrown by waiter. Used that instead.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: I see your point. Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/Cloudformation.kt
zhangzhx: Done.

core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -0,0 +1,42 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.core.utils
+
+import software.amazon.awssdk.awscore.exception.AwsServiceException
+import java.time.Duration
+
+fun <T> wait(
```
kiiadi: Should prolly comment each of these params since this is quite a big function signature

core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -0,0 +1,42 @@
...
+    call: () -> T,
+    success: (T) -> Boolean,
+    fail: (T) -> Boolean,
```
kiiadi: can we combine `fail` / `errorMessage` and `failByException` / `errorMessageByException` into single parameters that accept `T`/`Exception` and return `String?` - where the return type represents the error message (`null` is considered not to be a an error).

core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -0,0 +1,42 @@
...
+                successByException(e) -> return
+                failByException(e) -> throw WaiterUnrecoverableException(errorMessageByException(e))
+                else -> throw e
+            }
+        }
+    }
+    throw WaiterTimeOutException(timeoutErrorMessage)
+}
+
+class WaiterTimeOutException(message: String) : RuntimeException(message)
```
kiiadi: nit: `Timeout` not `TimeOut`

core/src/software/aws/toolkits/core/utils/Waiter.kt
zhangzhx: Done. It's always good to shrink the parameter size. :)

core/src/software/aws/toolkits/core/utils/Waiter.kt
zhangzhx: Done.

Issue #456, opened 2018-11-14
Only prevent deployment of Sam Functions as standalone Lambdas if Sam Template is in same module


awschristou:
#417 leveraged `CloudFormationTemplateIndex`, which contains sam template for any/all modules in a project. As a result, Sam Functions referenced by a template in any of the project's modules are not allowed to be deployed as standalone Lambda functions.

This should be refined so that the deploy is prevented only if a function is referenced by a Sam template in the same module.

Issue #417, opened 2018-11-09
[0.5] Disable Create new lambda function action in "applications"


kiiadi:
if function exists in a template definition it should not be possible to deploy it as naked lambda
* prevent deploying handlers to standalone lambda functions from the gutter icons

PR #457, opened 2018-11-14
Lambda Functions referenced by a Sam Template can no longer be deploy…

awschristou:
## Types of changes

## Description

The gutter icon contains an action to deploy standalone lambda functions. This action has been removed for functions that are referenced by a Sam Template, since they are considered a part of a Sam Application.

Known Caveat:
* Applies to Lambda function handlers referenced by all Sam Templates within a project, not just those in the respective module
* #456 will deal with this

## Related Issue(s)

#417

## Testing

Opened a Java project in IntelliJ, Python project in PyCharm.
See unit tests for handler/template combinations tried in-IDE

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaHandlerResolver.kt
```diff
@@ -67,5 +68,11 @@ interface LambdaHandlerResolver {
      */
     fun shouldShowLineMarker(handler: String): Boolean = false
 
+    /**
+     * Given a handler and the element it originated from, produce the appropriate Action that can
+     * be used to create a Lambda Function.
+     */
+    fun generateCreateLambdaFunctionAction(handlerName: String, element: PsiElement): CreateLambdaFunction
```
kiiadi: I'm not sure this belongs here, the create action should be the same for all languages right? It's only the packaging that differs.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,80 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+class CreateLambdaFunctionFromJavaClass(private val handlerName: String, private val project: Project)
+    : CreateElementBasedLambdaFunction(handlerName, project) {
+
+    init {
+        handlerSearch = "$handlerName::"
+        isExactMatch = false
+    }
+}
+
+class CreateLambdaFunctionFromJavaMethod(private val handlerName: String, private val project: Project)
```
kiiadi: We shouldn't have any language specific implementations in the core - the only place they should be brought is via explicit extension points.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,80 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+        val dialog = getEditFunctionDialog(project, runtime)
 
         dialog.show()
     }
+
+    protected open fun getEditFunctionDialog(project: Project, runtime: Runtime?): EditFunctionDialog =
+        EditFunctionDialog(project = project, mode = NEW, runtime = runtime)
+}
+
+open class CreateElementBasedLambdaFunction(private val handlerName: String, private val project: Project)
```
kiiadi: I don't understand what these are trying to achieve. How does this actually suppress the action from being visible?

awschristou:
@kiiadi when the gutter icon represents a Java class instead of a method, the handler is (for example) `helloworld.App`, which requires a `startsWith` match instead of an equality match against indexed handlers.  The template could contain `helloworld.App::handlerX`, `helloworld.App::handlerY`, etc, but wouldn't have an exact match of `helloworld.App`

If it wasn't for that, we could simplify to something like your sample code.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: I should have renamed the classes to something more general.

What if I expose the difference in functionality (exact vs startswith) with a class member, and collapse the classes back down?

I do like the idea of separating the most basic `CreateLambdaFunction` used by `plugin.xml` from the rest, to avoid working with nullable types.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: The change doesn't affect visibility on `CreateLambdaFunction` , but makes it so that `plugin.xml` is the only thing directly referencing that class (which doesn't have visibility requirements at this time). The derived class has visibility suppression, and the gutter icon code works with the derived class(es)

abrooksv:
Move the creation of the handler string into the action, then use `determineHandlers()` and check each of those against the `CloudFormationTemplateIndex` instead then

kiiadi:
`helloworld.App::handlerX` and `helloworld.App` would be different definitions.

I'm not actually that worried if we don't suppress on exact match....

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,53 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+        val dialog = getEditFunctionDialog(project, runtime)
 
         dialog.show()
     }
+
+    protected open fun getEditFunctionDialog(project: Project, runtime: Runtime?): EditFunctionDialog =
+        EditFunctionDialog(project = project, mode = NEW, runtime = runtime)
+}
+
+class ConditionalCreateLambdaFunction(
```
kiiadi: I don't understand the benefit of this additional class? I'd much rather just roll this logic into `CreateLambdaFunction`, if a handler is specified, we validate that it's not a member of the template as part of the update, if there's no handler (e.g it was created from the `plugin.xml`) then we skip that logic.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,53 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+    private val handlerName: String,
+    private val element: PsiElement,
+    private val lambdaHandlerResolver: LambdaHandlerResolver
+) : CreateLambdaFunction() {
+    override fun update(e: AnActionEvent?) {
+        super.update(e)
+
+        val templateFunctionHandlers = CloudFormationTemplateIndex.listFunctions(element.project)
+            .mapNotNull { it.handler() }
+            .distinct()
```
kiiadi: if we make this a `Set` instead of a list then we can do a much more efficient 'contains' below.

```
val templateFunctionHandlers = CloudFormationTemplateIndex.listFunctions(element.project)
            .mapNotNull { it.handler() }
            .toSet()

val allowAction = lambdaHandlerResolver.determineHandlers(element, element.containingFile.virtualFile).none { it in templateFunctionHandlers } 
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: I wasn't a fan of the additional null checks. I'm re-combining the classes.

kiiadi: `null` is a first-class thing in Kotlin - it's built-in to the language :P

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,58 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+    }
+
     override fun actionPerformed(event: AnActionEvent) {
         val project = event.getRequiredData(LangDataKeys.PROJECT)
         val runtime = event.runtime()
 
-        val dialog = if (handlerName != null) {
-            EditFunctionDialog(project = project, mode = NEW, runtime = runtime, handlerName = handlerName)
-        } else {
+        val dialog = if (handlerName == null) {
```
kiiadi: nit: (sorry @awschristou) this is a pointless change - all you've done is reverse the if-statement branches - it clutters up the diff!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,58 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
             EditFunctionDialog(project = project, mode = NEW, runtime = runtime)
+        } else {
+            EditFunctionDialog(project = project, mode = NEW, runtime = runtime, handlerName = handlerName)
         }
 
         dialog.show()
     }
+
+    override fun update(e: AnActionEvent?) {
```
kiiadi: 👍

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,58 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+import software.aws.toolkits.jetbrains.services.cloudformation.CloudFormationTemplateIndex
 import software.aws.toolkits.jetbrains.services.iam.IamRole
+import software.aws.toolkits.jetbrains.services.lambda.LambdaHandlerResolver
 import software.aws.toolkits.jetbrains.services.lambda.runtime
 import software.aws.toolkits.jetbrains.services.lambda.upload.EditFunctionMode.NEW
 import software.aws.toolkits.resources.message
 
-class CreateLambdaFunction(private val handlerName: String? = null) : AnAction(message("lambda.create_new"), null, AwsIcons.Actions.LAMBDA_FUNCTION_NEW) {
+class CreateLambdaFunction(
+    private val handlerName: String?,
```
kiiadi: I'd be surprised if this works from the `plugin.xml` since there's no defaults for these parameters.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,58 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
+            EditFunctionDialog(project = project, mode = NEW, runtime = runtime, handlerName = handlerName)
         }
 
         dialog.show()
     }
+
+    override fun update(e: AnActionEvent?) {
+        super.update(e)
+
+        if (handlerName == null || element == null || lambdaHandlerResolver == null) {
```
kiiadi: since `handlerName` being non-null forces the other 2 to be non-null also do you need to include them in the if statement?

Based on `init` it's not possible to construct this class with `handlerName` being non-null but `element` or `lambdaHandlerResolver` is null.

That said - I guess one thing that this does give you is the smart casting... so maybe forget me ?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -6,26 +6,58 @@ package software.aws.toolkits.jetbrains.services.lambda.upload
...
-class CreateLambdaFunction(private val handlerName: String? = null) : AnAction(message("lambda.create_new"), null, AwsIcons.Actions.LAMBDA_FUNCTION_NEW) {
+class CreateLambdaFunction(
+    private val handlerName: String?,
+    private val element: PsiElement?,
+    private val lambdaHandlerResolver: LambdaHandlerResolver?
+) : AnAction(message("lambda.create_new"), null, AwsIcons.Actions.LAMBDA_FUNCTION_NEW) {
+
+    init {
+        if (handlerName != null) {
+            element ?: throw Error("element must be provided if handlerName is provided")
```
kiiadi: Should almost never throw `Error` it's reserved for catastrophic VM failures like `OutOfMemory` - change these to be `IllegalArgumentException` - this is the typical exception used when the args passed to a function or constructor are not valid.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: chalk this up to all the churn in this PR, "reverting"

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: It does successfully bring up the dialog, but I am adding null defaults

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
awschristou: Smart casting and defensive programming by not making assumptions on the init validation.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
kiiadi: no no, don't do it if we don't need it - it must have smart logic to pass null to nullable fields.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
kiiadi: kk

PR #492, opened 2018-11-16
IntelliJ: Mark any template-specified CodeUri as a source folder

rli:
## Types of changes

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamCommon.kt
```diff
@@ -68,5 +69,15 @@ class SamCommon {
             }
             return codeUris
         }
+
+        fun setPythonSourceRoots(projectRoot: VirtualFile, project: Project, modifiableModel: ModifiableRootModel) {
```
abrooksv: should we call this setSourceRoots? since it should work for Java too

rli: possibly, but since the java templates come with a maven/gradle config we should be pushing users to use that, which would set the source/test roots for us

Issue #487, opened 2018-11-16
Save source bucket used for Stack/Lambda upload


kiiadi:
It's sill that I have to re-select the bucket every time I got to update a Lambda/Stack.

PR #514, opened 2018-11-17
Save sam deploy settings.

zhangzhx:
Save deploy settings so that the bucket/stack settings could be auto-filled for the second time.

## Types of changes

## Related Issue(s)

#487 

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/DeployServerlessApplicationAction.kt
```diff
@@ -34,10 +37,14 @@ class DeployServerlessApplicationAction : DumbAwareAction(
         val templateFile = getSamTemplateFile(e) ?: throw Exception("Could not detect template file")
         val template = CloudFormationTemplate.parse(project, templateFile)
 
-        val stackDialog = DeployServerlessApplicationDialog(project, template.parameters())
+        val module = ModuleUtil.findModuleForFile(templateFile, project)
```
kiiadi: Is module the right thing to tie this to? I feel like it should maybe be template file - though then we have a problem with tracking renames...

My concern with module is then we can only have one stack per module - what if they have more than one template file / stack per module?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationSettings.kt
```diff
@@ -0,0 +1,49 @@
...
+
+package software.aws.toolkits.jetbrains.services.lambda.deploy
+
+import com.intellij.openapi.components.PersistentStateComponent
+import com.intellij.openapi.components.State
+import com.intellij.openapi.module.Module
+import com.intellij.openapi.module.ModuleServiceManager
+
+@State(name = "deploySam")
+class DeployServerlessApplicationSettings : PersistentStateComponent<DeployServerlessApplicationConfiguration> {
```
kiiadi: I'd love to make this a bit more generic for being able to store these sorts of settings across IDE restarts - is there a way we could create a pattern that would allow us to easily store other things, for example : bucket name for lambda update.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
```diff
@@ -24,9 +28,14 @@ import javax.swing.JComponent
 
 class DeployServerlessApplicationDialog(
     private val project: Project,
+    private val templateFile: VirtualFile,
```
awschristou: If you are passing in `templateFile`, the `parameters` parameter is redundant, since it will never be sourced from somewhere else.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/DeployServerlessApplicationAction.kt
zhangzhx: Moved to file level and the key is the relative path of the template file to the root of the module.

Open question: currently, we only recognize template file with the hard coded name **template.yaml** or **template.yml** for deployment. In this case, there are no way to have multiple yaml files in the same folder. Do we need to loosen the restriction for sam template files?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationSettings.kt
zhangzhx: As discussed offline, the bucket setting could be shared through all deployment actions in the same module. I've renamed this class to *DeploySettings* to allow them share settings like s3 bucket.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -36,15 +36,20 @@ fun <T> ComboBox<T>.populateValues(
...
         ApplicationManager.getApplication().invokeLater({
             val model = this.model as DefaultComboBoxModel<T>
-            val previouslySelected = model.selectedItem
             model.removeAllElements()
             values.forEach { model.addElement(it) }
             this.selectedItem = if (forceSelectDefault || previouslySelected == null) default else previouslySelected
             if (updateStatus) {
                 this.isEnabled = values.isNotEmpty()
+            } else {
+                this.isEnabled = previousState
```
awschristou: Should this be in a finally block in case `block()` fails in a bad way?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
zhangzhx: Good catch. I removed the parameter `parameters` here to derive directly from `templateFile`

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
zhangzhx: This could be a bug for the error handling when loading resources. We need to create a separate issue for handling this. The correct behavior in this case should be showing a special item `Failed` and a validation failure should be returned. Created an issue #535 for tracking this.

PR #517, opened 2018-11-19
Added AWS X-Ray support. Option is hidden if region does not support X-Ray.

awschristou:
## Types of changes

## Description

Creating and Updating Lambda Functions can now set whether or not X-Ray is used.

image

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
@@ -10,6 +10,7 @@ import software.amazon.awssdk.services.lambda.model.CreateFunctionResponse
 import software.amazon.awssdk.services.lambda.model.FunctionConfiguration
 import software.amazon.awssdk.services.lambda.model.GetFunctionConfigurationResponse
 import software.amazon.awssdk.services.lambda.model.Runtime
+import software.amazon.awssdk.services.lambda.model.TracingMode
 import software.amazon.awssdk.services.lambda.model.UpdateFunctionConfigurationResponse
 import software.aws.toolkits.core.region.AwsRegion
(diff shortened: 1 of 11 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -31,6 +32,7 @@ data class LambdaFunction(
     val envVariables: Map<String, String>?,
     val timeout: Int,
     val memorySize: Int,
+    val enableXray: Boolean,
```
kiiadi: can we call this `xrayEnabled` rather than the verb version.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/CreateLambdaFunction.kt
```diff
@@ -69,5 +70,13 @@ data class FunctionUploadDetails(
...
     val envVars: Map<String, String>,
     val timeout: Int,
-    val memorySize: Int
-)
\ No newline at end of file
+    val memorySize: Int,
+    val enableXray: Boolean
+)
+
+fun FunctionUploadDetails.enableXrayAsTracingMode(): TracingMode =
```
kiiadi: A function that begins with a verb I would expect a mutation - we're just returning a state it should be a noun.

kiiadi:
Also - can we change the UI to include a label "X-Ray Enabled:" on the left hand side and then have the check-box on the right, lined up with the other inputs - with no label.

abrooksv:
@kiiadi https://jetbrains.github.io/ui/principles/layout/ indicates we do it with label on the right

kiiadi:
I stand-down. All hail the UI standards.

Issue #386, opened 2018-11-06
SAM Local Run Configuration environment variables substitution


kiiadi:
**Describe the solution you'd like**

Serverless Functions can define, as part of their properties, environment variables that refer to other resources in the template. For example : [template.yaml](https://github.com/aws-samples/cookiecutter-aws-sam-s3-rekognition-dynamodb-python/blob/b3bb84a51ba50efc84ebd4c4f065e7532c8fbf1a/%7B%7Bcookiecutter.project_name%7D%7D/template.yaml#L18) defines a `TABLE_NAME` environment variable. These should be automatically populated if the stack has been deployed and it's possible to determine the actual resource name.

In the above example `TABLE_NAME` refers to the `ResultsTable` resource that is also defined in the template. In order to determine the actual physical resource ID, we'll have to describe the CloudFormation stack, and determine the physical ID for the given `ResultsTable` logical ID, and then automatically create it as an environment variable.

![screen shot 2018-11-05 at 6 32 17 pm](https://user-images.githubusercontent.com/4661536/48039377-5b82af00-e129-11e8-9c15-18a83bb04acb.png)

borlander:
Normally an Environment Variables field (EnvironmentVariablesComponent) is used to allow the user to *edit* the environment. 

AFAIU, in your use case the values are either not available, or computed externally and are read-only for the user, so I am not sure that this component is the best fit for the use case. 

However, app servers integrations sometimes have the mix of read-only and writable env vars in the similar UI. Here is the screenshot of JBoss run-config, Startup/Connection tab | Debug, and JAVA_OPTS has a read-only value, precomputed by some (also much easier) rule: 

<img width="956" alt="jboss_java_opts" src="https://user-images.githubusercontent.com/1390992/48143721-b5ad7c80-e2af-11e8-97f4-5237bebf65ef.png">

kiiadi:
The intention wasn't to make the environment variables read-only. We definitely want them to be editable by the user; but if we can pre-populate some of the values (that the customer can then later edit) then that's useful.

borlander:
Some related docs:

Resources: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html

Describe stack in CLI: 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-describing-stacks.html

borlander:
Could you please clarify 
[...]

PR #542, opened 2018-11-20
SAM Local Run Configuration environment variables substitution

viatra2:
#386 - SAM Local Run Configuration environment variables substitution

## Types of changes

## Description
- Yaml model extended to support resource Environment Variable's 
- environment variables defined in a resource with the `Ref` to local name considered as "predefined" in the local SAM run config 
- - user can't delete them or change their names 
- - user can add any other variables which are not managed by the feature
- Value column of the environment table is changed to have combobox editor 
- - combo values are matching physical resource IDs 
- - combobox is soft [editable], user may override the value if needed
- - the full resource mapping is requested every time on run-config reopening, from all resources in all available stacks
- - if value is not set (or cleared) the next reopening will set the value to the first physical ID found 
- - the existing values are never cleared even if there are no matching physical resources

## Related Issue(s)

#386 

see screen cast in #386 

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
@@ -54,12 +54,22 @@ interface NamedMap {
 interface Resource : NamedMap {
     fun isType(requestedType: String): Boolean
     fun type(): String?
+
+    fun getEnvironmentVariables(): Sequence<EnvironmentVariable>
 }
 
 interface Parameter : NamedMap {
     fun defaultValue(): String?
     fun description(): String?
     fun constraintDescription(): String?
+}
+
+
+interface EnvironmentVariable {
+    val variableName: String
+    val variableValue: String
+
+    fun isReference(): Boolean
 }
 
 class CloudFormationParameter(private val delegate: NamedMap) : NamedMap by delegate, Parameter {
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
@@ -55,7 +55,7 @@ interface Resource : NamedMap {
     fun isType(requestedType: String): Boolean
     fun type(): String?
 
-    fun getEnvironmentVariables(): Sequence<EnvironmentVariable>
+    fun getEnvironmentVariables(): Sequence<Variable>
 }
 
 interface Parameter : NamedMap {
@@ -64,8 +64,7 @@ interface Parameter : NamedMap {
     fun constraintDescription(): String?
 }
 
-
-interface EnvironmentVariable {
+interface Variable {
     val variableName: String
     val variableValue: String
 
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
@@ -1,0 +1,86 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cloudformation
+
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.project.Project
+import com.intellij.util.containers.MultiMap
+import software.amazon.awssdk.services.cloudformation.CloudFormationClient
+import software.amazon.awssdk.services.cloudformation.model.*
+import software.aws.toolkits.jetbrains.core.AwsClientManager
+import java.util.*
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionStage
+import java.util.stream.Stream
+
+interface ResourceMapping {
+    fun listAllSummaries(logicalResource: String): Collection<StackResourceSummary>
+
+    fun guessPhysicalResourceId(logicalResource: String): String?
+
+    fun isEmpty(): Boolean
+
+    object EMPTY : ResourceMapping {
+        override fun listAllSummaries(logicalResource: String): Collection<StackResourceSummary> = Collections.emptyList()
+
+        override fun guessPhysicalResourceId(logicalResource: String): String? = null
+
+        override fun isEmpty(): Boolean = true
+    }
+}
+
+internal class ResourceMappingImpl : ResourceMapping {
+    private val logicalToSummary: MultiMap<String, StackResourceSummary> = MultiMap.createLinked()
+
+    override fun listAllSummaries(logicalResource: String): Collection<StackResourceSummary> {
+        return logicalToSummary.get(logicalResource)
+    }
+
+    override fun guessPhysicalResourceId(logicalResource: String): String? {
+        return listAllSummaries(logicalResource).firstOrNull()?.physicalResourceId()
+    }
+
+    private fun registerResource(resourceSummary: StackResourceSummary) {
+        resourceSummary.logicalResourceId()?.let { logicalToSummary.putValue(it, resourceSummary) }
+    }
+
+    override fun isEmpty(): Boolean = logicalToSummary.isEmpty
+
+    companion object {
+        @JvmStatic
+        fun promiseMapping(project: Project): CompletionStage<ResourceMapping> {
+            val client: CloudFormationClient = AwsClientManager.getInstance(project).getClient()
+            return promiseMapping(client)
+        }
+
+        @JvmStatic
+        fun promiseMapping(client: CloudFormationClient): CompletionStage<ResourceMapping> {
+            val promise = CompletableFuture<ResourceMapping>()
+
+            ApplicationManager.getApplication().executeOnPooledThread {
+                val result = ResourceMappingImpl()
+                try {
+                    val stackRequest = ListStacksRequest.builder()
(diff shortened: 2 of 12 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
```diff
@@ -54,6 +54,8 @@ interface NamedMap {
 interface Resource : NamedMap {
     fun isType(requestedType: String): Boolean
     fun type(): String?
+
+    fun getEnvironmentVariables(): Sequence<Variable>
```
kiiadi: Environment variables are only on a handful of resources, for our purposes they're available on [`AWS::Lambda::Function`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html) and [`AWS::Serverless::Function`](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction) - both of which are encapsulated in the `software.aws.toolkits.jetbrains.services.cloudformation.Function` implementations - we should probably move this to there.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/yaml/YamlCloudFormationTemplate.kt
```diff
@@ -40,20 +42,22 @@ class YamlCloudFormationTemplate(template: YAMLFile) : CloudFormationTemplate {
 
     override fun resources(): Sequence<Resource> {
         val resourcesBlock = templateRoot.getKeyValueByKey("Resources") ?: return emptySequence()
-        val resources = PsiTreeUtil.findChildOfAnyType(resourcesBlock, YAMLMapping::class.java) ?: return emptySequence()
+        val resources = PsiTreeUtil.findChildOfAnyType(resourcesBlock, YAMLMapping::class.java)
+                ?: return emptySequence()
```
kiiadi: nit: seems like unnecessary whitespace only change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplate.kt
```diff
@@ -62,6 +64,13 @@ interface Parameter : NamedMap {
     fun constraintDescription(): String?
 }
 
+interface Variable {
```
kiiadi: nit (naming): We seem to use two terms to refer to these `Variable` and `EnvironmentVariable` - we should pick one and be consistent.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
```diff
@@ -0,0 +1,86 @@
...
+    object EMPTY : ResourceMapping {
+        override fun listAllSummaries(logicalResource: String): Collection<StackResourceSummary> = Collections.emptyList()
+
+        override fun guessPhysicalResourceId(logicalResource: String): String? = null
+
+        override fun isEmpty(): Boolean = true
+    }
+}
+
+internal class ResourceMappingImpl : ResourceMapping {
```
kiiadi: nit (naming): the pattern we've used in the toolkit where we have an interface with a single implementation is to remove `Impl` and prefix with `Default` - so this would become `DefaultResourceMapping`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
```diff
@@ -0,0 +1,86 @@
...
+
+            ApplicationManager.getApplication().executeOnPooledThread {
+                val result = ResourceMappingImpl()
+                try {
+                    val stackRequest = ListStacksRequest.builder()
+                            .stackStatusFilters(StackStatus.CREATE_COMPLETE, StackStatus.UPDATE_COMPLETE)
+                            .build()
+                    client.listStacksPaginator(stackRequest).stream()
+                            .flatMap { it.stackSummaries().stream() }
+                            .flatMap { client.listResourcesForStack(it) }
```
kiiadi: Our `AWS Resources` toolwindow is getting hammered by Cloudformations' low TPS - listResourcesForStack is 0.5 TPS with bursts to 10 - these calls get throttled extremely quickly. Instead of iterating over all the stacks I think we'll need to have the user select a stack they want to use to populate the values.

Nice thing about that is you can then get rid of the drop-down on the value itself - we'll just populate all values for the given reference off the stack selected (if we can find it).

The other thing I like about this approach is that we can (in the future) have a 'populate from remote lambda' option where the customer picks a lambda and we import the environment variables from there.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    private fun registerResource(resourceSummary: StackResourceSummary) =
+            resourceSummary.logicalResourceId()?.let { logicalToSummary.putValue(it, resourceSummary) }
+
+    override fun isEmpty(): Boolean = logicalToSummary.isEmpty
+
+    companion object {
+        @JvmStatic
+        fun promiseMapping(project: Project): CompletionStage<ResourceMapping> {
+            val client: CloudFormationClient = AwsClientManager.getInstance(project).getClient()
```
kiiadi: We have an extension function on `Project` called `awsClient()` that you can use to shortcut this.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/ResourceMapping.kt
```diff
@@ -0,0 +1,86 @@
...
+
+                    promise.complete(result)
+                } catch (e: Exception) {
+                    promise.completeExceptionally(e)
+                }
+            }
+            return promise
+        }
+
+        private fun CloudFormationClient.listResourcesForStack(stackSummary: StackSummary): Stream<StackResourceSummary> {
```
kiiadi: nit: should we use a Kotlin `Sequence` instead?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/yaml/YamlCloudFormationTemplate.kt
```diff
@@ -90,19 +107,21 @@ class YamlCloudFormationTemplate(template: YAMLFile) : CloudFormationTemplate {
...
         private fun loadYamlFile(project: Project, templateFile: VirtualFile): YAMLFile = PsiFileFactory.getInstance(project).createFileFromText(
-            "template_temp.yaml",
-            YAMLLanguage.INSTANCE,
-            StringUtil.convertLineSeparators(VfsUtil.loadText(templateFile)),
-            false,
-            false,
-            true
+                "template_temp.yaml",
+                YAMLLanguage.INSTANCE,
+                StringUtil.convertLineSeparators(VfsUtil.loadText(templateFile)),
```
kiiadi: spacing here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamRunConfiguration.kt
```diff
@@ -215,6 +216,8 @@ class SamRunSettingsEditor(project: Project) : SettingsEditor<SamRunConfiguratio
     private val credentialManager = CredentialManager.getInstance()
 
     init {
+        Disposer.register(this, view)
```
kiiadi: Good catch!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamRunSettingsEditorPanel.java
```diff
@@ -62,7 +64,14 @@ public SamRunSettingsEditorPanel(Project project, TextCompletionProvider handler
         useTemplate.addActionListener(e -> updateComponents());
         addQuickSelect(templateFile.getTextField(), useTemplate, this::updateComponents);
         templateFile.addActionListener(new TemplateFileBrowseListener());
+
         updateComponents();
+        startRefreshResourceMappings();
+    }
+
+    private void startRefreshResourceMappings() {
+        ResourceMappingImpl.promiseMapping(project)
```
kiiadi: The pattern we've used in the past is `ApplicationManager`'s `invokeLater` and `executeOnPooledThread` in order to push something into the background and then come back to update the UI. Happy to take a steer on what's better - but is there a reason for this approach over the other?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamRunSettingsEditorPanel.java
```diff
@@ -71,28 +80,44 @@ private void createUIComponents() {
...
+        Function selectedFunction = null;
         if (useTemplate.isSelected()) {
             handler.setBackground(UIUtil.getComboBoxDisabledBackground());
             handler.setForeground(UIUtil.getComboBoxDisabledForeground());
 
             if (functionModels.getSelectedItem() instanceof Function) {
-                Function selected = (Function) functionModels.getSelectedItem();
-                handler.setText(selected.handler());
-                runtime.setSelectedItem(Runtime.fromValue(selected.runtime()));
+                selectedFunction = (Function) functionModels.getSelectedItem();
```
kiiadi: Can we push this logic into a function/method and return the `function` instance so that we don't have to declare the `Function selectedFunction = null` call above?

Issue #563, opened 2018-11-21
ChangeCredentialsAction doesn't handle setting credentials with no active creds


abrooksv:
```
update failed for ActionGroup: Profile:mogren (null)[Profile:mogren]
 
software.aws.toolkits.core.credentials.CredentialProviderNotFound: No active credential provider configured
      at software.aws.toolkits.jetbrains.core.credentials.DefaultProjectAccountSettingsManager.getActiveCredentialProvider(ProjectAccountSettingsManager.kt:114)
      at software.aws.toolkits.jetbrains.core.ChangeCredentialsAction.isSelected(AwsSettingsPanel.kt:217)
      at com.intellij.openapi.actionSystem.ToggleAction.update(ToggleAction.java:65)
      at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:171)
      at com.intellij.openapi.actionSystem.impl.Utils.doUpdate(Utils.java:256)
      at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:199)
      at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:150)
      at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:97)
      at com.intellij.ui.popup.ActionStepBuilder.appendActionsFromGroup(ActionStepBuilder.java:117)
      at com.intellij.ui.popup.ActionStepBuilder.buildGroup(ActionStepBuilder.java:73)
      at com.intellij.ui.popup.PopupFactoryImpl$ActionGroupPopup.getActionItems(PopupFactoryImpl.java:279)
      at com.intellij.ui.popup.PopupFactoryImpl$ActionGroupPopup.createStep(PopupFactoryImpl.java:260)
      at com.intellij.ui.popup.PopupFactoryImpl$ActionGroupPopup.<init>(PopupFactoryImpl.java:198)
      at com.intellij.ui.popup.PopupFactoryImpl$ActionGroupPopup.<init>(PopupFactoryImpl.java:182)
      at com.intellij.ui.popup.PopupFactoryImpl.createActionGroupPopup(PopupFactoryImpl.java:356)
      at com.intellij.openapi.ui.popup.JBPopupFactory.createActionGroupPopup(JBPopupFactory.java:198)
[...]

PR #568, opened 2018-11-21
Fix being unable to set active creds, if there has never been any

abrooksv:
Also cache the lookup of credential factories

## Types of changes

## Description
If the active cred provider was null (never set, no default profile), we would crash when trying to determine which of the credentials should be selected in the pop ups.

If there was no default profile, we also paid an expensive cost (and locked up the UI)  due to re-creating all the credential providers and parsing of the profile file for each possible profile. We instead cache the profile factories so they are only ever created once.

## Related Issue(s)
#563

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSettingsPanel.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSettingsPanel.kt
@@ -26,6 +26,7 @@ import com.intellij.util.Consumer
 import software.aws.toolkits.core.credentials.CredentialProviderNotFound
 import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
 import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.core.utils.tryOrNull
 import software.aws.toolkits.jetbrains.core.credentials.CredentialManager
 import software.aws.toolkits.jetbrains.core.credentials.ProjectAccountSettingsManager
 import software.aws.toolkits.jetbrains.core.credentials.ProjectAccountSettingsManager.AccountSettingsChangedNotifier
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSettingsPanel.kt
```diff
@@ -214,7 +214,11 @@ private class ChangeRegionAction(val region: AwsRegion) : ToggleAction(region.di
 
 private class ChangeCredentialsAction(val credentialsProvider: ToolkitCredentialsProvider) : ToggleAction(credentialsProvider.displayName), DumbAware {
 
-    override fun isSelected(e: AnActionEvent): Boolean = getAccountSetting(e).activeCredentialProvider == credentialsProvider
+    override fun isSelected(e: AnActionEvent): Boolean = try {
```
kiiadi: another way to do this is 

```
tryOrNull { getAccountSetting(e).activeCredentialProvider == credentialsProvider } ?: false
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ExtensionPointCredentialsProviderRegistry.kt
```diff
@@ -33,20 +33,24 @@ class CredentialProviderFactoryEP : AbstractExtensionPointBean() {
...
     companion object {
         private val LOG = Logger.getInstance(ExtensionPointCredentialsProviderRegistry::class.java)
         private const val EP_NAME = "aws.toolkit.credentialProviderFactory"
-        val EXTENSION_POINT = ExtensionPointName.create<CredentialProviderFactoryEP>(EP_NAME)
+        private val EXTENSION_POINT = ExtensionPointName.create<CredentialProviderFactoryEP>(EP_NAME)
+
+        private val factories by lazy {
+            EXTENSION_POINT.extensions
+                .mapNotNull {
+                    try {
```
kiiadi: `LOG.tryOrNull { it.getHandler() }`

abrooksv: That is an extension on slf4j, not the IJ diagnostic Logger

abrooksv: Guess using slf4j isnt a big deal here

kiiadi: one just delegates to the other so it's much of a muchness.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSettingsPanel.kt
abrooksv: Done

mogren:
Thanks for the quick fix!

Issue #618, opened 2018-11-27
Runtime in run configuration is not disabled on initial dialog open


rli:
![untitled2](https://user-images.githubusercontent.com/742829/49099455-58f60f80-f226-11e8-873f-f7e690f5bde9.gif)

zhangzhx:
This is due to race condition that the enability status of the combo box is accessed by multiple threads.

The fix is here https://github.com/aws/aws-toolkit-jetbrains/pull/620.

abrooksv:
#620 is not the best way to fix this in this use case.

PR #620, opened 2018-11-27
Fix bug of race condition to the Resource Selector ComboBox.

zhangzhx:
## Types of changes

## Description
Fix the race condition to the `ResourceSelector` combo box. See #618 for detailed description.

## Motivation and Context
We should always disable the combo box when it is in loading status even the status is also controlled by other actions, eg a radio button. The status should be changed to the desired one once the loading finishes successfully. If the loading failed, we should always disable it.

## Screenshots
**Before**
before

**After**
after

## Checklist

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/ResourceSelectorTest.kt
```diff
@@ -59,6 +60,32 @@ class ResourceSelectorTest {
         assertThat(comboBox.isEnabled).isEqualTo(false)
     }
 
+    @Test
+    fun comboBoxPopulation_updateStateToDesired() {
```
abrooksv: This test confuses me.

What puts the loading status into SUCCESSFUL? There is a while loop blocking the unlock call while the populateValues runs in a loop while the lock is locked. So how is it unlocked, so that it loads?

zhangzhx: The initial status for `ResourceSelector` is `SUCCESSFUL` and the status could be changed into `SUCCESSFUL` after the resources are loaded successfully (ie. the `block` call has been ran successfully).

When loading resources, the status changes as `SUCCESSFUL(the initial status) -> LOADING -> SUCCESSFUL`. We need to wait the status to change to `LOADING` until we call `enable`. That is why a `while` loop is there waiting for the status to change to `LOADING`. When calling `enable`, we need to make sure the status is in `LOADING` and the lock is meant to act as loading the resources so that after unlocking, the resources are loaded.

abrooksv: It seems to conflate 2 ideas into 1 boolean.

From an API perspective, the combo box has 2 states. isEnabled and isLoading. 
The actual UI element should be disabled if isEnabled == false || isLoading == true.

The fact that this code exists:
```
 comboBox.isEnabled = true
 assertThat(comboBox.isEnabled).isEqualTo(false)
``` 
is super confusing. I just set it to enabled and yet it is still false.

abrooksv: Cut #726 to revisit the design of this combo box

PR #637, opened 2018-11-29
AWS Explorer error node now provides a little more detail from AwsServiceExceptions

awschristou:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Description

Show a little more detail in the AWS Explorer Node when resources cannot be loaded

## Related Issue(s)

#634 #578 

## Screenshots (if appropriate)
image

## Checklist

- [ ] I have read the **README** document
- [ ] I have read the **CONTRIBUTING** document
- [ ] Local run of `gradlew check` succeeds
- [ ] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] A short description of the change has been added to the **CHANGELOG**

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #638, opened 2018-11-29
Fix case where user doesn't have Lambda access and tries to open a CloudFormation stack

awschristou:
## Types of changes

## Description

Accounts without Lambda access throw an exception when trying to open a CloudFormation Stack that contains Lambda Functions.
Now the stack can be opened, and Lambda stack resources will have Error Nodes as a more graceful handling.

image


abrooksv:
Can we give a better reason like we just did in #637

awschristou:
@abrooksv when #637 gets submitted, this one automatically benefits from it

abrooksv:
Ah. K

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationExplorerNodes.kt
```diff
@@ -105,23 +106,29 @@ class CloudFormationStackNode(project: Project, val stackName: String, private v
...
-
-            LambdaFunctionNode(
-                nodeProject,
-                client,
-                response.configuration().toDataClass(credentialProvider.id, region),
-                true
-            )
+    private fun loadServerlessStackResources(): List<AwsExplorerNode<*>> {
+        try {
+            return cfnClient
```
abrooksv: pull the return out of the try and do `return try {`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationExplorerNodes.kt
```diff
@@ -105,23 +106,27 @@ class CloudFormationStackNode(project: Project, val stackName: String, private v
...
+
+                LambdaFunctionNode(
+                    nodeProject,
+                    client,
+                    response.configuration().toDataClass(credentialProvider.id, region),
+                    true
+                )
+            }
+            .toList()
+    } catch (e: Exception) {
```
abrooksv: When we add our 2nd service, we should make this more fine graned

Issue #653, opened 2018-12-01
Update function code should not have xray configuration


zhangzhx:
**Describe the bug**

**To reproduce**
1. Go to AWS Explorer and select **Lambda**
1. Right click one Lambda function, and choose **Update Function Code** action
1. The dialog has enable xray configuration.

**Expected behavior**

The dialog should not have xray configuration. It should be only in the **Update Function Configuration** dialog

**Screenshots**
<img width="777" alt="screen shot 2018-11-30 at 4 16 15 pm" src="https://user-images.githubusercontent.com/7167513/49321286-552eeb00-f4bb-11e8-880e-fecbc2d29196.png">

**Your Environment**
 - OS: 
 - JetBrains' Product:
 - JetBrains' Product Version:
 - Toolkit Version:

**Additional context**

PR #654, opened 2018-12-01
Not showing xray config when updating function code only.

zhangzhx:
Update Function Code dialog should not show XRay config.

## Types of changes

## Related Issue(s)

#653 

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
@@ -77,7 +77,7 @@ class EditFunctionDialog(
     private val role: IamRole? = null
 ) : DialogWrapper(project) {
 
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
```diff
@@ -155,7 +155,7 @@ class EditFunctionDialog(
 
         val regionProvider = AwsRegionProvider.getInstance()
         val settings = ProjectAccountSettingsManager.getInstance(project)
-        view.setXrayControlVisibility(regionProvider.isServiceSupported(settings.activeRegion, "xray"))
+        view.setXrayControlVisibility(mode !== UPDATE_CODE && regionProvider.isServiceSupported(settings.activeRegion, "xray"))
```
awschristou: Other checks in this method use `==` on `mode`. Are you sure you don't want to use `!=` instead of `!==` ?

zhangzhx: Fixed this. Not sure the reason why we used `==` for an enum, but to be consistent, I updated.

Issue #721, opened 2019-01-04
Endpoints file not always downloaded correctly.


abrooksv:
**Describe the bug**
Endpoints file was downloaded but got a FileNotFoundException.

```
org.picocontainer.defaults.PicoInvocationTargetInitializationException: InvocationTargetException: java.util.concurrent.ExecutionException java.nio.file.NoSuchFileException: …\AppData\Local\Temp\service-endpoints.json-1258f77d-000f-4977-9880-3ae7c40683746418780833454079290.tmp
	at com.intellij.util.pico.CachingConstructorInjectionComponentAdapter.doGetComponentInstance(CachingConstructorInjectionComponentAdapter.java:92)
	at com.intellij.util.pico.CachingConstructorInjectionComponentAdapter.instantiateGuarded(CachingConstructorInjectionComponentAdapter.java:62)
	at com.intellij.util.pico.CachingConstructorInjectionComponentAdapter.getComponentInstance(CachingConstructorInjectionComponentAdapter.java:45)
	at com.intellij.openapi.components.impl.ServiceManagerImpl$MyComponentAdapter.getComponentInstance(ServiceManagerImpl.java:210)
	at com.intellij.util.pico.DefaultPicoContainer.getLocalInstance(DefaultPicoContainer.java:246)
	at com.intellij.util.pico.DefaultPicoContainer.getComponentInstance(DefaultPicoContainer.java:213)
	at com.intellij.openapi.components.ServiceManager.doGetService(ServiceManager.java:52)
	at com.intellij.openapi.components.ServiceManager.getService(ServiceManager.java:26)
	at software.aws.toolkits.jetbrains.core.region.AwsRegionProvider$Companion.getInstance(AwsRegionProvider.kt:42)
	at software.aws.toolkits.jetbrains.core.credentials.DefaultProjectAccountSettingsManager.<init>(ProjectAccountSettingsManager.kt:90)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.picocontainer.defaults.InstantiatingComponentAdapter.newInstance(InstantiatingComponentAdapter.java:193)
[...]

PR #722, opened 2019-01-04
Alias Path.delete to Files.deleteIfExists instead to prevent FileNotF…

abrooksv:
…oundExceptions

## Types of changes

## Description
When we delete files we do not actually care if it existed in the first place (if we truly do we can check the result code) so instead switch to deleteIfExists since that suppresses the FileNotFoundException

## Related Issue(s)
#721 

## Checklist

## License

### DIFF
--- a/core/src/software/aws/toolkits/core/utils/PathUtils.kt
+++ b/core/src/software/aws/toolkits/core/utils/PathUtils.kt
@@ -11,7 +11,7 @@ import java.nio.file.attribute.FileTime
 
 fun Path.inputStream(): InputStream = Files.newInputStream(this)
 fun Path.exists() = Files.exists(this)
-fun Path.delete() = Files.delete(this)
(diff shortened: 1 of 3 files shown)


core/src/software/aws/toolkits/core/utils/PathUtils.kt
```diff
@@ -11,7 +11,7 @@ import java.nio.file.attribute.FileTime
 
 fun Path.inputStream(): InputStream = Files.newInputStream(this)
 fun Path.exists() = Files.exists(this)
-fun Path.delete() = Files.delete(this)
+fun Path.delete() = Files.deleteIfExists(this)
```
kiiadi: I'd probably change the alias name too to `Path.deleteIfExists`

abrooksv: makes sense

abrooksv:
Updated commit message

PR #729, opened 2019-01-08
Increase compatible sam version to < 0.16

abrooksv:
## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamVersionTest.kt
```diff
@@ -56,7 +56,6 @@ class SamVersionTest {
     @Test
     fun incompatableSamVersion_tooHigh() {
         val versions = arrayOf(
-                "0.11.0",
```
awschristou: Consider adding some kind of `SamCommon.expectedSamMaxVersion  + increment` value here, so it is always testing a relevant value

abrooksv: It is non-inclusive so we could just use the version itself

PR #741, opened 2019-01-09
Remove dokka

abrooksv:
We don't publish it, so remove it.
Also remove unused gradle property for Java version

## License


kiiadi:
We'd probably want this back when (if) we ever publicize our extension points.

abrooksv:
Yaaa but since no one else can even depend on our plugin due to missing mocks... :-D

kiiadi:
Obvi.

PR #744, opened 2019-01-10
GUI testing using JetBrains' Test GUI Framework

rli:
`./gradlew cleanTest guiTest`

Initial thoughts:

* It appears that tests can't be run from IntelliJ
* `GuiTestRunner` should allow alternative Gradle executable name and Gradle task (i.e. don't hardcode `./gradlew runIde` so we can do something like `../gradlew :jetbrains-core:runIde`)
    * currently hacking around this by changing the test working directory and only enabling the guitest runIde task
* jacoco: refuses to instrument plugin classes without `includeNoLocationClasses`
* guitest.log location is hardcoded. Would like for this to be located in the `build` folder
* guitest sometimes gets stuck and it is unclear if the test will timeout or not
    * fixed by explicitly waiting for modals to close
* appears that the framework tries to open the message log at the end, but fails
* would be nice if we didn't have to explicitly filter out the tests so that only the TestSuites are run

## License


abrooksv:
CodeBuild spec:

```
version: 0.2

phases:
  install:
    commands:
      - apt-get update
      - apt-get install -y xvfb
      - pip install --user aws-sam-cli
  build:
    commands:
      - Xvfb :99 &
      - export DISPLAY=:99
      - chmod +x gradlew
      - export SAM_CLI_EXEC=$(python -m site --user-base)/bin/sam
      - ./gradlew guiTest
      - VCS_COMMIT_ID="$(git rev-parse HEAD)"
      - VCS_BRANCH_NAME="$CODEBUILD_SOURCE_VERSION"
      - CI_BUILD_URL=$(echo $CODEBUILD_AGENT_ENV_CODEBUILD_BUILD_URL | sed 's/#/%23/g') # Encode `#` in the URL because otherwise the url is clipped in the Codecov.io site
      - CI_BUILD_ID="${CODEBUILD_BUILD_ID}"
      - COV_SCRIPT=/root/.cache/codecov.sh
      - curl -s https://codecov.io/bash > "$COV_SCRIPT"
      - echo "$CODEBUILD_INITIATOR" | grep GitHub && bash "$COV_SCRIPT" -t $CODECV_TOEEN -F guitest || true
artifacts:
    files:
        - jetbrains-core-gui/build/idea-sandbox/system/log/gui-tests/**/*
```

Issue #760, opened 2019-01-21
Track SAM CLI version


abrooksv:
We should track the SAM version so we can get a better idea of when to drop older versions

PR #772, opened 2019-01-30
Include SAM version in metric event

abrooksv:
## Types of changes

## Related Issue(s)
#760 

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamCommon.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamCommon.kt
@@ -112,16 +112,16 @@ class SamCommon {
         }
 
         /**
-         * @return The SemVer of SAM, or null if unable to determine it
+         * @return The string representation of the SAM version else "UNKNOWN"
          */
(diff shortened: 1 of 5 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamCommon.kt
```diff
@@ -112,16 +112,16 @@ class SamCommon {
...
             val sanitizedPath = path.nullize(true)
-                ?: return null
+                ?: return "UNKNOWN"
 
             return try {
-                versionCache.getResult(sanitizedPath)
+                versionCache.getResult(sanitizedPath).rawVersion
             } catch (e: Exception) {
-                return null
+                return "UNKNOWN"
```
awschristou: Is it worth differentiating the unknown "we don't have a sam path" vs the unknown "we have sam but an error happened" ?

abrooksv: I dont think so from a telemetry standpoint. These are all gated by the validation logic anyway

PR #800, opened 2019-02-12
Fix cancel button in serverless deploy dialog

rli:
## Types of changes

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployDialog.kt
```diff
@@ -211,7 +213,9 @@ open class SamDeployDialog(
             }
         })
 
-        processHandler.startNotify()
+        ApplicationManager.getApplication().executeOnPooledThread {
+            CapturingProcessRunner(processHandler).runProcess(progressIndicator)
```
abrooksv: Can we move the future.complete to after this so we can only have 1 capture?

It has the exit code, so we should be able to completely remove lines 201-214, and move them down here

PR #810, opened 2019-02-22
Switch to using latches instead of sleeps

abrooksv:
## Types of changes

## Description
Attempt to fix flaky test

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/ResourceSelectorTest.kt
```diff
@@ -64,14 +64,11 @@ class ResourceSelectorTest {
     fun comboBoxPopulation_updateStateToDesired() {
         val items = listOf("foo", "bar", "baz")
 
-        val lock = ReentrantLock()
-        lock.lock()
+        val latch = CountDownLatch(1)
 
         comboBox.isEnabled = false
         comboBox.populateValues(updateStatus = false) {
-            while (lock.isLocked) {
```
rli: public boolean isLocked()
Queries if this lock is held by any thread. This method is designed for use in monitoring of the system state, not for synchronization control.

oops

PR #812, opened 2019-02-28
Make a new path mapper

abrooksv:
We need a custom path mapper that can handle multiple local mappings going to a single remote location for sam build

## Types of changes

## Description
Extracting out from sam build migration since its standalone

## Related Issue(s)
#811 

## License

### DIFF
--- a/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/PathMapperTest.kt
+++ b/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/PathMapperTest.kt
@@ -1,0 +1,114 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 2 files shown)


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/PathMapperTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        val file = tempFolder.root.toPath().resolve(path)
+        Files.createDirectories(file.parent)
+        if (!Files.exists(file)) {
+            Files.createFile(file)
+        }
+
+        return file.toString()
+    }
+
+    private fun mapper(init: MutableList<PathMapping>.() -> Unit) {
```
zhangzhx: nit: Can you rename this method to something like `initMapper`, it is a bit confusing to have the same name to the member variable.

abrooksv: Yep

Issue #865, opened 2019-03-29
Investigate and fix network calls on wrong Threads


abrooksv:
After fixing our assert statement, it has become evident we managed to put a few calls on the UI thread. We need to move these calls to the background / async.

#863 fixed some performance issues on start up, but it became evident that our credential switching / validation logic is wrong since it relies on UI calls and can't work in an async manor. We need to investigate the switching UX and implement it that is async friendly.

abrooksv:
Changes to credential flow:

1. User changes credentials
2. Toolkit broadcasts out a new state change across the `ACCOUNT_SETTINGS_CHANGED` MessageBus. This state represents that credentials are loading
3. Explorer and status bar go into a Loading Credentials state
4a. If credentials load, `ProjectAccountSettingsManager#activeCredentialProvider` will be set to the requested credentials
4b. If credentials fail,  `ProjectAccountSettingsManager#activeCredentialProvider` is nulled out
5. New state change is broadcasted with the new value

abrooksv:
Credential switching has been fixed:

Remaining issue(s) is:
1. expanding child nodes for CFN node
```
	at software.amazon.awssdk.services.cloudformation.DefaultCloudFormationClient.describeStackResources(DefaultCloudFormationClient.java:1195)
	at software.amazon.awssdk.services.cloudformation.CloudFormationClient.describeStackResources(CloudFormationClient.java:1938)
	at software.aws.toolkits.jetbrains.services.cloudformation.CloudFormationStackNode.loadServerlessStackResources(CloudFormationExplorerNodes.kt:119)
	at software.aws.toolkits.jetbrains.services.cloudformation.CloudFormationStackNode.updateCachedChildren(CloudFormationExplorerNodes.kt:105)
	at software.aws.toolkits.jetbrains.services.cloudformation.CloudFormationStackNode.getChildren(CloudFormationExplorerNodes.kt:90)
	at software.aws.toolkits.jetbrains.core.explorer.ExplorerToolWindow$createTree$2.treeWillExpand(ExplorerToolWindow.kt:125)
	at javax.swing.JTree.fireTreeWillExpand(JTree.java:2839)
	at javax.swing.JTree.setExpandedState(JTree.java:3630)
	at javax.swing.JTree.expandPath(JTree.java:2219)
	at javax.swing.plaf.basic.BasicTreeUI.toggleExpandState(BasicTreeUI.java:2286)
```

abrooksv:
[...]

PR #869, opened 2019-04-01
Re-work credential switching to work asynchronously

abrooksv:
## Types of changes

## Description
* Cut down on the number of validation calls we make
* Tweak when we fall back to the default profile
* Re-work the settings notification bus
* More info in #865

## Motivation and Context
This moves the STS calls off the UI thread

## Related Issue(s)
#865

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
@@ -3,8 +3,10 @@
 
 package software.aws.toolkits.jetbrains.core.credentials
 
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.components.PersistentStateComponent
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.components.State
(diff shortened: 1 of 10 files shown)


abrooksv:
Will have a follow up PR for the changes to status bar & explorer for loading mode

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -64,6 +75,21 @@ interface ProjectAccountSettingsManager {
...
 
+    /**
+     * Attempts to change the active credential provider.
+     *
+     * 1. Broadcasts a [AccountSettingsChangedNotifier] with [AccountSettingsEvent.isLoading] set to true
+     * 2. Kicks off an STS callerIdentity call to validate the credentials work in the background
+     * 3. If call succeeds, make the requested provider active, broadcast [AccountSettingsEvent.activeCredentialsChanged]
+     * 4. If call fails, null out the active provider, broadcast [AccountSettingsEvent.activeCredentialsChanged]
+     */
+    fun changeCredentialProvider(credentialsProvider: ToolkitCredentialsProvider?)
```
awschristou: The function comment seems like an implementation detail, which would belong on the `DefaultProjectAccountSettingsManager` version of `changeCredentialProvider`. Can it be on both?

abrooksv: Callers of the code never seen the true method since it is hidden behind the interface in the dependency injection system.

Javadoc / kdoc is also inherited on overriden methods unless explicitly overridden.

PR #877, opened 2019-04-03
Make sure calls to fetch Lambda configuration is not on UI thread

abrooksv:
## Types of changes

## Related Issue(s)
#865 

## Testing
Validated updating Lambda config still works

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/UpdateFunctionAction.kt
```diff
@@ -19,14 +21,20 @@ abstract class UpdateFunctionAction(private val mode: EditFunctionMode, title: S
         val project = e.getRequiredData(PlatformDataKeys.PROJECT)
 
         // Fetch latest version just in case
```
rli: Move this comment after the `executeOnPooledThread`

abrooksv: Fixed

Issue #887, opened 2019-04-05
Deploy Serverless Application dialog doesn't complain when putting an existing stack name in 'Create Stack' input


kyoge:
**Describe the bug**
When deploying serverless application, the dialog doesn't complain if I put the name of an existing stack (available in 'Update Stack' field) in 'Create Stack' input field and the deployment runs successfully as if I am updating the existing stack.

**To reproduce**

**Expected behavior**
The dialog should complain / shouldn't allow deployment.

**Screenshots**
<img width="1280" alt="0" src="https://user-images.githubusercontent.com/11089351/55648845-41133700-5796-11e9-8d78-eb10453e4346.png">

**Your Environment**
 - OS: 
 - JetBrains' Product:
 - JetBrains' Product Version:
 - Toolkit Version:
 - SAM CLI Version:
 - JVM/Python Version:

**Additional context**

zhangzhx:
This fix has been released in [version 1.3](https://plugins.jetbrains.com/plugin/11349-aws-toolkit/update/61424).

PR #899, opened 2019-04-09
Ensure that the new stack name in 'Create Stack' input is different than the existing stack name(s)

kyoge:
## Types of changes

## Description

- Check if the new stack name in 'Create Stack' input is same as an existing stack name.
- If the check passes, show message "New stack name must be different than the existing stack name(s)" to the user, deployment can't start until the stack name is corrected.

## Related Issue(s)

#887 

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
@@ -152,7 +152,7 @@ class DeploySamApplicationValidator {
 
     fun validateSettings(view: DeployServerlessApplicationPanel): ValidationInfo? {
         if (view.createStack.isSelected) {
-            validateStackName(view.newStackName.text)?.let {
+            validateStackName(view.newStackName.text, view.stacks.values())?.let {
                 return ValidationInfo(it, view.newStackName)
             }
         } else if (view.updateStack.isSelected && view.stacks.selected() == null) {
@@ -196,7 +196,7 @@ class DeploySamApplicationValidator {
         return null
     }
 
-    private fun validateStackName(name: String?): String? {
+    private fun validateStackName(name: String?, existingNames: List<String>): String? {
         if (name == null || name.isEmpty()) {
             return message("serverless.application.deploy.validation.new.stack.name.missing")
         }
@@ -205,6 +205,10 @@ class DeploySamApplicationValidator {
         }
         if (name.length > MAX_STACK_NAME_LENGTH) {
             return message("serverless.application.deploy.validation.new.stack.name.too.long", MAX_STACK_NAME_LENGTH)
+        }
+        // Check if the new stack name is same as an existing stack name
+        if (existingNames.contains(name)) {
+            return message("serverless.application.deploy.validation.new.stack.name.duplicate")
         }
         return null
     }
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
@@ -35,7 +35,7 @@ class DeployServerlessApplicationDialog(
     private val samPath: String = module?.let { relativeSamPath(it, templateFile) } ?: templateFile.name
 
     private val view = DeployServerlessApplicationPanel()
-    private val validator = DeploySamApplicationValidator()
+    private val validator = DeploySamApplicationValidator(view)
     private val s3Client: S3Client = project.awsClient()
     private val cloudFormationClient: CloudFormationClient = project.awsClient()
     private val templateParameters = CloudFormationTemplate.parse(project, templateFile).parameters().toList()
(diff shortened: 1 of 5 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
```diff
@@ -152,7 +152,7 @@ class DeploySamApplicationValidator {
 
     fun validateSettings(view: DeployServerlessApplicationPanel): ValidationInfo? {
         if (view.createStack.isSelected) {
-            validateStackName(view.newStackName.text)?.let {
+            validateStackName(view.newStackName.text, view.stacks.values())?.let {
```
kiiadi: Should prolly just make `view` an instance variable of `DeploySamApplicationValidator` since they have the same lifecycle - then we don't need to pass the parameters around so much.

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -138,6 +138,12 @@ class ResourceSelector<T> : ComboBox<T>() {
         }
     }
 
+    fun values(): List<T> = ArrayList<T>().apply {
+        for (i in 0..(itemCount - 1)) {
```
kiiadi: prefer the functional style here: `fun values(): List<T> = (0 until itemCount).map { getItemAt(it) }`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
```diff
@@ -196,7 +196,7 @@ class DeploySamApplicationValidator {
         return null
     }
 
-    private fun validateStackName(name: String?): String? {
+    private fun validateStackName(name: String?, existingNames: List<String>): String? {
```
kiiadi: if we are going to leave this as a parameter it should be a `Set` rather than a list - it's got a more efficient `exists` runtime complexity.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
```diff
@@ -206,6 +206,10 @@ class DeploySamApplicationValidator {
         if (name.length > MAX_STACK_NAME_LENGTH) {
             return message("serverless.application.deploy.validation.new.stack.name.too.long", MAX_STACK_NAME_LENGTH)
         }
+        // Check if the new stack name is same as an existing stack name
+        if (existingNames.contains(name)) {
```
kiiadi: Using the `in` operator is probably more idiomatic:

`if (name in existingNames) { .. }`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
kyoge: Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
kyoge: Done. Returned a set from here.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
kyoge: This parameter is removed now.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationDialog.kt
kyoge: Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -138,6 +138,8 @@ class ResourceSelector<T> : ComboBox<T>() {
         }
     }
 
+    fun values(): Set<T> = (0 until itemCount).map { getItemAt(it) }.toSet()
```
abrooksv: If you switch out the model to a CollectionComboBoxModel you dont need this method

abrooksv: But I guess if you want the set it doesnt save you

kyoge: Yes, switching to CollectionComboBoxModel gives me the List for free but since I want the Set, I'll keep it as is.

kiiadi: nit: I think the `values()` method should still return a list - other resource selectors may care about ordering (and may contain duplicates) which you will lose if you move this to a `Set`.

kiiadi: As a more long term comment I don't think we should be looking at the view ComboBox values for this at all - we should be looking at some canonical source of stacks - the combobox doesn't 'own' this data - the fact it's here is merely coincidental.

Soon we're going to need to have a cache of these sorts of things rather than calling down to the service all the time.

Issue #868, opened 2019-04-01
Java Gradle Multi-Module: "Module can only have 1 content root"


lukasbradley:
**Describe the bug**
In attempting to execute a Lambda in a multi-module Gradle project, the Plugin has a fatal error.

**To reproduce**
1.  Create a multi-module Gradle project.
2.  Select the RequestHandler class to execute.
3.  Run it...

**Expected behavior**
No fatal error...?

**Actual behavior**

IDE Fatal Error:  The Plugin crashes with the following stack trace.  
`
java.lang.IllegalArgumentException: Module can only have 1 content root
	at software.aws.toolkits.jetbrains.services.lambda.java.JavaLambdaBuilder.getBaseDirectory(JavaLambdaBuilder.kt:43)
	at software.aws.toolkits.jetbrains.services.lambda.java.JavaLambdaBuilder.buildLambda(JavaLambdaBuilder.kt:30)
	at software.aws.toolkits.jetbrains.services.lambda.LambdaBuilderUtils.runSamBuild(LambdaBuilderUtils.kt:133)
	at software.aws.toolkits.jetbrains.services.lambda.LambdaBuilderUtils.access$runSamBuild(LambdaBuilderUtils.kt:32)
	at software.aws.toolkits.jetbrains.services.lambda.LambdaBuilderUtils$buildAndReport$$inlined$runSamBuildInBackground$1.run(LambdaBuilderUtils.kt:177)
	at com.intellij.openapi.progress.impl.CoreProgressManager$TaskRunnable.run(CoreProgressManager.java:731)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:164)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:586)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:532)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:86)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:151)
	at com.intellij.openapi.progress.impl.CoreProgressManager$4.run(CoreProgressManager.java:403)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:311)
[...]

Issue #857, opened 2019-03-28
Lambda function in Gradle Project failed to Running locally


K1-Style:
**Describe the bug**
Running Lambda function in Gradle Project fails with "SAM build command failed"

**To reproduce**
1. Create a Gradle SAM application
2. Find `build.gradle` and right click `Import Gradle Project`
3. Open `App.java` and click Lambda icon on the left of the class definition.
4. Select `Run '[Local] App'`
5. Select my Credentials and Region, and Input Text(API Gateway AWS Proxy)
6. Execute `Run`, but it fails with "SAM build command failed"

Build Output details is below.

```
/usr/local/bin/sam build --template /private/var/folders/94/599b3s8j0tbgnn506tmtd17m0000gq/T/template381.yaml --build-dir /private/var/folders/94/599b3s8j0tbgnn506tmtd17m0000gq/T/lambdaBuild1175
Build Failed
2019-03-28 21:41:51 Building resource 'Function'
Error: Unable to find a supported build workflow for runtime 'java8'. Reason: None of the supported manifests '['build.gradle', 'pom.xml']' were found in the following paths '['/Users/nomura.keichi/project/serverless/test/HelloWorldFunction/src/main', '/private/var/folders/94/599b3s8j0tbgnn506tmtd17m0000gq/T']'
```

**Expected behavior**
The expected behavior is same as "Running a function locally" on [AWS Toolkit for IntelliJ – Now generally available \| AWS Developer Blog](https://aws.amazon.com/jp/blogs/developer/aws-toolkit-for-intellij-now-generally-available/)

**Screenshots**
![スクリーンショット 2019-03-28 21 49 29](https://user-images.githubusercontent.com/18533318/55158960-740c6800-51a3-11e9-9f61-ed6ce441612b.png)
![スクリーンショット 2019-03-28 21 27 57](https://user-images.githubusercontent.com/18533318/55159135-c2216b80-51a3-11e9-8f56-973d590b4036.png)
![スクリーンショット 2019-03-28 21 42 46](https://user-images.githubusercontent.com/18533318/55159162-cbaad380-51a3-11e9-98ed-bbeecf7f92a2.png)
![スクリーンショット 2019-03-28 21 43 03](https://user-images.githubusercontent.com/18533318/55159170-d1081e00-51a3-11e9-9863-0c991e45670b.png)

**Your Environment**
 - OS: macOS Mojave 10.14.4
 - JetBrains' Product: IntelliJ IDEA
[...]

PR #900, opened 2019-04-09
Fix java invalid base dir

abrooksv:
## Types of changes

## Description
Query project root from Maven/Gradle to determine the base directory for Java

When using Java handler based build, query Maven/Gradle for the project root.

## Related Issue(s)
#868, #857

## Testing
We now actually import the projects and then try to build them

## License

### DIFF
--- a/jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
+++ b/jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
@@ -3,6 +3,7 @@
 
 package software.aws.toolkits.jetbrains.services.lambda.java
 
+import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
@@ -11,6 +12,7 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaBuilder
 import software.aws.toolkits.jetbrains.utils.rules.HeavyJavaCodeInsightTestFixtureRule
 import software.aws.toolkits.jetbrains.utils.rules.addFileToModule
 import software.aws.toolkits.jetbrains.utils.rules.addModule
+import software.aws.toolkits.resources.message
 import java.nio.file.Paths
 
 class JavaLambdaBuilderTest : BaseLambdaBuilderTest() {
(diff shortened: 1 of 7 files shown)


abrooksv:
Opps, need to rebase on to clean origin/master

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/java/JavaTestUtils.kt
```diff
@@ -50,6 +75,68 @@ internal fun HeavyJavaCodeInsightTestFixtureRule.setUpGradleProject(): PsiClass
...
+            error.set(errorMessage)
+        }
+    }
+
+    val importSpecBuilder = ImportSpecBuilder(project, GradleConstants.SYSTEM_ID)
+        .callback(refreshCallback)
+        .forceWhenUptodate()
+        .use(ProgressExecutionMode.MODAL_SYNC)
+
+    ExternalSystemUtil.refreshProjects(importSpecBuilder)
```
kiiadi: does this happen synchronously?

abrooksv: Yes, that is controlled by the ProgressExecutionMode enum 2 lines up

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
```diff
@@ -130,4 +131,24 @@ class JavaLambdaBuilderTest : BaseLambdaBuilderTest() {
...
+                    return input.toUpperCase();
+                }
+            }
+            """.trimIndent()
+        )
+
+        assertThatThrownBy {
+            buildLambda(projectRule.module, handlerPsi, Runtime.JAVA8, "com.example.SomeClass")
+        }.isInstanceOf(IllegalStateException::class.java)
+            .hasMessageEndingWith("is not managed by Maven or Gradle")
```
zhangzhx: Hmm, as the exception message is managed by the localization properties, I'm wondering if it is appropriate to test the exception message

abrooksv: Ya it is fragile, but it does enforce the correct exception is thrown and not a random different IllegalState

zhangzhx: I understand you want to enforce the correct exception. It's better to fix the fragile test, but I'm fine with it too considering we are not expanding the localization in a short term.

abrooksv: I can just use the message() call here as well. I'll make the change tomorrow

PR #931, opened 2019-04-25
Bump min version to 183.2153.8 to support getHelpTopic in LambdaRunCo…

zhangzhx:
…nfiguration.

The [getHelpTopic](https://github.com/JetBrains/intellij-community/blob/master/platform/lang-api/src/com/intellij/execution/configurations/ConfigurationType.java#L69-L77) API was introduced at the second release of version 183. We need to update the min version to support this.

## License

### DIFF
--- a/jetbrains-core/resources/META-INF/plugin.xml
+++ b/jetbrains-core/resources/META-INF/plugin.xml
@@ -65,7 +65,7 @@
     ]]></description>
 
     <vendor email="aws-toolkit-jetbrains@amazon.com" url="https://aws.amazon.com/">Amazon Web Services</vendor>
-    <idea-version since-build="183" until-build="191.*"/>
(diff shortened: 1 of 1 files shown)


jetbrains-core/resources/META-INF/plugin.xml
```diff
@@ -65,7 +65,7 @@
     ]]></description>
 
     <vendor email="aws-toolkit-jetbrains@amazon.com" url="https://aws.amazon.com/">Amazon Web Services</vendor>
-    <idea-version since-build="183" until-build="191.*"/>
+    <idea-version since-build="183.2153.8" until-build="191.*"/>
```
rli: 2018.3 corresponds to 183.4284.148, so this should be safe.

PR #953, opened 2019-05-08
Refactor TelemetryService to allow recording project level properties

zhangzhx:
We need to send active AWS account and region to each metric as metric metadata. As such, we need to make the TelemetryService to accept project level properties.

## License


abrooksv:
Having second guesses on this. Lets keep this at the application level and make a record(project: Project, ...)

Issue #881, opened 2019-04-04
Add more metadata to telemetry calls


abrooksv:
active accountID can be retrieved form the sts getCallerInfo call, best effort.
active regionId

PR #963, opened 2019-05-15
Add active AWS account and active region to telemetry

zhangzhx:
## Types of changes

## Related Issue(s)
#881 

## Checklist

## License

### DIFF
--- a/core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
+++ b/core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
@@ -38,7 +38,7 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
      * Returns true or throws an Exception
      */
     @Throws(Exception::class)
-    open fun getAwsAccountOrThrow(stsClient: StsClient): String =
+    open fun getAwsAccount(stsClient: StsClient): String =
         stsClient.getCallerIdentity {
             it.overrideConfiguration { overrides ->
                 overrides.credentialsProvider(StaticCredentialsProvider.create(resolveCredentials()))
--- a/core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
+++ b/core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
@@ -38,13 +38,12 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
      * Returns true or throws an Exception
      */
     @Throws(Exception::class)
-    open fun isValidOrThrow(stsClient: StsClient) {
+    open fun getAwsAccountOrThrow(stsClient: StsClient): String =
         stsClient.getCallerIdentity {
             it.overrideConfiguration { overrides ->
                 overrides.credentialsProvider(StaticCredentialsProvider.create(resolveCredentials()))
             }
-        }
-    }
+        }.account()
 }
 
 /**
--- a/core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
+++ b/core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
@@ -4,20 +4,26 @@
 package software.aws.toolkits.core.telemetry
 
 import software.aws.toolkits.core.telemetry.MetricEvent.Companion.illegalCharsRegex
-import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
 import software.aws.toolkits.core.utils.getLogger
 import software.aws.toolkits.core.utils.warn
 import java.time.Instant
+import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
 
 interface MetricEvent {
     val namespace: String
     val createTime: Instant
+    val awsAccount: String
+    val awsRegion: String
     val data: Iterable<Datum>
 
     interface Builder {
         fun namespace(namespace: String): Builder
 
         fun createTime(createTime: Instant): Builder
+
+        fun awsAccount(awsAccount: String): Builder
+
+        fun awsRegion(awsRegion: String): Builder
 
         fun datum(name: String, buildDatum: Datum.Builder.() -> Unit = {}): Builder
 
@@ -61,10 +67,15 @@ fun String.replaceIllegal(replacement: String = "") = this.replace(illegalCharsR
 class DefaultMetricEvent internal constructor(
     override val namespace: String,
     override val createTime: Instant,
+    override val awsAccount: String,
+    override val awsRegion: String,
     override val data: Iterable<MetricEvent.Datum>
 ) : MetricEvent {
+
     class BuilderImpl(private var namespace: String) : MetricEvent.Builder {
         private var createTime: Instant = Instant.now()
+        private var awsAccount: String = METADATA_NA
+        private var awsRegion: String = METADATA_NA
         private var data: MutableCollection<MetricEvent.Datum> = mutableListOf()
 
         override fun namespace(namespace: String): MetricEvent.Builder {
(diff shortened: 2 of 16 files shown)


core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -38,13 +38,12 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
      * Returns true or throws an Exception
      */
     @Throws(Exception::class)
-    open fun isValidOrThrow(stsClient: StsClient) {
+    open fun getAwsAccountOrThrow(stsClient: StsClient): String =
```
abrooksv: I think we can drop the "OrThrow", its kinda an anti-pattern in naming

zhangzhx: Done.

abrooksv:
This worries me on action events. In the CloudFormation stack viewer, you can switch accounts but keep monitoring the old account as long as the window stays open. We should probably look into using the DataContextProvider system to provide the active credential provider ID for THAT UI / Action area.

abrooksv:
We also don't currently have any actions registered in the stack viewer though, which is also a large blindspot in our telemetry. /shrug

abrooksv:
We can probably attempt to load it from an action's datacontext and if that fails fall back to project's active.

zhangzhx:
Good catch. Also, it applies to the run configuration action where user can choose different set of credentials/region other than the global active one.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -60,6 +61,16 @@ interface ProjectAccountSettingsManager {
...
 
+    /**
+     * The underlying AWS account for current active credential provider. Return null if credential provider is not set
+     */
+    val activeAwsAccount: String?
+
+    /**
+     * Return the underlying AWS account for the given credential provider. Return null if the credential provider is invalid
+     */
+    fun awsAccount(credentialProvider: ToolkitCredentialsProvider): String?
```
kiiadi: Doesn't feel like this belongs here - all the other fields / methods surface information that's related to the current 'connection' or mutate the connection - this does neither.

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -38,13 +38,12 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
      * Returns true or throws an Exception
      */
     @Throws(Exception::class)
-    open fun isValidOrThrow(stsClient: StsClient) {
+    open fun getAwsAccount(stsClient: StsClient): String =
```
kiiadi: this doesn't feel like we have the correct dependency model here - you shouldn't have to pass an instance of an sts client in in order to be able to satisfy the call to this method.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -161,6 +176,16 @@ class DefaultProjectAccountSettingsManager(private val project: Project, private
         @Throws(CredentialProviderNotFound::class)
         get() = activeProfileInternal ?: throw CredentialProviderNotFound(message("credentials.profile.not_configured"))
 
+    override val activeAwsAccount: String?
+        get() = if (hasActiveCredentials()) awsAccountCache[activeCredentialProvider.id] else null
```
kiiadi: When does this expire?

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
zhangzhx: We shouldn't, but the trick here is that this class `ToolkitCredentialsProvider` is in `core` module, while the toolkit user agent is in `jetbrains-core` module. We want to use the same toolkit user agent everywhere.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
zhangzhx: Make sense. Moved the account cache management to a separate application level service.

zhangzhx:
When you do actions in the IDE:
**start the IDE** -> **change profile** -> **change region** -> **run SAM local with an invalid credential** -> **exit the IDE**.
The metric metadata will look something like below:
```
ToolkitStart
ChangeCredentialsAction.StatusBarPlace
  awsAccount: 111111111111
  awsRegion: us-east-1
ChangeRegionAction.StatusBarPlace
  awsAccount: 111111111111
  awsRegion: us-west-2
SamInvoke.Run
  samVersion: 0.16.0
  hasException: false
  runtime: JAVA8
  templateBased: false
  awsAccount: invalid
  awsRegion: us-east-1
ToolkitEnd.duration
  awsAccount: n/a
  awsRegion: n/a
```

core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
```diff
@@ -4,21 +4,27 @@
 package software.aws.toolkits.core.telemetry
 
 import software.aws.toolkits.core.telemetry.MetricEvent.Companion.illegalCharsRegex
-import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
 import software.aws.toolkits.core.utils.getLogger
 import software.aws.toolkits.core.utils.warn
 import java.time.Instant
+import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
```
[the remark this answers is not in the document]
zhangzhx: Created a backlog issue for tracking this: #973. Feel free to add more context to it.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
zhangzhx: We currently don't support expiration at the moment. If user select a session based profile, it will be valid until expiration, say 1 hour. After 1 hour, user will have runtime error if they don't switch active credential.

core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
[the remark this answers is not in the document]
kiiadi: there already are some ktlint rules that cover this - you can't import * for example.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsAccountCache.kt
```diff
@@ -0,0 +1,61 @@
...
+        AwsSdkClient.getInstance().sdkHttpClient,
+        Region.US_EAST_1,
+        AnonymousCredentialsProvider.create(),
+        AwsClientManager.userAgent
+    ))
+
+    init {
+        ApplicationManager.getApplication().messageBus.connect().subscribe(
+            CredentialManager.CREDENTIALS_CHANGED,
+            object : ToolkitCredentialsChangeListener {
```
kiiadi: could prolly just pass `this` and put the implementation methods on `DefaultAwsAccountCache`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsAccountCache.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+interface AwsAccountCache {
+    /**
+     * Return the underlying AWS account for the given credential provider. Return null if the credential provider is invalid
+     */
+    fun awsAccount(credentialProvider: ToolkitCredentialsProvider): String?
+
+    companion object {
+        fun getInstance(): AwsAccountCache = ServiceManager.getService(AwsAccountCache::class.java)
```
kiiadi: Should this be application wide or scoped to a `project` ?

zhangzhx: Like `CredentialManager`, I think this should be application wide since the credentials are shared through all projects.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsAccountCache.kt
zhangzhx: Done.

PR #966, opened 2019-05-16
Add support to nodejs10.x and fix a bug when inferring nodejs lambda

zhangzhx:
handler source root.

When the target potential lambda handler is from the dependency lib, the content root cannot be inferred. It should return null for it's source root instead of throwing exception.

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
@@ -32,7 +32,7 @@ class SamCommon {
         const val SAM_INVALID_OPTION_SUBSTRING = "no such option"
 
         // Inclusive
-        val expectedSamMinVersion = SemVer("0.14.1", 0, 14, 1)
+        val expectedSamMinVersion = SemVer("0.16.0", 0, 16, 0)
 
         // Exclusive
         val expectedSamMaxVersion = SemVer("0.23.0", 0, 23, 0)
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/wizard/SamInitProjects.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/wizard/SamInitProjects.kt
@@ -91,8 +91,7 @@ class SamDynamoDBCookieCutter : SamPythonProjectTemplate() {
 }
 
 abstract class SamNodeJsProjectTemplate : SamProjectTemplate() {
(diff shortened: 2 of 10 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
```diff
@@ -32,7 +32,7 @@ class SamCommon {
         const val SAM_INVALID_OPTION_SUBSTRING = "no such option"
 
         // Inclusive
-        val expectedSamMinVersion = SemVer("0.14.1", 0, 14, 1)
```
abrooksv: This will need a changelog entry of just like we did when we raised from 0.7,

I'd make a dedicated commit for this change

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/wizard/SamInitProjects.kt
```diff
@@ -91,8 +91,7 @@ class SamDynamoDBCookieCutter : SamPythonProjectTemplate() {
 }
 
 abstract class SamNodeJsProjectTemplate : SamProjectTemplate() {
-    // TODO support NodeJs 10.10 at some point
-    override fun supportedRuntimes(): Set<Runtime> = setOf(Runtime.NODEJS8_10)
+    override fun supportedRuntimes(): Set<Runtime> = setOf(Runtime.NODEJS8_10, Runtime.NODEJS10_X)
```
abrooksv: Hmmm should we just say NodeJsRuntimeGroup.runtimes()?

rli: No, I think we should be explicit here since we're also stating node8 and 10.x elsewhere

abrooksv: k

zhangzhx: For a long term, we need to load these configs dynamically. The template author should provide what runtimes it supports.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
zhangzhx: The whole nodejs feature branch will be merged into master in one commit. Is it worth having a separate commit for the min version bump?

Issue #128, opened 2018-08-02
Standardize on a assertion framework


abrooksv:
Discuss and settle on one assertion framework (assertj, assertk, hamcrest)

PR #967, opened 2019-05-16
Remove assertk

abrooksv:
## Motivation and Context
Standardized and removed some tech debt

## Related Issue(s)
#128

## License

### DIFF
--- a/core/tst/software/aws/toolkits/core/utils/ZipUtilsTest.kt
+++ b/core/tst/software/aws/toolkits/core/utils/ZipUtilsTest.kt
@@ -3,10 +3,8 @@
 
 package software.aws.toolkits.core.utils
 
-import assertk.assert
-import assertk.assertions.isEqualTo
-import assertk.assertions.isNotNull
-import assertk.fail
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.fail
 import org.junit.After
 import org.junit.Rule
 import org.junit.Test
@@ -34,6 +32,7 @@ class ZipUtilsTest {
         val fileToAdd = tmpFolder.newFile()
         fileToAdd.writeText("hello world", StandardCharsets.UTF_8)
         val zipFile = tmpFolder.newFile("blah.zip")?.toPath() ?: return fail("Couldn't create new file")
+
         ZipOutputStream(Files.newOutputStream(zipFile)).use {
             it.putNextEntry("file.txt", fileToAdd.toPath())
         }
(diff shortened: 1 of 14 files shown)


core/tst/software/aws/toolkits/core/utils/ZipUtilsTest.kt
```diff
@@ -62,9 +61,9 @@ class ZipUtilsTest {
         ZipFile(zipFile.toFile()).use { actualZip ->
             val actualEntry = actualZip.entries().toList().find { it.name == "file.txt" }
 
-            assert(actualEntry).isNotNull()
+            assertThat(actualEntry).isNotNull
```
rli: isNotNull `()`?

jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/NotificationUtilsTest.kt
```diff
@@ -32,11 +29,9 @@ class NotificationUtilsTest {
 
         NullPointerException().notifyError("ooops", project = project)
 
-        assert(notification).isNotNull {
-            assert(it.actual.content) {
-                startsWith("java.lang.NullPointerException")
-                contains("NotificationUtilsTest.kt")
-            }
-        }
+        assertThat(notification).isNotNull
```
rli: same here

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/stack/PageButtonsTest.kt
```diff
@@ -2,8 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package software.aws.toolkits.jetbrains.core.stack
 
-import assertk.assert
-import org.junit.Assert
```
kiiadi: we should create a ktlint rule that detects junit `Assert` (which is an abomination against mankind).

core/tst/software/aws/toolkits/core/utils/ZipUtilsTest.kt
abrooksv: Kotlin sees it as a property not a method

rli: weird but ok

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/stack/PageButtonsTest.kt
abrooksv: We use it in few places so we need to fix those first

core/tst/software/aws/toolkits/core/utils/ZipUtilsTest.kt
abrooksv: It is because it matches boolean Java Bean property syntax for a boolean

PR #975, opened 2019-05-20
Upgrade ktlint to 0.32.0

abrooksv:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/IndexedResources.kt
```diff
@@ -14,13 +14,21 @@ import java.io.DataOutput
...
+            this(
+                resource.type() ?: throw RuntimeException(message("cloudformation.template_index.missing_type")),
+                indexProperties
+                    .asSequence()
+                    .map {
+                        it to try {
+                            resource.getScalarProperty(it)
+                        } catch (e: Exception) {
+                            null
+                        }
```
zhangzhx: Nit: could use `tryOrNull` here?

PR #981, opened 2019-05-21
Add CodeCov to buildSpecs

abrooksv:
When we migrated to build specs in the repo, we lost codecov

## License


abrooksv:
Only gap now is we don't report Windows coverage

Issue #979, opened 2019-05-20
Bug in checkConfiguration leads to EDT assertion


abrooksv:
```
Access is allowed from event dispatch thread only.

com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: EventQueue.isDispatchThread()=false Toolkit.getEventQueue()=com.intellij.ide.IdeEventQueue@a5adf7a
Current thread: Thread[OurIconsCalculating Pool,4,Idea Thread Group] 1222850060
SystemEventQueueThread: Thread[AWT-EventQueue-0 2019.1#IC-191.6183.87 IDEA, eap:false, os:Mac OS X 10.12.6, java-version:JetBrains s.r.o 1.8.0_202-release-1483-b39,6,Idea Thread Group] 841377470
	at com.intellij.openapi.application.impl.ApplicationImpl.assertIsDispatchThread(ApplicationImpl.java:1128)
	at com.intellij.openapi.application.impl.ApplicationImpl.assertIsDispatchThread(ApplicationImpl.java:1117)
	at com.intellij.openapi.fileEditor.impl.FileDocumentManagerImpl.saveAllDocuments(FileDocumentManagerImpl.java:288)
	at com.intellij.openapi.fileEditor.impl.FileDocumentManagerImpl.saveAllDocuments(FileDocumentManagerImpl.java:281)
	at software.aws.toolkits.jetbrains.services.lambda.execution.LambdaRunConfigurationBase.resolveInput(LambdaRunConfigurationBase.kt:51)
	at software.aws.toolkits.jetbrains.services.lambda.execution.local.LocalLambdaRunConfiguration.checkConfiguration(LocalLamdaRunConfiguration.kt:73)
	at com.intellij.execution.impl.RunnerAndConfigurationSettingsImpl.checkSettings(RunnerAndConfigurationSettingsImpl.kt:310)
	at com.intellij.execution.RunnerAndConfigurationSettings.checkSettings(RunnerAndConfigurationSettings.java:116)
	at com.intellij.execution.impl.TimedIconCache.calcIcon(TimedIconCache.kt:71)
```

abrooksv:
`checkConfiguration` should not have to resolve the inputs, it should be good enough the file exists

abrooksv:
Fixed in 1.4

chkothe:
I'm on 1.4, but I'm getting pretty much the same stack trace every now and then.
[...]

PR #984, opened 2019-05-21
Change LocalLambdaRunConfig checkConfiguration to only check file exists

abrooksv:
## Types of changes

## Description
* Changed tests to call check configuration instead of getState since that is the real API used
* Moved to integ package due to needing SAM CLI installed
* Made a checkInput so input values are checked, and not actually read from disk

## Motivation and Context
Since we now save all files before resolving the input file (in case user is editing it), we need a cheaper checkConfiguration call. We should also make it cheaper since the UI calls it a lot to check if a run config has become invalidated.

## Related Issue(s)
#979

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/LambdaRunConfigurationBase.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/LambdaRunConfigurationBase.kt
@@ -57,7 +57,7 @@ abstract class LambdaRunConfigurationBase<T : BaseLambdaOptions>(
     }
 
     protected fun resolveInput() = inputSource()?.let {
-        if (isUsingInputFile() && inputSource()?.isNotEmpty() == true) {
+        if (isUsingInputFile() && it.isNotEmpty()) {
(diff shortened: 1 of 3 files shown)


abrooksv:
I will follow up with similar changes to RemoteLamda

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/LambdaRunConfigurationBase.kt
```diff
@@ -46,6 +47,15 @@ abstract class LambdaRunConfigurationBase<T : BaseLambdaOptions>(
...
+        inputSource()?.let {
+            if (isUsingInputFile() && FileUtil.exists(it)) {
+                return
+            }
+        }
+        throw RuntimeConfigurationError(message("lambda.run_configuration.no_input_specified"))
+    }
+
     protected fun resolveInput() = inputSource()?.let {
         if (isUsingInputFile() && inputSource()?.isNotEmpty() == true) {
```
zhangzhx: Not related to the CR, the `inputSource()?` here could be `it`

PR #1005, opened 2019-06-03
Show welcome page when installing the plugin for the first time.

zhangzhx:
Draft for welcome page. Need to revisit once done the UI review.

## License


zhangzhx:
@kiiadi , it's a prototype for supporting welcome page. As we had a discussion the other day, we should redesign it. We had created all the related issues in Github. Closing this in favor with the new issues.

Issue #950, opened 2019-05-04
Cannot run Helloworld application locally for Java


Snedden:
**Describe the bug**
When ever I try to run the sample Hello world application for java I get an error saying SAM build command failed

**To reproduce**
1. Create new AWS serverless project
  File>New>AWS>AWS serverless application:
      Runtime: java8
      SAM Template : AWS SAM Hello World(Maven)
      Project SDK : 1.8(java version "1.8.0_144")
    

 2. Go to App.java> Run application
![AWSToolKit1](https://user-images.githubusercontent.com/8083991/57185952-325a9580-6ea4-11e9-976b-58a26b92b98b.png)
 3. Enter configs as:
![AWSToolKit2](https://user-images.githubusercontent.com/8083991/57185953-469e9280-6ea4-11e9-9200-761c7fd97dc3.png)

   (Also tried to specify the docker network connected to java image explixitlly)
   ![AWSToolKit3](https://user-images.githubusercontent.com/8083991/57185959-528a5480-6ea4-11e9-9543-cf31e399290f.png)

4.  Click run 

   Errors out saying : Sam build command failed 
  
![image](https://user-images.githubusercontent.com/8083991/57185992-e8be7a80-6ea4-11e9-8c94-f51f99d1d6ec.png)

**Expected behavior**
Function to run locally

**Screenshots**
Also when I right on the template.yml file and run it gives me some other error ,` Exception in thread "main" java.lang.NoClassDefFoundError: org/testng/TestNG`

![image](https://user-images.githubusercontent.com/8083991/57186006-58346a00-6ea5-11e9-996d-ec00de85b864.png)

![image](https://user-images.githubusercontent.com/8083991/57186008-5e2a4b00-6ea5-11e9-9443-e7667cb6a8b7.png)

**Your Environment**
 - OS: Microsoft Windows 10 Home
 - JetBrains' Product: IntelliJ IDEA 2019.1.1 (Ultimate Edition)
 - JetBrains' Product Version: Build #IU-191.6707.61, built on April 16, 2019
 - Toolkit Version: v1.3
 - SAM CLI Version: SAM CLI, version 0.14.2
- JVM/Python Version: OpenJDK 64-Bit Server VM by JetBrains s.r.o
 - JRE: 1.8.0_202-release-1483-b44 amd64

**Additional context**

abrooksv:
Interesting that the TestNG plugin says that the yaml is executable.

[...]

PR #1034, opened 2019-06-13
Fix the bug when project is not located in the same drive of temp fol…

zhangzhx:
…der drive

## Types of changes

## Related Issue(s)
#950 

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaBuilder.kt
```diff
@@ -152,6 +171,17 @@ abstract class LambdaBuilder {
             zipLocation.toPath()
         }
 
+    /**
+     * Returns the build directory of the project. Create this if it doesn't exist yet.
+     */
+    private fun getOrCreateBuildDirectory(project: Project): File =
+        WriteAction.computeAndWait<File, Throwable> {
+            val projectBasePath = project.basePath ?: throw IllegalStateException("Invalid project with no base path!")
```
abrooksv: I think Project.guessProjectDir() may be better

abrooksv: Also use message() for the exception

abrooksv: Actually since we know the module, why are we not using the module content root?

zhangzhx: The implementation of `getBasePath` and `guessProjectDir` are different. `getBasePath` uses `Registry` and looks for `.idea` folder for the project base path. `guessProjectDir` uses modules content roots which as it describes is not always accurate.

I think here we still want to create this folder at the same level to `.idea` instead of in the `HelloWorldFunction` folder which is the model content root.

zhangzhx: Discussed offline. Decided to use the first content root of the module as the root folder of the build.

zhangzhx: > Also use message() for the exception

Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaBuilder.kt
```diff
@@ -152,6 +171,18 @@ abstract class LambdaBuilder {
             zipLocation.toPath()
         }
 
+    /**
+     * Returns the build directory of the project. Create this if it doesn't exist yet.
+     */
+    private fun getOrCreateBuildDirectory(module: Module): File =
+        WriteAction.computeAndWait<File, Throwable> {
```
abrooksv: What is making this a WriteAction?

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -288,3 +288,4 @@ codepipeline.resource.operation.deploy=deployment
 lambda.build.unable_to_locate_project_root=Unable to locate project diretory for Module ''{0}''
 lambda.build.java.unsupported_build_system=Module ''{0}'' is not managed by Maven or Gradle
 lambda.run_configuration.remote.function.tooltip=The name of the AWS Lambda function to use.
+lambda.build.module_with_no_content_root=invalid module ''{0}'' with no content root.
```
abrooksv: Module ''{0}'' lacks a content root

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaBuilder.kt
zhangzhx: Weird thing is that if I don't have this `WriteAction`, a threading exception (see below) will be thrown in [here](https://github.com/aws/aws-toolkit-jetbrains/blob/master/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt#L28).

```
java.lang.Throwable: Do not call synchronous refresh under read lock (except from EDT) - this will cause a deadlock if there are any events to fire.
```

Maybe I'm wrong, but it seems when you are creating files under a module content root, the VFS will be rebuilt, therefore a `WriteAction` is needed here.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
zhangzhx: Done.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/LambdaBuilder.kt
abrooksv: The refresh probably triggers write action in the VFS if it has no entry yet.

Issue #1054, opened 2019-06-24
AWS Toolkit for PyCharm does not find handlers when source roots are not set


chkothe:
**Describe the bug**
When opening a pre-existing SAM project (e.g., the sam init hello-world example) with the AWS Toolkit for PyCharm and creating a launch configuration from the template.yaml, a broken launch configuration that reports the error "Cannot find handler 'app.lambda_handler' in project." is created.

**To reproduce**
1. Create SAM project outside PyCharm (e.g., `sam init --runtime=python3.6`)
2. Open project folder in PyCharm
3. Create new AWS Lambda [Local] launch configuration, choosing the template as the template
4. Observe that the generated launch configuration `[Local] HelloWorldFunction` is defunct, and shows the error "Error: Cannot find handler 'app.lambda_handler' in project." is displayed.

**Expected behavior**
The launch configuration should be runnable, or the error message should point at the likely issue.

**Your Environment**
 - OS: Windows 10 1803 x64
 - JetBrains' Product: PyCharm Professional Edition
 - JetBrains' Product Version: 2019.1.3 (Build #PY-191.7479.30, built on May 29, 2019)
 - Toolkit Version: v1.4
 - SAM CLI Version: 0.17.0
 - JVM/Python Version: Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
 - Docker Desktop Version: Community Edition, Version 2.0.0.3 (31259)
 
**Additional context**
This seems to be because the hello_world folder is not marked as PyCharm Sources Root upon opening an external SAM project for the first time. Probably the toolkit could mark the SAM CodeUri as source root upon first-time opening (especially if there's only one CodeUri occurrence, where it's clearly the right way to fix the situation, and which seems to match what the toolkit sets up when it creates its own projects).

abrooksv:
We should be able to infer it from the requirements.txt parent folder if source roots aren't set

abrooksv:
The source roots come from when we did not use SAM, so it should probably actually be looked into and removed if no longer needed.

zhangzhx:
The fix has been released in the [1.5 eap](https://plugins.jetbrains.com/plugin/11349-aws-toolkit/update/64705). Can you try that out? If you still see this problem, feel free to open another issue.

chkothe:
Great, confirmed that it's working now.

PR #1056, opened 2019-06-24
Enable using the presence of a requirements.txt to determine if a met…

abrooksv:
…hod is a handler in Python

## Types of changes

## Description
This allows us to not rely on source roots but instead matches what SAM build will do, which is look for the requirements.txt at the codeUri location

## Related Issue(s)
#1054

## Testing
New unit tests
Manually tested that removing a source root can still run the Lambda handler

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/python/PythonLambdaBuilder.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/python/PythonLambdaBuilder.kt
@@ -23,8 +23,22 @@ class PythonLambdaBuilder : LambdaBuilder() {
 
     private fun getBaseDirectory(project: Project, virtualFile: VirtualFile): VirtualFile {
         val fileIndex = ProjectFileIndex.getInstance(project)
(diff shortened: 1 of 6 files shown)


abrooksv:
Looking at https://codecov.io/gh/aws/aws-toolkit-jetbrains/commit/52b9952e8dd4674699af2c400923e575b484e9c4, it looks like there is still some test gaps. Will add a negative test case to each test suite tomorrow.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/python/PythonLambdaBuilder.kt
```diff
@@ -43,8 +43,22 @@ class PythonLambdaBuilder : LambdaBuilder() {
...
+            while (dir != null) {
+                if (dir == contentRoot || dir.findChild("requirements.txt") != null) {
+                    return dir
+                }
+
+                dir = dir.parent
+            }
+        }
+
+        throw IllegalStateException("Failed to locate the root of the handler")
```
zhangzhx: This goes to `message`?

PR #1059, opened 2019-06-25
Add Gradle cache locations to build specs, enable 2.0 image

abrooksv:
## Types of changes

## License


abrooksv:
Do not merge, this is testing performance delta

buildspec/windowsTests.yml
```diff
@@ -1,6 +1,10 @@
 version: 0.2
 
+env:
+  variables:
+    CI: true
+
```
zhangzhx: Do we need to explicitly provide `openjdk8` install as well in the build spec?

abrooksv: No, only the Ubuntu 2.0 image uses that

Issue #1064, opened 2019-06-26
PyCharm crashes when creating nodejs project


zhangzhx:
**Describe the bug**
Plugin artifact from webstorm branch doesn't work on PyCharm when creating nodejs project with the following exception on project creation.

```
java.lang.ClassCastException: class com.intellij.openapi.module.UnknownModuleType cannot be cast to class com.intellij.openapi.module.WebModuleTypeBase (com.intellij.openapi.module.UnknownModuleType and com.intellij.openapi.module.WebModuleTypeBase are in unnamed module of loader com.intellij.util.lang.UrlClassLoader @1d16f93d)
	at com.intellij.openapi.module.WebModuleTypeBase.getInstance(WebModuleTypeBase.java:24)
	at software.aws.toolkits.jetbrains.services.lambda.nodejs.NodeJsRuntimeGroup.getModuleType(NodeJsRuntimeGroup.kt:44)
	at software.aws.toolkits.jetbrains.services.lambda.RuntimeGroup$getModuleType$1.invoke(RuntimeGroup.kt:46)
	at software.aws.toolkits.jetbrains.services.lambda.RuntimeGroup$getModuleType$1.invoke(RuntimeGroup.kt:31)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:172)
	at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:132)
	at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:156)
	at kotlin.sequences.SequencesKt___SequencesKt.firstOrNull(_Sequences.kt:129)
	at software.aws.toolkits.jetbrains.services.lambda.RuntimeGroup.getModuleType(RuntimeGroup.kt:46)
	at software.aws.toolkits.jetbrains.ui.wizard.SamProjectBuilder.setupRootModel(SamProjectGeneratorIntelliJShims.kt:51)
	at software.aws.toolkits.jetbrains.ui.wizard.SamProjectGenerator$generateProject$$inlined$runInEdt$lambda$1.compute(actions.kt:77)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$runWriteAction$19(ApplicationImpl.java:1064)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteActionWithClass(ApplicationImpl.java:1020)
[...]

PR #1067, opened 2019-06-26
Fix PyCharm crashes for not returning the correct nodejs module.

zhangzhx:
## Types of changes

## Description
PyCharm doesn't return WebModule correctly from [here](https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/openapi/module/WebModuleTypeBase.java#L24). We use the default module type instead.

## Related Issue(s)
#1064 

## License


jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJsRuntimeGroup.kt
```diff
@@ -41,7 +44,10 @@ class NodeJsRuntimeGroup : SdkBasedRuntimeGroupInformation() {
      */
     override fun runtimeForSdk(sdk: Sdk): Runtime? = null
 
-    override fun getModuleType(): ModuleType<*>? = WebModuleTypeBase.getInstance()
+    // WebModuleType.getInstance() is deprecated in 191, use WebModuleTypeBase#getInstance for version 191+
```
abrooksv: Please use the TODO marker for 191 changes. (FIX_WHEN_MIN_IS_192)

PR #1076, opened 2019-07-04
Re-enable 192 EAP

abrooksv:
## Types of changes

## Testing
`ALTERNATIVE_IDE_PROFILE_NAME=2018.3 ./gradlew integrationTest check`
`ALTERNATIVE_IDE_PROFILE_NAME=2019.2 ./gradlew integrationTest check`

## License

### DIFF
--- a/intellijJVersions.gradle
+++ b/intellijJVersions.gradle
@@ -12,12 +12,24 @@ static def ideProfiles() {
                             ]
                     ]
             ],
-            "2019.1": [
+            "2019.1": [ // TODO: Marker comment: FIX_WHEN_MIN_IS_192
                     "IC": [
                             version: "IC-2019.1",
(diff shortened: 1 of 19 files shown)


intellijJVersions.gradle
```diff
@@ -12,7 +12,7 @@ static def ideProfiles() {
                             ]
                     ]
             ],
-            "2019.1": [
+            "2019.1": [ // TODO: Marker comment: FIX_WHEN_MIN_IS_192
```
zhangzhx: Do we need to mark 183 as well?

abrooksv: We will probably remove 183 at same time as 191

zhangzhx: OK. I think one marker is enough then.

PR #1092, opened 2019-07-16
Simplify explorer service node

abrooksv:
## Description
* Pull data from enum so it is in one place
* Rename serviceName to serviceId since that is more accurate

## Testing
Validated Explorer still works

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractActions.kt
```diff
@@ -16,7 +16,11 @@ import javax.swing.Icon
  *
```
kiiadi: This whole file just looks like formatting.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationExplorerNodes.kt
```diff
@@ -26,9 +27,10 @@ import software.aws.toolkits.jetbrains.utils.TaggingResourceType
 import software.aws.toolkits.jetbrains.utils.toHumanReadable
 import software.aws.toolkits.resources.message
 
-class CloudFormationServiceNode(project: Project) : AwsExplorerServiceRootNode(project, message("explorer.node.cloudformation")) {
-    override fun serviceName() = CloudFormationClient.SERVICE_NAME
-
+class CloudFormationServiceNode(
+    project: Project,
+    explorerService: AwsExplorerService
```
kiiadi: What's the value of passing this in vs having it be part of the `CloudFormationServiceNode` implementation? Is there a time where `explorerService` would be anything other than `AwsExplorerService. CLOUDFORMATION` ?

abrooksv: That....is a great question. Got blinders on and didnt see the obvious thing.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractActions.kt
abrooksv: Ya, will revert

PR #1102, opened 2019-07-22
Implement resource cache - to be used for caching describe calls

kiiadi:
## Types of changes

## Description

Implementation of an `AwsResourceCache` - generic implementation of a cache that allows in-memory storage of AWS service call results - typically used for list/describe calls. Currently supports:
- caching by region/credential ID
- time-based cache expiry
- optionally falling back to stale data in the event of a failed refresh
- optionally forcing a refresh (including updating the cached version) - for use when it's known that a key has changed (e.g. new bucket created)
- customizable cache expiry based on resource type

Future features:
- support for pagination
- alternate storage mechanisms (e.g. on disk)

## Motivation and Context

Several dialogs repeatedly make the same describe or list resources call to populate UI drop-downs etc - this results in unnecessary calls to AWS that are problematic for several reasons:
- additional network load
- calls make get throttled
- delay populating UI for data that is already known

## Testing

New unit tests added to cover cache implementation

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -6,64 +6,121 @@ package software.aws.toolkits.jetbrains.core
...
+/**
+ * Intended to prevent repeated unnecessary calls to AWS to understand resource state.
+ *
+ * Will cache responses from AWS by [AwsRegion]/[ToolkitCredentialsProvider] - generically applicable to any AWS call.
+ */
 interface AwsResourceCache {
-    fun lambdaFunctions(): CompletableFuture<List<LambdaFunction>>
+
+    /**
+     * Get a [resource] either by making a call or returning it from the cache if present and unexpired. Uses the currently [AwsRegion]
```
abrooksv: drop `currently` before region

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -6,64 +6,121 @@ package software.aws.toolkits.jetbrains.core
...
+    override fun <T> getResource(
+        resource: CachedResource<T>,
+        region: AwsRegion,
+        credentialProvider: ToolkitCredentialsProvider,
+        useStale: Boolean,
+        forceFetch: Boolean
+    ): CompletionStage<T> {
+        val key = Key(resource, region, credentialProvider)
+        val future = CompletableFuture<T>()
         ApplicationManager.getApplication().executeOnPooledThread {
```
abrooksv: Should we check for cache hit before off loading the thread? This way its faster since no need to switch threads

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
kiiadi: Done

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
kiiadi: Opted for a simpler implementation for now both on the part of the caller (no behavioural difference between a cache hit and cache miss - just lower latency) and the cache itself (easier handling of the cache map, fewer locking concerns). Whilst I agree that it's potentially more efficient to stay in the same thread to check the map - at this point I don't think the optimization is worth the complexity.

abrooksv: Finneee

kiiadi:
@abrooksv added a guava based base-cache so that we can support more intelligent cache eviction and not be such a memory hog.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
abrooksv: I would put active before the [AwsRegion], current way reads very odd

abrooksv:
Hold off merging till after 1.5 release please

PR #1108, opened 2019-07-25
Adding the ability to have 'view' of cached resources

kiiadi:
## Types of changes

## Description

Builds on #1102 adds the ability to define views on top of `CachedResources` that mean the underlying raw data is only cached once. Views are applied in memory at the point of cache retrieval.

## Testing

Added new tests for the view logic.

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
@@ -34,7 +34,7 @@ interface AwsResourceCache {
      * @param[useStale] if an exception occurs attempting to refresh the resource return a cached version if it exists (even if it's expired). Default: true
      * @param[forceFetch] force the resource to refresh (and update cache) even if a valid cache version exists. Default: false
      */
-    fun <T> getResource(resource: CachedResource<T>, useStale: Boolean = true, forceFetch: Boolean = false): CompletionStage<T>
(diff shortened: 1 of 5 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -107,8 +115,13 @@ class DefaultAwsResourceCache(private val project: Project, private val clock: C
         useStale: Boolean,
         forceFetch: Boolean
     ): CompletionStage<T> {
-        val key = Key(resource, region, credentialProvider)
+
+        if (resource is CachedResourceView<*, T>) {
+            return getResource(resource.underlying, region, credentialProvider, useStale, forceFetch).thenApply { resource.doMap(it as Any) }
```
abrooksv: Why is the .doMap needed? Isn't it transformed at the fetch already?

kiiadi:
@abrooksv have taken your feedback about the interfaces on - now there's a sealed `Resource<T>` class with two possible implementations: 
- an `abstract Cached<T>` which includes `fetch()` & `expiry()`
- a concrete final `class View<Input, Output>` which takes an `underlying: Resource<Input>` and `transform: (Input) -> Output` as arguments

The caller doesn't need to care what type of `Resource<T>` is - the cache implementation takes care of applying the transformations and fetching from cache.

This doesn't specifically address your list-bucket-region concern - I think we can add that as a future enhancement - I don't have the capacity to do it now - but I've made a mention of it in the design posted as part of #1113

Issue #977, opened 2019-05-20
Run configuration for non-template based lamda should be able to set timeout


abrooksv:
If not using a CFN template, I should be able to modify the timeout setting of the Lambda

If debugging, it should always be max so that breakpoints do not kill the Lambda

Issue #976, opened 2019-05-20
Run configuration for non-template based lamda should be able to set memory limit


abrooksv:
If not using a CFN template, I should be able to modify the memory limit of the Lambda

PR #1131, opened 2019-08-01
Support timeout and memory size in run configuration.

zhangzhx:
## Types of changes

## Description
* Refactor the Slider + TextField panel to a reusable component
* Rename the `LambdaRemoteRunSettingsEditorPanel` to `RemoteLambdaRunSettingsEditorPanel` to be consistent with local.
* Add `timeout` and `memorySize` to the `Function` resource when parsing CloudFormation template
* Add `timeout` and `memorySize` widgets in `LocalLambdaRunSettingsEditorPanel`

## Related Issue(s)
#976  #977 

Screen Shot 2019-08-01 at 11 37 27 AM
Screen Shot 2019-08-01 at 11 38 03 AM

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt
```diff
@@ -64,7 +66,8 @@ object SamTemplateUtils {
                         keyValue("Handler", handler)
                         keyValue("CodeUri", codeUri)
                         keyValue("Runtime", runtime.toString())
-                        keyValue("Timeout", "900")
+                        keyValue("Timeout", timeout.toString())
```
kiiadi: how does this work if we hit a break-point while debugging?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/SliderPanel.java
```diff
@@ -0,0 +1,125 @@
...
+// A panel with a slider and text field, of which the slider and text field always synced up.
+public class SliderPanel {
+    private final int min, max, minTick, maxTick, minorTick, majorTick, defaultValue;
+    private final boolean snap;
+
+    private JPanel content;
+    public JSlider slider;
+    public JTextField textField;
+
+    public static SliderPanel lambdaTimeout() {
```
kiiadi: This definitely doesn't live here - the `jetbrains.ui` stuff should be pretty much service agnostic

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt
zhangzhx: I've manually tested running in debug mode by calling `context.getRemainingTimeInMillis()`. Sam local set the timeout to be 10 hours ignoring the actual timeout in the template. So it should be ok in debug mode.

zhangzhx:
> Does this only occur when _not_ running locally from a template?

Yes. When running from template, we disable the timeout and memory size settings same way as handler and runtime.

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/SliderPanel.java
zhangzhx: Moved this to `Lambda` file.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt
kiiadi: Does it only do this in debug mode? If it does it all the time what's the point of setting the value at all?

zhangzhx: It only does it in debug mode. I am waiting for the confirmation from the sam team.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -22,6 +31,25 @@ object Lambda {
...
+    const val MAX_MEMORY = 3008
+    const val MEMORY_INCREMENT = 64
+    const val DEFAULT_MEMORY_SIZE = 128
+    const val MIN_TIMEOUT = 1
+    @JvmField val MAX_TIMEOUT = TimeUnit.MINUTES.toSeconds(15).toInt()
+    @JvmField val DEFAULT_TIMEOUT = TimeUnit.MINUTES.toSeconds(5).toInt()
+}
+
+object LambdaWidgets {
+    fun lambdaTimeout(): SliderPanel =
```
kiiadi: If you annotate these methods with `@JvmStatic` you won't have to grab the `INSTANCE` in the java code they'll appear as top-level static methods.

zhangzhx: Done.

PR #1153, opened 2019-08-09
Adding design for the refactor of our executables to be more extensible

kiiadi:


designs/executables.md
```diff
@@ -0,0 +1,54 @@
...
+
+```kt
+interface Executable
+
+interface AutoResolvable {
+
+    /**
+     * Attempt to automatically resolve the path
+     */
+    fun resolve(): CompletionStage<Path?>
```
kiiadi: Maybe we can remove the `CompletionStage` and make this the callers concern -- thoughts?

rli: I think you could go either way. For consistency across the interfaces, I think it makes more sense keeping it as a CompletionStage.

abrooksv: It would also make it harder to do stuff on the UI thread

kiiadi: +1

abrooksv: Can we enhance the documentation for these.

Like is this only the auto-discover logic, or is this whenever we need the CLI path, we need to use this.

If it is both auto-discover, and get path then having it be async may be very difficult to use. Especially if the usage is guarded by a validate()

designs/executables.md
```diff
@@ -0,0 +1,54 @@
...
+* An ability for parts of the toolkit to be able to determine said path
+
+A CLI may also be able to 
+* Automatically resolve the path of the executable for the customer (e.g. find it locally, or resolve it from some remote location)
+* Determine if the executable is valid (e.g. can be executed, is a valid version etc)
+* Surface version information
+
+These optional features will be determined based on a set of mix-in interfaces.
+
+## Proposed Interfaces
```
awschristou: I would be interested in seeing some information on how these interfaces are used (for example, is there a corresponding manager that I would call to leverage these).

designs/executables.md
```diff
@@ -0,0 +1,118 @@
...
+val manager = ExecutableManager.getInstance()
+when(val executable = manager.getExecutable<SamExecutable>()) {
+  is ExecutableInstance.Executable -> doSomethingWithExecutable(executable.executablePath)
+  is ExecutableInstance.InvalidExecutable -> surfaceErrorToUserAboutInvalidVersion(exectuable.validationError)
+  is ExecutableInstance.UnresolvedExecutable -> promptCustomerToConfigure(executable.resolutionError)
+}
+```
+
+## Open Question
+* How do path's get manually set?
```
awschristou: Can we get all `ExecutableType` from `ExecutableManager` that are not `AutoResolvable`? We could hook this into the toolkit config system somehow, possibly with another interface that helps describe the ExecutableType to the config UI.

Issue #1161, opened 2019-08-14
Toolkit throws unhandled exception "No active credential provider configured" on startup


zhangzhx:
**Describe the bug**
The toolkit throws unhandled exception when 

**To reproduce**
1. Create a new SAM project.
2. Toolkit throws an exception when no active credential is configured.

**Expected behavior**
No exception is thrown.

**Screenshots**
```
software.aws.toolkits.core.credentials.CredentialProviderNotFound: No active credential provider configured
	at software.aws.toolkits.jetbrains.core.credentials.DefaultProjectAccountSettingsManager.getActiveCredentialProvider(ProjectAccountSettingsManager.kt:155)
	at software.aws.toolkits.jetbrains.core.credentials.ProjectAccountSettingsManagerKt.activeCredentialProvider(ProjectAccountSettingsManager.kt:115)
	at software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerNode.<init>(AwsExplorerNode.kt:22)
	at software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerServiceRootNode.<init>(AwsExplorerServiceRootNode.kt:15)
	at software.aws.toolkits.jetbrains.core.explorer.nodes.CacheBackedAwsExplorerServiceRootNode.<init>(AwsExplorerServiceRootNode.kt:24)
	at software.aws.toolkits.jetbrains.services.cloudformation.CloudFormationServiceNode.<init>(CloudFormationExplorerNodes.kt:33)
	at software.aws.toolkits.jetbrains.core.explorer.AwsExplorerService$CLOUDFORMATION.buildServiceRootNode(AwsExplorerService.kt:18)
	at software.aws.toolkits.jetbrains.core.explorer.AwsExplorerService$CLOUDFORMATION.buildServiceRootNode(AwsExplorerService.kt:17)
	at software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerRootNode.getChildren(AwsExplorerRootNode.kt:22)
	at software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerRootNode.getChildren(AwsExplorerRootNode.kt:16)
	at com.intellij.ide.util.treeView.AbstractTreeStructureBase.getChildElements(AbstractTreeStructureBase.java:37)
	at software.aws.toolkits.jetbrains.ui.tree.StructureTreeModel.getValidChildren(StructureTreeModel.java:414)
[...]

PR #1162, opened 2019-08-14
Fix the bug that toolkit throws unhandled exception on startup when a…

zhangzhx:
…ctive credential is not configured

## Types of changes

## Related Issue(s)
#1161 

## Testing
Manually tested it.

## License


kiiadi:
@zhangzhx looks like this might have introduced a blinking test - see the travis failure...

zhangzhx:
> @zhangzhx looks like this might have introduced a blinking test - see the travis failure...

The test failure is due to the inaccurate mock, specifically `MockResourceCache`. We want to pass in an invalid credentialProvider but in this case, it will always return a valid one.

The test failure is subtle though that if it runs fast enough, it will always pass. We should wait for the change event to complete to make sure the active credential update has been finished.

PR #1166, opened 2019-08-15
Ripped out Guava cache implementation to avoid AssertionError

kiiadi:
Replaced with `ConcurrentHashMap` which is periodically pruned to ensure memory remains managable.

## Types of changes

## Description

Removed Guava cache implementation. New implementation is a basic Java `ConcurrentHashMap` with a periodically scheduled job that prunes entries once the map exceeds a certain number of (configurable) entries - currently set to 1000. Entries are sorted by expiration with those closest to expiry being removed.

Entries are given a weighting (rather than just a straight number) in an attempt to size them, collection based entries use the size of the collection as the weight, all others are 1.

## Motivation and Context

Guava cache implementation randomly throws `AssertionError` exceptions on some cache accesses (as part of the cache eviction process) when using `asMap()`

See https://github.com/google/guava/issues/3464

## Testing

New unit tests added.

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
@@ -3,10 +3,11 @@
 
 package software.aws.toolkits.jetbrains.core
 
-import com.google.common.cache.CacheBuilder
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.project.Project
+import com.intellij.util.Alarm
+import com.intellij.util.AlarmFactory
 import software.amazon.awssdk.core.SdkClient
 import software.aws.toolkits.core.credentials.ToolkitCredentialsChangeListener
 import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
@@ -20,6 +21,7 @@ import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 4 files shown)


core/src/software/aws/toolkits/core/utils/CollectionUtils.kt
```diff
@@ -0,0 +1,10 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.core.utils
+
+fun <T> List<T>.tail(): List<T> = if (size > 1) {
```
abrooksv: Should be the same as kotlin .drop(1)

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -199,28 +208,56 @@ class DefaultAwsResourceCache(private val project: Project, private val clock: C
...
-                }
             } catch (e: Throwable) {
                 context.future.completeExceptionally(e)
             }
         }
     }
 
+    private fun runCacheMaintenance() {
+        try {
+            val entries = cache.entries.toList()
```
abrooksv: Does it make sense to go entries to sequence do the sum onEach and then toList?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -199,28 +208,56 @@ class DefaultAwsResourceCache(private val project: Project, private val clock: C
...
+        val nowRemoved = if (removeIfStillSame(head)) {
+            removedSoFar + head.value.weight
+        } else {
+            removedSoFar
+        }
+        prune(entries.tail(), nowRemoved, targetRemoval)
+    }
+
+    private fun removeIfStillSame(head: Map.Entry<CacheKey, Entry<*>>): Boolean =
+        cache.computeIfPresent(head.key) { _, currentEntry -> if (currentEntry == head.value) null else currentEntry } == null
```
abrooksv: This should be an explicit identity check instead of ==

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
kiiadi: yeah that probably is more efficient, we'd only be traversing the list once. more mutation - *sigh*

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
kiiadi: yup - updated

PR #1176, opened 2019-08-22
Explorer action tidy up

kiiadi:
## Motivation and Context

Couple of things that were just bugging me about the explorer:
- ECS Task Definitions didn't have a Copy-Arn
- ECS Task Definitions were showing inactive items
- Copy Arn didn't have an icon
- Delete Stack didn't have a separator and included the service name (whereas lambda delete function did not)

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/CopyArn.kt
```diff
@@ -3,14 +3,15 @@
...
 
+import com.intellij.icons.AllIcons
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.ide.CopyPasteManager
 import software.aws.toolkits.resources.message
 import java.awt.datatransfer.StringSelection
 
-class CopyArn(private val arn: String) : AnAction(message("explorer.copy_arn")) {
+class CopyArn(private val arnProvider: () -> String) : AnAction(message("explorer.copy_arn"), null, AllIcons.Actions.Copy) {
```
abrooksv: So crazy theory here...

```
class CopyArn : SingleResourceNodeAction<AwsExplorerResourceNode<*>>(message("explorer.copy_arn")) {
    override fun actionPerformed(selected: AwsExplorerResourceNode<*>, e: AnActionEvent) {
        CopyPasteManager.getInstance().setContents(StringSelection(selected.resourceArn()))
    }
}
```

We can then stick CopyArn into plugin.xml

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -280,7 +280,7 @@ sam.build.title=SAM build
 sam.build.running=Running SAM build
 sam.build.failed=SAM build command failed
 sam.build.succeeded=SAM build command succeeded
-cloudformation.stack.delete.action=Delete CloudFormation Stack
+cloudformation.stack.delete.action=Delete Stack
```
awschristou: Please add ellipses (`Delete Stack...`) to indicate that further prompting/interaction will happen after selecting this item.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/CopyArn.kt
kiiadi: Only problem with that is that you then have to put it in the plugin.xml for each resource.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
kiiadi: Good catch

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/CopyArn.kt
abrooksv: Ya, but we were just complaining about how we don't like the ordering, so we would be more consistent and have more control

kiiadi: What if we did the opposite - and created a new marker interface `DeleteableResource` that our menu creation can add automatically, we can then control the location of delete in the menu too.

PR #1198, opened 2019-09-06
Add completion for Lambda handler text field in run configuration editor

sdubov:
## Types of changes

## Description
Add completion for Lambda Handler text field in the Lambda run configuration editor.

- Extend frontend logic to set text field handler completion per language and runtime group
- Add backend logic to find all possible lambda handlers in a solution to provide them in a completion lookup
- Move logic to get Amazon Lambda handlers into a static Util class
- Add docs for Lambda protocol model

## Motivation and Context
Completion makes it easier to fill handler string in Lambda run configuration. Completion contains all possible handler strings from all projects inside a solution.

## Related Issue(s)
None.

## Testing
Environment:
```
Mac OS Mojave 10.14.5 (18F132)
```
- Verified handlers from multiple projects are available in completion list
- Verified completion popup speed on huge solution with ~200 projects

LambdaCompletion

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/DefaultHandlerCompletion.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.intellij.openapi.project.Project
+import software.aws.toolkits.jetbrains.services.lambda.LambdaHandlerIndex
+
+class DefaultHandlerCompletion : HandlerCompletion {
+
+    override fun getPrefixMatcher(prefix: String): PrefixMatcher =
+        PlainPrefixMatcher(prefix)
+
+    override fun getLookupElements(project: Project): Collection<LookupElement> =
+        LambdaHandlerIndex.listHandlers(project).map { LookupElementBuilder.create(it) }
```
kiiadi: FYI the index was removed in master a while back see https://github.com/aws/aws-toolkit-jetbrains/pull/1107

We'll have do to RuntimeGroup specific implementations most likely.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        val completion = HandlerCompletion.getInstance(runtimeGroup)
+        if (completion == null) {
+            val message = "Unable to get HandlerCompletion instance for Lambda handler completion provider"
+            logger.error { message }
+            throw RuntimeException(message)
+        }
+
+        completion!!
```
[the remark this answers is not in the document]
sdubov: Sure. Fixed. Rewrite with inline check for null value.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/DefaultHandlerCompletion.kt
sdubov: Hm. Does that mean you expect no completion in lambda run configuration anymore? Do we expect to support completion in Rider or is it gonna be consistent behavior per all products?

abrooksv: We had to remove the index because in 2019.2 we were getting re-entrant indexing errors. Due to the tight deadlines of getting support out it was removed. I am not against bringing back auto-complete, but could it be done with:

1. A better index written by someone knowing what they are doing
2. Just loading completion candidates by filtering the method indexes per runtime group.. (Not sure what performance would look like here...)

We technically never had it working for Python, so we were already a bit inconsistent, but ideally that should be an anti-pattern.

sdubov: @abrooksv, @kiiadi, got it. Thank you.

I could take a look on this issue in Idea and Python. Just a note: I'm using the second approach for Rider since we have all indexes on a backend. I've checked on a solution with ~250 project (+ one SAM Lambda project), the completion popup is showed up almost immediately with a top-level filtering by "lambda-defined" project type.

I'm just wondering what we should do with this commit now :) It makes not a lot of sense after modifications in master. Do you think we should freeze it until master is updated?

abrooksv: 2nd approach is fine if its fast. The completion system is async anyway so it is not like we block the UI.

I do like this, but maybe it makes sense to get master in and then restore the functionality in this branch. @kiiadi Thoughts?

sdubov:
Merged the logic to enable completion for handler field per language and runtime in master. I've created a separate PR from latest updates with all necessary logic to enable completion for DotNet - please see https://github.com/aws/aws-toolkit-jetbrains/pull/1247. Closing the original PR since it is outdated now.

PR #1200, opened 2019-09-06
Add EditorConfig to project

rli:
Based on https://github.com/aws/aws-sdk-java-v2/blob/master/build-tools/src/main/resources/software/amazon/awssdk/intellij-codestyle.xml

## License


orozcoadrian:
@rli just curious: did you add all the newlines at the end of files through some "auto-formatting"? I can get intellij to auto-format some settings like `indent_size` but i can't get it to add a newline at the end with `insert_final_newline = true` (i've tried using all the settings in this PR). thanks.

rli:
@orozcoadrian ktlint has support for `insert_final_newline` in `.editorconfig`files, so I just ran ktlint with the auto-format option enabled (`-F`) and manually fixed anything it couldn't handle automatically.

PR #1209, opened 2019-09-11
Adding two new ktlint rules to look for print/println and !!

kiiadi:


jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployTest.kt
```diff
@@ -160,7 +163,7 @@ class SamDeployTest {
                                 object : OutputListener() {
                                     override fun onTextAvailable(event: ProcessEvent, outputType: Key<*>) {
                                         super.onTextAvailable(event, outputType)
-                                        println("SAM CLI: ${event.text}")
+                                        LOG.info { "SAM CLI: ${event.text}" }
```
abrooksv: This was written this way so the output shows in the gradle logs.

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployTest.kt
```diff
@@ -182,8 +185,12 @@ class SamDeployTest {
                     it.stackName(stackName)
                 }
             } catch (e: Exception) {
-                println("Failed to delete stack $stackName: ${ExceptionUtil.getMessage(e)}")
```
abrooksv: Again, it was like this so we could know something is wrong with our IAM permissions. No one reads the idea.log in CodeBuild....

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployTest.kt
kiiadi: how do we fix it so that we get log statements in gradle logs?

abrooksv: I THINK we can add a log4j.xml (v1 format) to a test resources and have that log to console.

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/python/PythonLocalLamdaRunConfigurationIntegrationTest.kt
```diff
@@ -64,7 +64,7 @@ class PythonLocalLamdaRunConfigurationIntegrationTest(private val runtime: Runti
             import os
 
             def lambda_handler(event, context):
-                print(os.environ)
+                print os.environ
```
rli: Is this okay? Looks like we set some breakpoints and verify output for all the python runtimes we support.

justinmk3:
Should the branch be deleted?

PR #1211, opened 2019-09-12
Fix error with find hander protocol calls when creating new lambda

sdubov:
* Optimized logic to get handlers inside a solution
* Update call for find handler logic to sync call to make sure response is updated on a correct thread

## Types of changes

## Description
Fix issue with creating a new AWS Lambda and validating a handler inside a configuration panel.

## Motivation and Context
The main purpose for this commit is to fix the issue with validating a handler in a backend PSI while creating a new AWS Lambda. This issue prevents from creating a Lambda because of panel validation error.

## Related Issue(s)
None.

## Testing
Checked new AWS Lambda creation logic.

None.

## Checklist

## License


jetbrains-rider/ReSharper.AWS/src/AWS.Daemon/Lambda/LambdaFinder.cs
```diff
@@ -0,0 +1,336 @@
...
+using JetBrains.ProjectModel.ProjectsHost.Dependencies;
+using JetBrains.ReSharper.Psi;
+using JetBrains.ReSharper.Psi.CSharp;
+using JetBrains.ReSharper.Psi.Resolve;
+using JetBrains.ReSharper.Psi.Util;
+using JetBrains.UI.ThemedIcons;
+using JetBrains.Util;
+using JetBrains.Util.Logging;
+
+namespace ReSharper.AWS.Lambda
```
abrooksv: What tests covers this logic?

abrooksv:
Production code looks good, just question around what tests hit that logic

jetbrains-rider/ReSharper.AWS/src/AWS.Daemon/Lambda/LambdaFinder.cs
sdubov: This is a pease of refactoring for defining lambda handlers. This logic is used LambdaGutterMarkHighlighting and is covered by LambdaGutterMarkHighlightingTest. I left the duplicate code in the original commit. Removed it + removed the unused method to  collect all handlers from backend logic to not mix it for now. Sorry for confusion. Pushed an updated.

abrooksv: Okay, sounds good.

PR #1218, opened 2019-09-17
Deprecate "namespace" in telemetry calls

rli:
"Namespace" as a concept doesn't mesh well with our team's new naming strategy for telemetry events. This change marks all references to "namespace" as deprecated and allows building telemetry events that are not tied to a top-level "namespace".

In the future, we should:
* have all of our metric names in a schematized format that can be shared across our products.
* remove the `ActionWrappers` and `LoggingDialogWrapper` system
* migrate the existing telemetry names into the new format

## Checklist

## License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,9 +136,6 @@ subprojects {
         tasks.withType(JavaCompile).all {
             options.compilerArgs << "-Werror"
         }
-        tasks.withType(KotlinCompile).all {
-            kotlinOptions.allWarningsAsErrors = true
-        }
     }
 
     task integrationTest(type: Test) {
--- a/core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
+++ b/core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
@@ -10,13 +10,15 @@ import java.time.Instant
 import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
 
 interface MetricEvent {
-    val namespace: String
+    @Deprecated("namespace should not be used")
+    val namespace: String?
     val createTime: Instant
     val awsAccount: String
     val awsRegion: String
     val data: Iterable<Datum>
 
     interface Builder {
+        @Deprecated("MetricEvent should not have a namespace")
         fun namespace(namespace: String): Builder
 
         fun createTime(createTime: Instant): Builder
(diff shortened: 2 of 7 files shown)


core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
```diff
@@ -10,13 +10,15 @@ import java.time.Instant
 import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
 
 interface MetricEvent {
-    val namespace: String
+    @Deprecated("namespace should not be used")
```
rli: Alternatively, should we deprecate this whole file and have the TelemetryService record individual `Datum` that are not encapsulated by a `MetricEvent`?

build.gradle
```diff
@@ -136,9 +136,6 @@ subprojects {
         tasks.withType(JavaCompile).all {
             options.compilerArgs << "-Werror"
         }
-        tasks.withType(KotlinCompile).all {
```
hunterwerlla: Was this accidentally removed?

bryceitoc9:
I'm assuming this won't affect any existing metric names? If so, would we just not be providing namespaces to new metrics?

build.gradle
rli: No, since deprecation result in warnings we can no longer fail the build

rli:
> I'm assuming this won't affect any existing metric names? If so, would we just not be providing namespaces to new metrics?

That's correct

core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
awschristou: What does it mean to have more than one datum with the same createTime?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
```diff
@@ -58,9 +58,12 @@ class DefaultTelemetryPublisher(
     private fun Collection<MetricEvent>.toMetricData(): Collection<MetricDatum> = this
         .flatMap { metricEvent ->
             metricEvent.data.map { datum ->
+                val metricName = metricEvent.namespace?.let {
+                    "$it.${datum.name}"
```
awschristou: `it.` or `it_` ?

core/src/software/aws/toolkits/core/telemetry/MetricEvent.kt
hunterwerlla: Is there a drawback of recording individual Datum?

rli: Data with the same createTime results in each datum with the same timestamp when recorded in the back-end. I see the whole class as a convenience, so there is no downside in recording datum individually.

awschristou: I'm not seeing the value in sending Datum with a matching timestamp. I'm neutral on the initial question.

rli: If we're planning on eventually codegen-ing our metric classes, we're going to throw this whole file away anyways, so I'm inclined to merge as-is.

PR #1229, opened 2019-09-24
Add extension to enable completion in lambda run configuration per language

sdubov:
## Types of changes

## Description
* Add an extension to be able to enable completion in Local Lambda run configuration handler string field
* Add default disabled compleltion handler implementation for JAVA and PYTHON
* Update tests

## Motivation and Context
Extend the ability to provide completion support inside run configurations with handler text field. This will allow to apply completion option per language (e.g. Rider have a logic to walk through PSI, find a valid lambda handlers and file a value into handler field). This will simplify composing the handler text field.

## Related Issue(s)
None.

## Testing
Verified forms: Create new Lambda, Run local lambda and make sure forms remains unchanged since completion support is disabled for JAVA and PYTHON languages.

None.

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaHandlerCompletion.kt
```diff
@@ -0,0 +1,18 @@
...
+package software.aws.toolkits.jetbrains.services.lambda.java
+
+import com.intellij.codeInsight.completion.PrefixMatcher
+import com.intellij.codeInsight.lookup.LookupElement
+import com.intellij.openapi.project.Project
+import software.aws.toolkits.jetbrains.services.lambda.completion.HandlerCompletion
+
+class JavaHandlerCompletion : HandlerCompletion {
+
+    override fun isSupported(): Boolean = false
```
kiiadi: Is there any point having these? Can't we just not implement them for the JAVA / PYTHON run times - and their 'absence' determines whether or not they're supported?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
```diff
@@ -0,0 +1,70 @@
...
+import software.aws.toolkits.jetbrains.services.lambda.RuntimeGroup
+import software.aws.toolkits.jetbrains.services.lambda.runtimeGroup
+import java.lang.IllegalStateException
+
+class HandlerCompletionProvider(private val project: Project) : TextCompletionProvider {
+
+    private val logger = getLogger<HandlerCompletionProvider>()
+
+    // Suppress since gradle compiler mismatch return types for property delegate and fail build step
+    private val handlerCompletion: HandlerCompletion by lazy {
```
kiiadi: As per the above comment - I'd probably have this just be `nullable` and if there was no handler registered for the given language - then it'll act like it's 'not supported'

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaHandlerCompletion.kt
sdubov: Updated.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
sdubov: Updated.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            project
+        )?.runtimeGroup
+        if (runtimeGroup == null) {
+            val message = "Unable to define Runtime Group for Lambda handler completion provider"
+            logger.error { message }
+            throw IllegalStateException(message)
+        }
+
+        return@lazy HandlerCompletion.getInstance(runtimeGroup) ?: let {
+            val message =
```
kiiadi: probably doesn't need the assignment here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+import software.aws.toolkits.jetbrains.services.lambda.runtimeGroup
+import java.lang.IllegalStateException
+
+class HandlerCompletionProvider(private val project: Project) : TextCompletionProvider {
+
+    private val logger = getLogger<HandlerCompletionProvider>()
+
+    // Suppress since gradle compiler mismatch return types for property delegate and fail build step
+    private val handlerCompletion: HandlerCompletion? by lazy {
+        val runtimeGroup = RuntimeGroup.determineRuntime(
```
kiiadi: I assume this means we only support one type of handler per project - so if they have different source languages then they'd only get one - which is probably fine.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
sdubov: Fixed.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/completion/HandlerCompletionProvider.kt
sdubov: @kiiadi, do you mean the case when you have different files with python and java that both define lambda handlers inside one project?

kiiadi: yes

sdubov: @kiiadi, could you pleas take a look on a new commit? I've updated the logic to make HanderCompletion runtime dependent. In run configuration, it will be re-created every time new runtime is selected.

sdubov:
@kiiadi, could you pleas take a look on a new commit? I've updated the logic to make HanderCompletion runtime dependent. In run configuration, it will be re-created every time new runtime is selected.

sdubov:
@kiiadi, can we merge this RP in master? I do not have merge access.

PR #1241, opened 2019-10-02
Added support for 2019.3 based on the latest EAP release.

kiiadi:
## Types of changes

## Description

Adding required compatibility utils to be able to support 2018.3 through 2019.3


jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CompatibilityUtils.kt
```diff
@@ -23,4 +40,59 @@ object CompatibilityUtils {
...
+            return method.invoke(null, path) as Project
+        }
+        val legacyCreate = PlatformTestCase::class.java.getMethod("createProject", File::class.java, String::class.java)
+        return legacyCreate.invoke(null, path.toFile(), "Fake") as Project
+    }
+
+    /**
+     * FIX_WHEN_MIN_IS_192 WebModuleType.getInstance() is deprecated in 191, use WebModuleTypeBase#getInstance for version 191+
+     */
+    @TestOnly
```
hunterwerlla: This is called from `NodeJsRuntimeGroup.kt` so it's definitely not `@TestOnly`

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CompatibilityUtils.kt
```diff
@@ -23,4 +40,59 @@ object CompatibilityUtils {
     @JvmStatic
     fun createPopupBuilder(validationInfo: ValidationInfo, configurator: Consumer<JEditorPane>?): ComponentPopupBuilder? =
         ComponentValidator::class.staticFunctions.find { it.name == "createPopupBuilder" }?.call(validationInfo, configurator) as? ComponentPopupBuilder
+
+    /**
+     * Can be removed when min-version is 19.1 FIX_WHEN_MIN_IS_191
```
hunterwerlla: All things that are actually `FIX_WHEN_MIN_IS_191` are currently `FIX_WHEN_MIN_IS_192` to have a single string to look for

kiiadi: ? the idea here would be that there's a common pattern string to look for when we move off a version ... not sure what you mean

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CompatibilityUtils.kt
kiiadi: yes - good catch

PR #1242, opened 2019-10-03
Awschristou/change notification

awschristou:
## Types of changes

## Description

General Feature:
* A reusable notify-on-startup system. Notices can be registered as extension points, containing their own logic around if a notice needs to be shown.

Concrete Usage:
* A notice is shown on startup for people using IDE versions older than 2019.2. The notice indicates that an upcoming toolkit release will require 2019.2 or newer.

Not in this Change:
* A Toolkit Settings facility to reset all silenced notifications

image

This is what it looks like when more than one notice appears
image

## Testing

Using the notification about needing 2019.2 in a future release...
* Ran IntelliJ 2019.2, notification did not appear.
* Ran IntelliJ 2019.1, saw notification appear.
  * Pressing "Don't show again" and restarting the IDE resulted in the notification not appearing again

## Checklist

## License


resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -308,3 +308,7 @@ aws.settings.executables.find.description=Select path to {0} executable
 aws.settings.executables.cannot_determine_version=Cannot determine version of {0}: {1}
 aws.settings.executables.executable_invalid=Invalid {0} executable: {1}
 aws.settings.executables.resolution_exception=Exception occurred attempting to resolve {0} executable: {1}
+
+notice.suppress=Don't show this notification again
```
kiiadi: I'd shorten this to "Don't show this again"

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -308,3 +308,7 @@ aws.settings.executables.find.description=Select path to {0} executable
...
+notice.title.jetbrains.minimum.version.2019.2=AWS Toolkit Requirements Notice
+notice.message.jetbrains.minimum.version.2019.2=An upcoming release of the AWS Toolkit will require version 2019.2 or newer of the JetBrains IDE.
```
kiiadi: Can we get smart about the IDE name? I think it would be clearer to customers if we said something like "require version 2019.2 of IntelliJ" /PyCharm/etc; should be able to use `ApplicationNamesInfo.getInstance().fullProductName`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeManager.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    private fun notify(notice: NoticeType, project: Project) {
+        val notification = notificationGroup.createNotification(
+            notice.getNoticeContents().title,
+            notice.getNoticeContents().message,
+            NotificationType.INFORMATION,
+            null
+        )
+
+        notification.addAction(
```
kiiadi: like it!

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        return true
+    }
+
+    override fun getNoticeContents(): NoticeContents = noticeContents
+}
+
+data class NoticeContents(var title: String, var message: String)
```
kiiadi: can these be `val` ? or are they persisted somewhere?

awschristou: These are sent to the Notification balloon for display to the user so they are required. They aren't persisted.

awschristou:
> Do you think we should have a standard "title" for all AWS Toolkit notices? Something like "AWS Toolkit Notification"
> 

I think there's a greater chance of different notice reasons. For example, this would be a good use of the "first time" notification to enable/disable telemetry settings. A more relevant title seems appropriate.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
awschristou: I like this suggestion, I'm taking a look

awschristou: Coming soon:
image

awschristou: image

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
kiiadi: Ok let's make them val then since it's immutable

awschristou: Oops, `var` vs `val` was an oversight. I mis-read your initial question as `val?`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeManager.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            true
+        }
+
+    override fun notify(notices: List<NoticeType>, project: Project) {
+        notices.forEach { notify(it, project) }
+    }
+
+    private fun notify(notice: NoticeType, project: Project) {
```
rli: also take a look at `NotificationUtils`, which already has some of this logic

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.notification
+
+import com.intellij.openapi.application.ApplicationInfo
+import com.intellij.openapi.application.ApplicationNamesInfo
+import com.intellij.openapi.extensions.ExtensionPointName
+import software.aws.toolkits.resources.message
+
+interface NoticeType {
```
bryceitoc9: Should we allow for additional actions? For instance, for the example case, you may want a link to upgrade the IDE.

awschristou: That's a good thought. I would consider this out of scope for this change, but `NoticeType` could certainly expose something like `getActions` which the Manager could add to the notification that is produced.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeManager.kt
awschristou: I decided to leave this as is (for now) so that I have the notification object for use with the `notification.hideBalloon()` call

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
bryceitoc9: I'm on the fence about considering this out-of-scope for the change, since the task is about writing the framework for these changes as well.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    override fun isNotificationRequired(): Boolean {
+        val appInfo = ApplicationInfo.getInstance()
+        val majorVersion = appInfo.majorVersion.toIntOrNull()
+        val minorVersion = appInfo.minorVersion.toIntOrNull()
+
+        majorVersion?.let {
+            minorVersion?.let {
+                return majorVersion < 2019 || (majorVersion == 2019 && minorVersion < 2)
```
bryceitoc9: Will updating the min version involve upping these numbers? Can we pull from the plugin configuration instead?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    companion object {
+        val EP_NAME = ExtensionPointName<NoticeType>("aws.toolkit.notice")
+
+        internal fun notices(): List<NoticeType> = EP_NAME.extensions.toList()
+    }
+}
+
+class JetBrainsMinimumVersionChange : NoticeType {
+    override val id: String = "JetBrainsMinimumVersion_192"
```
bryceitoc9: Will any new notifications for newer versions require us to manually up this value? Not a huge fan of that personally; it adds a lot of moving parts.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val noticeContents = NoticeContents(
+        message("notice.title.jetbrains.minimum.version.2019.2"),
+        message("notice.message.jetbrains.minimum.version.2019.2", ApplicationNamesInfo.getInstance().fullProductName)
```
bryceitoc9: Will updating the min version involve upping these numbers? Can we pull from the plugin configuration instead?

(also goes for the line above)

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun getNoticeContents(): NoticeContents
+
+    companion object {
+        val EP_NAME = ExtensionPointName<NoticeType>("aws.toolkit.notice")
+
+        internal fun notices(): List<NoticeType> = EP_NAME.extensions.toList()
+    }
+}
+
+class JetBrainsMinimumVersionChange : NoticeType {
```
bryceitoc9: Move this to a separate file

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
kiiadi: YAGNI - it'll be easy to add in a backwards compatible way in the future - I tend to agree with @awschristou that it's not a requirement for this change. We've done as much of the generic version as we needed for our requirement and left it open for other notices to extend in the future.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
kiiadi: Problem is we want the notice before we do the bump.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/NoticeType.kt
kiiadi: agree

PR #1244, opened 2019-10-03
Fixing problem with node integration tests running on 2019.2

kiiadi:


abrooksv:
Destory iit on after class?

kiiadi:
> Destory iit on after class?

There isn't really a shutdown/destroy/close method.

PR #1248, opened 2019-10-03
Fixing problem with the ProjectRule lifecycle which was causing an NPE

kiiadi:


jetbrains-ultimate/it/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJsLocalLambdaRunConfigurationIntegrationTest.kt
```diff
@@ -220,6 +215,13 @@ class NodeJsLocalLambdaRunConfigurationIntegrationTest(private val runtime: Runt
         assertThat(debuggerIsHit.get()).isTrue()
     }
 
+    private fun ensureServerStarted() {
+        if (!serverStarted.getAndSet(true)) {
```
awschristou: If `waitForStart` is slow, concurrent runs will pass this part, which defeats your intent.

kiiadi: Good catch! fixed

PR #1249, opened 2019-10-04
Fixing versioning with the 2019 so that we're not hitting a moving target

kiiadi:


intellijJVersions.gradle
```diff
@@ -127,6 +127,9 @@ def resolveIdeProfileName() {
     return project.ideProfileName
 }
 
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
awschristou: What's this?

kiiadi: that's IDEA being a jerk - haha.

PR #1258, opened 2019-10-09
Updating version notice, and fixing version detection logic

kiiadi:
## Types of changes

## Description

Adding deprecation to the title of the message for a stronger call to action.

Screen Shot 2019-10-09 at 12 58 09 PM
Screen Shot 2019-10-09 at 1 06 39 PM

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/JetBrainsMinimumVersionChange.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/JetBrainsMinimumVersionChange.kt
@@ -13,6 +13,7 @@ class JetBrainsMinimumVersionChange : NoticeType {
         message("notice.title.jetbrains.minimum.version.2019.2"),
         message(
             "notice.message.jetbrains.minimum.version.2019.2",
+            ApplicationInfo.getInstance().fullVersion,
             ApplicationNamesInfo.getInstance().fullProductName
         )
     )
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/notification/JetBrainsMinimumVersionChange.kt
```diff
@@ -29,7 +30,7 @@ class JetBrainsMinimumVersionChange : NoticeType {
     override fun isNotificationRequired(): Boolean {
         val appInfo = ApplicationInfo.getInstance()
         val majorVersion = appInfo.majorVersion.toIntOrNull()
-        val minorVersion = appInfo.minorVersion.toIntOrNull()
+        val minorVersion = appInfo.minorVersion.toFloatOrNull()
```
justinmk3: Strange. Was that a JB/legacy/historical choice , or why was a float used?

kiiadi: It's actually a string "2.2" which obviously didn't parse as an `Int`

kiiadi: the `toFloatOrNull` is a kotlin built-in extension function on `String`

justinmk3: In a version string x.y.z., "y" is the minor version and "z" is the "patch", and both could be integers then.

PR #1284, opened 2019-10-18
Tweak the build

abrooksv:
* Remove hamcrest, we dont use it
* Remove the check for test modules, we removed testutils
* Fix runIde for Ultimate, when ran through IntelliJ always skipping

## License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -96,11 +96,9 @@ subprojects {
     dependencies {
         compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
         compile "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
-        compile "org.jetbrains.kotlin:kotlin-script-runtime:$kotlinVersion"
         testCompile "com.nhaarman.mockitokotlin2:mockito-kotlin:2.0.0"
         testCompile "org.assertj:assertj-core:$assertjVersion"
         testCompile "junit:junit:$junitVersion"
-        testCompile "org.hamcrest:hamcrest-all:1.3"
     }
 
     testlogger {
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -234,28 +232,31 @@ task ktlint(type: JavaExec, group: "verification") {
...
 }
 check.dependsOn coverageReport
 
 // Workaround for runIde being defined in multiple projects, if we request the root project runIde, "alias" it to
 // community edition
 if (gradle.startParameter.taskNames.contains("runIde")) {
+    // Only disable this if running from root project
+    if (gradle.startParameter.projectDir != project.rootProject.rootDir) {
+        return
+    }
```
justinmk3: seems surprising to return from a top-level scope, but I assume this works somehow. (in Lua this would end the script, I guess gradle somehow skips the `if` scope and continues evaluating the rest of the script?)

abrooksv: In Groovy, we are inside a Lambda so it just exits back to main scope

abrooksv: It's odd looking, ill invert and wrap

build.gradle
```diff
@@ -234,40 +232,41 @@ task ktlint(type: JavaExec, group: "verification") {
...
-    println("Top level runIde selected, excluding sub-projects' runIde")
-    gradle.taskGraph.whenReady { graph ->
-        graph.allTasks.forEach {
-            if (gradle.startParameter.systemPropertiesArgs.getOrDefault("exec.args", "").contains("guitest")) {
-                if (it.name == "runIde" && it.project != project(':jetbrains-core-gui')) {
-                    it.enabled = false
-                }
-            } else {
-                if (it.name == "runIde" &&
+    // Only disable this if running from root project
```
hunterwerlla: Uh that is not the UI I was expecting after making that comment, commenting on the most recent version of the PR:
nit: this comment is now backwards

hunterwerlla: Nevermind, the comment is correct, the logic is just kind of weird

kiiadi:
how do you feel about introducing a `--skipRider` flag that I can pass to a build to omit the rider stuff altogether; to lower the barrier to entry for customers who want to contribute (but not to rider)

Or perhaps a `--withRider` that does the opposite (making Rider opt-in) ?

abrooksv:
Agree Kyle, but first gotta get the build stable before I do that 😄

PR #1308, opened 2019-10-25
Update since in plugin xml to reflect 2019.2 min version

kiiadi:


kiiadi:
> Whoops

@abrooksv yap :P - do we feel that `192.*` is the correct version? Or are we dependent on some sub-release of 2019.2 ?

PR #1315, opened 2019-10-28
Return an error if SAM template has no resources

hunterwerlla:
## Description
- Deploying an empty SAM template, or one without resources, caused a non helpful error. Detect this scenario and report a more useful error.

## Screenshots
Screen Shot 2019-10-28 at 2 09 35 PM

## License


abrooksv:
Should we move this validation into the dialog?

PR #1316, opened 2019-10-28
Fix service constructor initialization issues

rli:
## Types of changes

## Description
* JetBrains started throwing errors in the latest EAP if a service constructor takes a parameter that can't be auto-injected. They provide a new annotation `@NonInjectable`, but that is not available unless you pull in the EAP SDK, so instead fix by removing those constructors.

## Testing
Look no error
image

## Checklist

## License

### DIFF
--- a/core/src/software/aws/toolkits/core/telemetry/TelemetryBatcher.kt
+++ b/core/src/software/aws/toolkits/core/telemetry/TelemetryBatcher.kt
@@ -18,6 +18,11 @@ interface TelemetryBatcher {
 
     fun flush(retry: Boolean)
 
+    /**
+     * Immediately shutdown the current batcher and delegate remaining events to a new batcher
+     */
+    fun shutdownAndFlush(batcher: TelemetryBatcher)
+
     fun onTelemetryEnabledChanged(newValue: Boolean)
 
     fun shutdown()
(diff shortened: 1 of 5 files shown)


core/src/software/aws/toolkits/core/telemetry/TelemetryBatcher.kt
```diff
@@ -76,6 +81,13 @@ open class DefaultTelemetryBatcher(
         flush(retry, isTelemetryEnabled.get())
     }
 
+    @Synchronized
+    override fun shutdownAndFlush(batcher: TelemetryBatcher) {
```
hunterwerlla: nit: this method name is weird, I first read it as we are shutting down and flushing the batcher that is passed in. Maybe something like `setBatcher` would be better? What we are actually doing is setting a new batcher in a way that works.

hunterwerlla:
Can you add in the PR why this actually fixes it?

justinmk3:
> Can you add in the PR why this actually fixes it?

Would be great to have that explanation in the commit message too :)

rli:
> > Can you add in the PR why this actually fixes it?
> 
> Would be great to have that explanation in the commit message too :)

added!

PR #1319, opened 2019-10-29
Raise command output limit for ProcessCredentialsProvider

RhodesBrownAmazon:
Override the default output limit on credential_process outputs to use a limit of 2048 bytes, which permits larger session tokens.

## Types of changes

## Description

Commands that generate credentials for credential_process (in aws/config file)
may include long session tokens that raise the total command output volume beyond
the AWS SDK default limit of 1024 bytes.

## Motivation and Context

If a process generates credentials with long session tokens (e.g., Amazon's internal Conduit tool), these cannot be used by the IntelliJ plugin via credentials_process in config since the SDK throws an exception.

An upstream request has been submitted to AWS SDK to raise the default, but this PR
will make IntelliJ plugin usable immediately with larger session credentials generated by an external process. See https://github.com/aws/aws-sdk-java-v2/issues/1490

## Testing

1. Developed a command that generated JSON credentials per https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#sourcing-credentials-from-external-processes
1. Verified these credentials could be used for AWS CLI commands via `credential_process` setting in `~/.aws/config` file.
1. Build plugin from source at 1.7 release tag.
1. Confirmed that config file did not work for plugin to select profile (error message in event log)
1. Built this PR and installed from disk into my IntelliJ
1. Verified the new config with raised limit worked as a profile to expose AWS resources via AWS Explorer.

### Environment

* macOS Mojave 10.14.6
* JDK: OpenJDK Corretto-8.232.09.1
* IntelliJ: IDEA 2019.2.3 (Ultimate Edition), Build #IU-192.6817.14

## Checklist

- [ ] I have added tests to cover my changes - difficult to test without valid, long credentials

## License


abrooksv:
Actually can you increase it to 4096, as per https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

```
The size of the security token that AWS STS API operations return is not fixed. We strongly recommend that you make no assumptions about the maximum size. The typical token size is less than 4096 bytes, but that can vary. 
```

RhodesBrownAmazon:
Happy to post a PR with a larger limit. As best I can tell, the limit is there to prevent a rogue command from running out of control. Setting it to Long.MAX is kinda subverting the intent of the SDK. How about something very generous like 64KB?

PR #1328, opened 2019-10-31
Fixing error where service exceptions were not visible when updating Lambdas

kiiadi:
Solves https://youtrack.jetbrains.com/issue/RIDER-34242

## Types of changes

Screen Shot 2019-10-30 at 5 14 29 PM

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.module.ModuleUtil
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.ValidationInfo
+import com.intellij.util.ExceptionUtil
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
```diff
@@ -235,15 +236,15 @@ class EditFunctionDialog(
...
-                                content = message("lambda.function.configuration_updated.notification", functionDetails.name)
-                            )
-                            is Exception -> error.notifyError(title = NOTIFICATION_TITLE)
-                        }
+                    .thenAccept {
+                        notifyInfo(
+                            title = NOTIFICATION_TITLE,
+                            content = message("lambda.function.configuration_updated.notification", functionDetails.name)
+                        )
+                        runInEdt(ModalityState.any()) { close(OK_EXIT_CODE) }
```
abrooksv: Does this now keep the window open the entire build / packaging process? Do we tell user what is going on or does it just sit there ?

kiiadi: No this only fires for config updates

PR #1329, opened 2019-10-31
Fix Local Lambda gutter icons running on '.aws-sam' contents

awschristou:
## Types of changes

## Description

This change skips the SAM Build folder(s) when looking for files that could contain Lambda Handlers.

It is a fix for the issue where the Lambda Handler gutter icons would try to build/invoke the files in the `.aws-sam` SAM Build folder.

This is more prevalent in scenarios where you open a folder that contains more than one SAM Application, located within child folders.

## Testing

Opened a folder containing several JS SAM Applications in WebStorm. Tried to locally run/debug JS Lambda Handlers from the various applications, both when the project root's `.aws-sam` folder was present and empty.

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
@@ -57,10 +57,10 @@ object Lambda {
         }?.isHandlerValid(project, handler) == true
     }
 
-    fun findSamBuildContents(project: Project): Collection<VirtualFile> =
+    private fun findSamBuildContents(project: Project): Collection<VirtualFile> =
         ModuleManager.getInstance(project).modules.flatMap { findSamBuildContents(it) }
 
-    fun findSamBuildContents(module: Module): Collection<VirtualFile> =
+    private fun findSamBuildContents(module: Module): Collection<VirtualFile> =
         ModuleRootManager.getInstance(module).contentRoots.map {
             it.findChild(SamCommon.SAM_BUILD_DIR)
         }.filterNotNull()
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
@@ -31,7 +31,6 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MIN_MEMORY
 import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MIN_TIMEOUT
 import software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon
 import software.aws.toolkits.jetbrains.ui.SliderPanel
-import java.lang.StringBuilder
 import java.util.concurrent.TimeUnit
 
 object Lambda {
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -22,20 +29,56 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MAX_TIMEOUT
...
+            .flatMap {
+                VfsUtil.collectChildrenRecursively(it)
+            }
+
+    private fun logHandlerPsiElements(handler: String, elements: Array<NavigatablePsiElement>) {
+        val sb = StringBuilder()
+        sb.appendln("Found ${elements.size} PsiElements for Handler: $handler")
+
+        elements.forEach {
+            sb.appendln(it.containingFile.virtualFile.path)
```
kiiadi: You can do this with `joinToString` eg:

```
elements.joinToString { it.containingFile.virtualFile.path }
```

Also you probably want to do this inside the `LOG.debug { }` statement so that it only runs if debug is enabled.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -22,20 +29,51 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MAX_TIMEOUT
...
+        ModuleRootManager.getInstance(module).contentRoots.map {
+            it.findChild(SamCommon.SAM_BUILD_DIR)
+        }.filterNotNull()
+            .flatMap {
+                VfsUtil.collectChildrenRecursively(it)
+            }
+
+    private fun logHandlerPsiElements(handler: String, elements: Array<NavigatablePsiElement>) = LOG.debug {
+        elements.joinToString(
+            prefix = "Found ${elements.size} PsiElements for Handler: $handler\n",
```
hunterwerlla: nit: this will insert two line returns after the prefix, is that intended?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -22,20 +29,51 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MAX_TIMEOUT
...
     fun isHandlerValid(project: Project, runtime: Runtime, handler: String): Boolean = ReadAction.compute<Boolean, Throwable> {
         runtime.runtimeGroup?.let {
             LambdaHandlerResolver.getInstance(it)
         }?.isHandlerValid(project, handler) == true
     }
+
+    fun findSamBuildContents(project: Project): Collection<VirtualFile> =
+        ModuleManager.getInstance(project).modules.flatMap { findSamBuildContents(it) }
+
+    fun findSamBuildContents(module: Module): Collection<VirtualFile> =
```
hunterwerlla: nit: make this one private?

kiiadi: did the needful

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
```diff
@@ -22,20 +29,51 @@ import software.aws.toolkits.jetbrains.services.lambda.LambdaLimits.MAX_TIMEOUT
...
+
+    private fun findSamBuildContents(module: Module): Collection<VirtualFile> =
+        ModuleRootManager.getInstance(module).contentRoots.map {
+            it.findChild(SamCommon.SAM_BUILD_DIR)
+        }.filterNotNull()
+            .flatMap {
+                VfsUtil.collectChildrenRecursively(it)
+            }
+
+    private fun logHandlerPsiElements(handler: String, elements: Array<NavigatablePsiElement>) = LOG.debug {
```
abrooksv: This has no reason to return anything, remove the return type and put it in braces

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/Lambda.kt
awschristou: The separator is only used in between items in `elements`. There is only one newline between prefix and the joined text.

abrooksv: Chris is right,
```
    buffer.append(prefix)
    var count = 0
    for (element in this) {
        if (++count > 1) buffer.append(separator)
        if (limit < 0 || count <= limit) {
            buffer.appendElement(element, transform)
        } else break
    }
    if (limit >= 0 && count > limit) buffer.append(truncated)
    buffer.append(postfix)
    return buffer
```

PR #1330, opened 2019-10-31
Filter the SAM Runtimes list for Rider

hunterwerlla:
## Description

Issue: Rider supports Node, our Intellij plugin supports node, but our Rider plugin does not support Node. So, Node shows up in the list of runtimes but generates a C# project.

To fix this, add a filter to `SamInitSelectionPanel` which is where the runtime list is actually generated.

Other approaches considered: 
- Filter the list after the fact in `initSamPanel`. This seems less ideal because it fixes the problem after the fact.
- Implement Node support for Rider. I think this should be done in the future, but all of the existing code makes assumptions that it's C# in a csproj in a sln, so it would require major refactoring.

## Types of changes

Screen Shot 2019-10-31 at 8 21 04 AM

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/wizard/SamInitSelectionPanel.java
```diff
@@ -42,7 +42,20 @@
...
     private final SamProjectGenerator generator;
 
+    /**
+     * Takes a runtime name and filters based on it.
+     */
+    private interface RuntimeFilter {
+        boolean filterExpression(String runtime);
+    }
+
+    private static final RuntimeFilter filterNoRuntimes = (s) -> true;
```
awschristou: Consider renaming to `includeAllRuntimes` to increase clarity by removing a negative from the name.

hunterwerlla: That's a better name, it used to be `filterNone` which is even less clear. Also added a changelog entry.

jetbrains-rider/src/software/aws/toolkits/jetbrains/ui/wizard/DotNetSamProjectGenerator.kt
```diff
@@ -56,7 +57,7 @@ class DotNetSamProjectGenerator(
     )
 
     private val generator = SamProjectGenerator()
-    private val samPanel = SamInitSelectionPanel(generator)
+    private val samPanel = SamInitSelectionPanel(generator) { it.toLowerCase().startsWith("dotnetcore") }
```
abrooksv: Query the `RuntmieGroup.DOTNET` so we dont have this logic that is different than elsewhere

hunterwerlla: Good idea, fixed

PR #1331, opened 2019-10-31
Allow plugin to be built without Rider by simply commenting it out of…

abrooksv:
… settings.gradle

## License


kiiadi:
Add this to the `settings.gradle`

```
if (!startParameter.properties.containsKey("skipRider")) {
    include 'jetbrains-rider'
}
```

Then we can skip without making code changes by running: `./gradlew build -PskipRider`

PR #1340, opened 2019-11-05
Fix compilation issues with dotnet handler test

abrooksv:
* Also make the test also test the RPC call instead of replicating the logic

## Testing
Tests do compile, but they do not pass. (None of the Rider tests do.)

## Checklist

## License

### DIFF
--- a/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletion.kt
+++ b/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletion.kt
@@ -18,6 +18,7 @@ import com.jetbrains.rider.model.DetermineHandlersRequest
 import com.jetbrains.rider.model.HandlerCompletionItem
 import com.jetbrains.rider.model.lambdaPsiModel
 import com.jetbrains.rider.projectView.solution
+import org.jetbrains.annotations.TestOnly
 
 class DotNetHandlerCompletion : HandlerCompletion {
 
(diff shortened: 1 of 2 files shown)


jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletion.kt
```diff
@@ -30,31 +31,38 @@ class DotNetHandlerCompletion : HandlerCompletion {
...
+                }
             }
+
+            model.determineHandlersRequest.fire(DetermineHandlersRequest(requestId))
+
+            spinUntil(handlerRequestTimeoutMs) { handlersValue != null }
+
+            handlersValue ?: throw IllegalStateException("Timeout after $handlerRequestTimeoutMs ms waiting for available handlers calculation")
+        } finally {
+            lifetime.terminate()
```
abrooksv: @sdubov This would not terminate if the `handlersValue` was null before. Is this correct?

sdubov: Your change is valid. We need to manage lifetime ourselves using this API and make sure lifetime is terminated after timeout.
Could you please update `DotNetLambdaHandlerResolver::isMethodExists` function call with the same change?

abrooksv: Thanks for confirming, will put out a PR to fix that one too

abrooksv: #1342 fixes it

PR #1344, opened 2019-11-07
ResourceSelector extensions for broader applicability

nikp:
## Types of changes

## Description
 ResourceSelector extensions for broader applicability:
    * More advanced cell rendering
    * Provided ComboBox model
    * Making sorting optional
    * Provide aws credentials override on reload

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
@@ -11,7 +11,6 @@ import com.intellij.openapi.ui.ComponentValidator
 import com.intellij.openapi.ui.ValidationInfo
 import com.intellij.ui.ColoredListCellRenderer
 import com.intellij.ui.MutableCollectionComboBoxModel
-import com.intellij.ui.SimpleColoredComponent
 import com.intellij.util.ExceptionUtil
 import org.jetbrains.annotations.TestOnly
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -74,7 +61,7 @@ class ResourceSelector<T> private constructor(
 
     @JvmOverloads
     @Synchronized
-    fun reload(forceFetch: Boolean = false) {
+    fun reload(forceFetch: Boolean = false, awsConnectionOverride: () -> Pair<AwsRegion, ToolkitCredentialsProvider> = awsConnection) {
```
kiiadi: Why can't this be hidden inside the `awsConnection` implementation?

nikp: I'm not sure I understand what you mean - this is so that reload can be invoked on an existing selector to re-fetch using new connection settings.

So the goal here is to allow an optional parameter (or default to the one the REsourceSelector was initialized with

kiiadi: My point is why have both - the 'default' parameter is a function call - so the instantiator can control what gets returned there at any time.

nikp: Discussed offline, on the same page

Issue #1346, opened 2019-11-07
Rider/Edit Configuration reports Error: Cannot find handler in project


koendhondt:
**Describe the bug**
The handler completion does not work as expected when you define a 'public object FunctionHandler' Lambda function handler in Rider's Run/Edit configuration dialog. It provides no suggestions and displays the following message at the bottom of the window:

Error: Cannot find handler 'HelloWorld:HelloWorld.Function::FunctionHandler' in project

When changing the return type to one of the basic types (string, int, double, bool) or an AWS-specific object (APIGatewayProxyResponse, SNSEvent, S3Event, etc...), the handler is recognized.

**To reproduce**
Create a project using the following code snippet (example for processing a synchronous SES receipt rule lambda):

```csharp
using Amazon.Lambda.Core;
using Amazon.Lambda.SimpleEmailEvents;
using Amazon.Lambda.SimpleEmailEvents.Actions;

[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]

namespace HelloWorld
{
    public class Function
    {
        public object FunctionHandler(SimpleEmailEvent<LambdaReceiptAction> input, ILambdaContext context)
        {
            return new {disposition = "continue"};
        }
    }
}
```
- Go to: Run -> Edit Configurations...
- Select +, Lambda, Local
- Select Runtime: dotnetcore2.1
- Press Ctrl + Space: no suggestions appear
- Type the handler name: HelloWorld:HelloWorld.Function::FunctionHandler
- The dialog shows the error message at the bottom of the window

Now replace the code with the following code snippet and follow the same steps to create a Run/Debug configuration:

```csharp
using Amazon.Lambda.Core;
using Amazon.Lambda.SimpleEmailEvents;
using Amazon.Lambda.SimpleEmailEvents.Actions;

[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]

namespace HelloWorld
{
    public class Function
    {
        public string FunctionHandler(SimpleEmailEvent<LambdaReceiptAction> input, ILambdaContext context)
        {
            return "{\"disposition\": \"continue\"}";
        }
    }
}
```

This time, when pressing Ctrl + Space, the correct suggestion appears and the handler is recognized (this does not produce valid JSON output though).

**Expected behavior**
The handler is properly recognized when the FunctionHandler return type is any object that can be serialized by the default serializer (such as object, JObject, ...)

[...]

PR #1348, opened 2019-11-07
Remove required return type check for .NET Lambda handlers

hunterwerlla:
- Runtime construction is valid for .NET Lambda handlers, so object is a valid return type. Since that is the case, we should not filter potential Lambda handlers based on return type.

## Types of changes

## Related Issue(s)

#1346

## License


jetbrains-rider/ReSharper.AWS/src/AWS.Psi/Lambda/LambdaFinder.cs
```diff
@@ -69,8 +69,7 @@ public static bool IsSuitableLambdaMethod(IMethod method)
             return method != null &&
                    method.GetAccessRights() == AccessRights.PUBLIC &&
                    IsValidInstanceOrStaticMethod(method) &&
-                   HasRequiredParameters(method) &&
-                   HasRequiredReturnType(method);
+                   HasRequiredParameters(method);
```
normj: Noticed the HasRequiredParameters has a couple problems. First it is valid to have a method that takes in zero parameters. The ILambdaContext is optional as well. This is common when a user has existing code that they just want to run via a CloudWatch Event schedule.

The second issue it looks like the logic is requiring a serializer to be defined. That is not true for case where there is no input parameter or they are passing in a stream.

abrooksv: Your comment doesnt seem to align with the docs, I read it as ILambdaContext is required in the first paragraphs.

https://docs.aws.amazon.com/lambda/latest/dg/dotnet-programming-model-handler-types.html

Can we have the doc writers fix this, please and clarify what you said is possible in the docs?

normj: Seems like the line "If you want access to the Lambda context object, it is available by defining a method parameter of type ILambdaContext" make it sound optional.

abrooksv: Would it not be better to be explicit like the Java page? https://docs.aws.amazon.com/lambda/latest/dg/java-programming-model-handler-types.html

`You can omit the Context object from the handler method signature if it isn't needed. For more information, see AWS Lambda Context Object in Java. `

normj: I submitted a ticket to the Lambda doc team to be explicit like the Java page.

abrooksv: Thanks!

Issue #1261, opened 2019-10-11
Error while getting SAM executable version.


gamename:
**Describe the bug**
When I try to create a lambda function on PyCharm, the sam build process hangs and I get the following error in the IDE:
```
Error while getting SAM executable version.

java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at org.jetbrains.concurrency.AsyncPromise.get(AsyncPromise.kt:21)
	at org.jetbrains.concurrency.AsyncPromise.blockingGet(AsyncPromise.kt:92)
	at software.aws.toolkits.jetbrains.utils.CachingAsyncEvaluator.evaluateBlocking(CachingAsyncEvaluator.kt:77)
	at software.aws.toolkits.jetbrains.utils.CachingAsyncEvaluator.evaluateBlocking$default(CachingAsyncEvaluator.kt:75)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion.getVersionString(SamCommon.kt:115)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion.getVersionString$default(SamCommon.kt:110)
	at software.aws.toolkits.jetbrains.services.lambda.LambdaBuilder$buildLambdaFromTemplate$2.run(LambdaBuilder.kt:127)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:294)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

**To reproduce**
1. Right-click on the "Lambda" section of the "AWS Explorer" window and choose "Create new AWS Lambda"
2. Fill out the dialog box with the needed lambda function information 
3. Hit the "Create Function" button at he bottom right of the dialog
4. Click on the "Build" tool window and I get the following
[...]

PR #1365, opened 2019-11-14
Bump SAM Version default timeout to 5000ms from 500ms

hunterwerlla:
* In reasonable circumstances this call can time out (like if there is a security scanner or a slow computer), so bump the timeout when it is called

## Types of changes

## Related Issue(s)

#1261

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
@@ -96,7 +96,8 @@ class SamCommon {
             return try {
                 getInvalidVersionMessage(
                     SamVersionCache.evaluateBlocking(
-                        sanitizedPath
+                        sanitizedPath,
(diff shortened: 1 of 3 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamCommon.kt
```diff
@@ -112,7 +113,7 @@ class SamCommon {
                 ?: return "UNKNOWN"
 
             return try {
-                SamVersionCache.evaluateBlocking(sanitizedPath).result.rawVersion
+                SamVersionCache.evaluateBlocking(sanitizedPath, SamVersionCache.DEFAULT_TIMEOUT_MS).result.rawVersion
```
rli: Should add a comment so we don't accidentally remove this timeout

hunterwerlla: Added a big doc comment

PR #1369, opened 2019-11-16
build: improve ALTERNATIVE_IDE message

justinmk3:
## Types of changes

Small improvement to build script.

## Description

Problem: If path has a trailing slash, the "does not exist" message is confusing.

Solution: show a hint if the path has a trailing slash.

## Testing

Tested with (1) valid path, (2) valid path with trailing slash, (3) invalid path.

## License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,10 @@ allprojects {
                 if(file(System.env.ALTERNATIVE_IDE).exists()) {
                     alternativeIdePath = System.env.ALTERNATIVE_IDE
                 } else {
-                    throw new GradleException("ALTERNATIVE_IDE path ${System.env.ALTERNATIVE_IDE} not found")
+                    throw new GradleException("ALTERNATIVE_IDE path not found"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -50,7 +50,10 @@ allprojects {
                 if(file(System.env.ALTERNATIVE_IDE).exists()) {
                     alternativeIdePath = System.env.ALTERNATIVE_IDE
                 } else {
-                    throw new GradleException("ALTERNATIVE_IDE path ${System.env.ALTERNATIVE_IDE} not found")
+                    throw new GradleException("ALTERNATIVE_IDE path not found"
+                      + (System.env.ALTERNATIVE_IDE ==~ /.*[\/\\] *$/
+                        ? " (HINT: remove trailing slash '/')"
+                        : ": ${System.env.ALTERNATIVE_IDE}"))
```
rli: Should the env value always be printed?

justinmk3: I intentionally chose not, because the "hint" is the most important part. When a slash is present, most likely the "hint" will solve the problem. Else the worst case is that the full message will be shown after removing the slash.

PR #1382, opened 2019-11-23
Adding telemetry around cloud debug confirmation dialog.

kiiadi:


jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/InstrumentDialogWrapper.kt
```diff
@@ -56,9 +59,23 @@ class ConfirmNonProductionDialogWrapper(project: Project, serviceName: String) :
...
+
+    override fun doCancelAction() {
+        recordTelemetry(TelemetryConstants.TelemetryResult.Cancelled)
+        super.doCancelAction()
+    }
+
+    private fun recordTelemetry(result: TelemetryConstants.TelemetryResult) {
+        telemetry.record(project) {
+            datum("${TelemetryConstants.CLOUDDEBUG_TELEMETRY_PREFIX}_confirm_enable") {
+                metadata(TelemetryConstants.RESULT, result.name)
```
rli: add an explicit call to `count()` since it's only implicit in the action wrapper system

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/InstrumentDialogWrapper.kt
```diff
@@ -56,9 +59,23 @@ class ConfirmNonProductionDialogWrapper(project: Project, serviceName: String) :
...
     }
+
+    override fun doCancelAction() {
+        recordTelemetry(TelemetryConstants.TelemetryResult.Cancelled)
+        super.doCancelAction()
+    }
+
+    private fun recordTelemetry(result: TelemetryConstants.TelemetryResult) {
+        telemetry.record(project) {
+            datum("${TelemetryConstants.CLOUDDEBUG_TELEMETRY_PREFIX}_confirm_enable") {
```
bryceitoc9: Our current telemetry scheme is prefix_metricname, with no underscores or other punctuation in the metric name itself.

We're also not camelcasing or anything, but that's a standardization rule we were saying we might ditch. Optional if you want to camel case it, I won't knock you if you prefer that.

PR #1388, opened 2019-11-25
Enable Rider tests in 192 SDK

sdubov:
## Types of changes

## Description
Enable Rider tests in 192 SDK

## Motivation and Context
Fix allow to run Rider integration tests for lambda and others in 192 branch using RUN_LOCAL_ENV environment variable. The fix was cherry picked from 193 to 192 Rider SDK.

## Related Issue(s)
None.

## Testing
Verified tests can run in 192 in Jetbrains external network

None.

## Checklist

## License


sdubov:
### Failed tests status:
1. **DotNetHandlerCompletionTest.kt:**
- `testDetermineHandlers_MultipleHandlers`
*Related changes:* d80d3fa42e7299543e313a6de3bb38a2d72e1cea
*Notes:* 
\- C# logic expect we pass input data as a first parameter. Object in `ILambdaContext` type does not match this condition. We check for `ILambdaContext` only for a second (support) parameter. @abrooksv, @hunterwerlla please let me know if we expect to have lambdas with single context parameter... we will need to update logic then.

2. **LambdaGutterMarkHighlightingTest.kt:**
- `testMethod_StaticMainWithSerializer_Detected`
Fixed

- `testReturn_SyncVoid_NotDetected`
*Related changes:* d80d3fa42e7299543e313a6de3bb38a2d72e1cea
*Notes:*
\- The commit contains logic that check for return type. This start fail tests that verify for return types. @abrooksv, @hunterwerlla, we need to update tests if we drop check for return type.

- `testSerializer_AssemblyLevelInherited_Detected`
- `testSerializer_MethodLevelInherited_Detected`
*Related changes:* d80d3fa42e7299543e313a6de3bb38a2d72e1cea
*Notes:*
\- Those tests start failing because we've dropped method return check for lambda detection logic. @abrooksv, @hunterwerlla, could you please check these cases and confirm that we should consider all these cases as lambda methods?

abrooksv:
Details on dropping return type are in #1346 and #1348
Summary: Basically the runtime of the return type is what matters, not the compile time, so things like 'object' could be used or interface types. So there was no way to guarantee the handler's return type could work so we dropped its validation.

abrooksv:
We also filed a ticket internally to have the docs for handler format updated, the input parameter and even the context parameter are optional.

sdubov:
@abrooksv, taking this into account, do we consider the following construction as a valid lambda then?
```
public object Foo()
{
    return null; 
}
```
If yes, I'm a bit concern we will mark half of methods with lambda gutter mark. 

More general question - how we going to separate lambdas from other methods? Can we rely on template.yaml in that case? Or is that better to wait for an updated handler doc?

Local question - Rider tests infrastructure in 192 is fixed. Tests are failed due to changes in business logic now. Should we update tests according to latest changes or wait until we get an updated expected behaviour?

abrooksv:
> @abrooksv, taking this into account, do we consider the following construction as a valid lambda then?
> 
> ```
> public object Foo()
> {
>     return null; 
> }
> ```
> 
> If yes, I'm a bit concern we will mark half of methods with lambda gutter mark.
> 
> More general question - how we going to separate lambdas from other methods? Can we rely on template.yaml in that case? Or is that better to wait for an updated handler doc?
> 
> Local question - Rider tests infrastructure in 192 is fixed. Tests are failed due to changes in business logic now. Should we update tests according to latest changes or wait until we get an updated expected behaviour?

Agree, what if we split validation and highlighting? Validation is least strict, while highlight at least requires the IContext to be present?

abrooksv:
I tried to build this locally to start fixing the tests and still getting `repo.labs.intellij.net` errors.

I also tried to change RD version to `RD-2019.2-SNAPSHOT`

PR #1391, opened 2019-11-25
Fix TravisCI builds by fixing failing test

hunterwerlla:
- Explicitly mark our credential file generated in the test as `FileTypes.UNKOWN` which fixes the test while continuing to test what it actually tests for
- Several failed attempts bring changes along for better cleanup and using functions that we had
## Types of changes

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/core/credentials/CreateOrUpdateCredentialProfilesActionTest.kt
```diff
@@ -126,18 +130,30 @@ class CreateOrUpdateCredentialProfilesActionTest {
         val configFile = File(folderRule.newFolder(), "config")
         val credFile = folderRule.newFile("credentials")
 
+        // Mark the file as unknown for the purpose of the test. This is needed because some
+        // other extensions can have weird file type association patterns (like Docker having
+        // *. (?)) which makes this test fail because it is not file type unkown
+        val file = listOf(localFileSystem.refreshAndFindFileByIoFile(credFile))
+        localFileSystem.refreshFiles(file, false, false) {
+            ApplicationManager.getApplication().runWriteAction {
+                FileTypeManagerEx.getInstanceEx().associatePattern(
```
rli: does this need to be unmarked?

hunterwerlla: Potentially. The test was mutating state already so it is possible that this will cause another test issue in the future, but at the same time I don't think it matters that much because tests around credentials should be in or around this file.

Issue #784, opened 2019-02-03
Connection timeout when starting debugger


himanshu219:
**Describe the bug**
After starting the debugger with AWS Lambda configuration the following line gets printed but it fails with the connection timeout error
/Users/hpal/.pyenv/versions/3.7.0/envs/testonpy3.7/envs/inferface/bin/sam local invoke --template "/private/var/folders/lj/gtdzxy0d6975vtrkgrh8_5gr0000gn/T/[Local] ExamPhoto-template.yaml" --event "/private/var/folders/lj/gtdzxy0d6975vtrkgrh8_5gr0000gn/T/[Local] ExamPhoto-event.json" --debug-port 51962 --debugger-path /Applications/PyCharm.app/Contents/helpers/pydev --debug-args "-u /tmp/lambci_debug_files/pydevd.py --multiprocess --port 51962 --file"
2019-02-03 12:18:05 Found credentials in environment variables.
2019-02-03 12:18:05 Invoking inferface.src.exam_photo.lambda_handler (python3.6)
2019-02-03 12:18:05 Decompressing /var/folders/lj/gtdzxy0d6975vtrkgrh8_5gr0000gn/T/1111133873687239728.zip

Connection to python debugger failed: Connection to the debugger script at localhost:51962 timed out

I have tried increasing the connection limit in  vmoptions settings but it seems it is not taking that value
-Didea.connection.timeout=20000000

**To reproduce**
Start the debugger

**Screenshots**
<img width="1677" alt="screen shot 2019-02-03 at 1 04 26 pm" src="https://user-images.githubusercontent.com/3620468/52174031-511d9300-27b4-11e9-82ed-1faeaec30066.png">

**Your Environment**
 - OS:  Mac OS High Sierra
 - JetBrains' Product: Pycharm Professional 
 - JetBrains' Product Version: 2018.3.3
 - Toolkit Version: v1.1
 - SAM CLI Version: SAM CLI, version 0.10.0
 - JVM/Python Version: java version "1.8.0_152"

**Additional context**

abrooksv:
Does it run correctly when using Run?

himanshu219:
yes I think it's because there is no connection to debugging server required in that case

abrooksv:
Hmm... It could be the debugger is failing to mount into Docker, try adding `/Applications/PyCharm.app/Contents/helpers/` as a shared folder in docker.

slichtenthal:
[...]

Issue #642, opened 2018-11-29
Configure Docker IP for debugging


jpoissant:
**Is your feature request related to a problem? Please describe.**
Using Docker Toolbox (Windows 7) my docker server IP is `192.168.99.100` and not `localhost`. When a Lambda is run in locally the debugger (PyCharm) cannot connect to the debug port and it timeout.

I receive the following error: `Connection to the debugger script at localhost:53327 timed out` (The port change on each execution)

**Describe the solution you'd like**
Add a configuration to the Run/Debug configuration to specify the Docker Server IP with a default value to `localhost`

**Describe alternatives you've considered**
Having a static configuration of SAM `debug-port`, this will require a port forwarding in VirtualBox in that case.

**Additional context**
Using:
- PyCharm 2018.3.1
- AWS Toolkit v1.0

abrooksv:
This was added a while back but forgot to close the issue.

PR #1402, opened 2019-11-28
Added a new debug host field to AWS plugin global settings

tbtmuse:
## Types of changes

## Description
Added a new text field above cloud debug called 'Local debug host', this value is saved on a global AWS plugin level and is then used by each debug extension in place of the hardcoded "localhost"  host. 

In the event that the local debug host field is left blank, the debug host defaults to "localhost". 

## Motivation and Context
The hard coded debugger host in the debug extensions support files, made it impossible to debug lamda functions locally when the debug host is not accessible via "localhost". This change exposes that setting allowing better flexibility. 

## Related Issue(s)
#642 #784 

## Testing
Tested changes by running local gradle tests and making sure those passed, and by building the plugin and running it in intellij ultimate using the debugging locally guide. Also testing both scenarios where no value is specified and where value is specified.

Testing environment:
Windows 10 - Slow ring build 19033
SAM CLI: 0.31.1
AWS CLI: 1.16.280
Pycharm & Intellij Ultimate

## Screenshots
screenshot

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/SamDebugSupport.kt
```diff
@@ -26,17 +27,29 @@ interface SamDebugSupport {
...
         state: SamRunningState,
+        debugHost: String,
         debugPorts: List<Int>
     ): XDebugProcessStarter?
 
     fun isSupported(): Boolean = true
 
+    fun getDebugHost(): String {
+
+        val localDebugHost = LocalDebugSettings.getInstance().localDebugHost
```
kiiadi: Could do this in a single statement with the Elvis operator.

tbtmuse: Noted, I can convert this to make use of the Elvis operator here, I chose to KISS, however will update the PR.

tbtmuse:
@kiiadi thanks for taking the time to review.

To my understanding the Sam executable setting is unlikely to change per run/debug configuration and so is the location of the cloud debug cli executable, and this is why they are global (I could be very wrong). 

I also considered the debugger host to be a setting that is unlikely to change per run/debug configuration, this so why I made it global.

I don't see this working with the ECS cloud debugger feature as they both seem very specify to the type of code, lambda VS code sitting in an EC2 ibsta

kiiadi:
Just circling back here - I want to understand a bit more about your setup to see why we need this? Is there a way we can 'infer' the "local" connection? e.g would replacing hard-coded `localhost` with `InetAddress.getLocalHost()` and/or `InetAddress.getLoopbackAddress()` work instead?

tbtmuse:
The only way I can see this working without allowing a user to explicitly set the IP of their debug host, is to determine if Intellij is connected to Docker via Docker Machine and the grabbing the IP of the Docker host somehow.

InetAddress.getLocalHost() and/or InetAddress.getLoopbackAddress() will not be good enough.

abrooksv:
Are you able to connect to Docker from the IDE?

tbtmuse:
Yes indeed, I am able to connect from the IDE.

abrooksv:
Querying the Docker plugin isnt too difficult, the `com.intellij.docker.DockerCloudConfiguration` contains the API URL, so we should be able to get that and then say `.host()` on the URI if the URI is not a unix socket.

tbtmuse:
Yes it does contain the same IP I need to debug.

abrooksv:
#1484 stores it in the run config and queries it from Docker plugin

abrooksv:
@tbtmuse Thank you for the pull request and bringing this to our attention. I have merged #1484 that bases this on your docker settings, you can select the host on the SAM CLI page of the Run Configuration.

PR #1419, opened 2019-12-06
Fix the integration tests for node 12

hunterwerlla:
Add an `addBreakpoint` that works properly for Node (the correct solution)

## License


justinmk3:
> `addBreakpoint(2)` has different behavior on node12. Long term we should fix it to not use explicit line number.

Definitely worth mentioning that in the commit message.

PR #1420, opened 2019-12-09
Remove next page and search from S3 browser for now

hunterwerlla:
## Description
Remove next page and search box. Next page will be replaced with a "load more" in the next PR, search box will be added after the S3 browser is merged to master in a more tested and working way.

Screen Shot 2019-12-09 at 12 27 16 PM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/UploadObjectAction.kt
```diff
@@ -75,14 +67,14 @@ class UploadObjectAction(
         nodeFile: VirtualFile?
     ) {
         val bucketName = bucket.getVirtualBucketName()
-        var key: String
-        if (nodeFile is S3VirtualDirectory) {
-            key = "${nodeFile.name}/${fileChosen.name}"
+        val key = if (nodeFile is S3VirtualDirectory) {
```
kiiadi: nice - another `var` bites the dust.

PR #1423, opened 2019-12-10
Remove deprecated method

abrooksv:
## Testing
Window still renders correctly

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
```diff
@@ -55,6 +54,7 @@ import software.aws.toolkits.resources.message
 import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.atomic.AtomicReference
+import software.amazon.awssdk.services.toolkittelemetry.model.Unit as MetricUnit
```
kiiadi: this looks like it's in the wrong spot...

abrooksv: Why did that even move.....

PR #1425, opened 2019-12-11
Redo ProjectAccountSettingsManager

abrooksv:
* Do not clear the settings on invalid state, instead keep the validation state separate.This will allow for switching partitions in 2 clicks, where before it would reset the credentials used
* Use coroutines to clean up threading
* Move notification message on error out to a listener to de-couple it

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager2.kt
```diff
@@ -0,0 +1,167 @@
...
+    var activeRegion: AwsRegion
+        get() = activeRegionInternal!! // TODO
+        set(value) {
+            activeRegionInternal = value
+        }
+
+    /**
+     * Setting the active provider will add to the recently used list, and evict the least recently used if at max size
+     */
+    var activeCredentialProvider: ToolkitCredentialsProvider
```
abrooksv: I am tempted to just allow this and region to be nullable, it already acts strange where if you have no settings active creds, we will pre-populate all region comboboxes.

I have not done a full check of the blast radius of that change though

kiiadi: If we're doing this refactor, I'd like to collapse 'region' and 'credentials' into a single object `AwsConnectionSettings` (or some-such) - we almost *always* use the two together to form a connection to AWS - I don't think it makes sense to have them separate.

kiiadi:
Like the move to coroutines - exciting.
What do you mean about 'change to partitions in 2 clicks'

abrooksv:
> Like the move to coroutines - exciting.
> What do you mean about 'change to partitions in 2 clicks'

So right now if I wanted to switch from 'default' and 'us-west-2' to 'default-cn' and 'cn-north-1', you need to alter 2 settings, but that will lead to STS call falling and invaliding the connection settings. This would null out your credential provider, and so you will fall into this, no connection configured loop.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager2.kt
abrooksv: That is a good call out

abrooksv:
Still to do: Keep migrating codebase off old API

abrooksv:
TODO:
1. Re-do `software.aws.toolkits.jetbrains.core.AwsSettingsPanel#getSelectedValue` to remove deprecated methods, along with any remaining usages.
2. Due to the async validating, we should put the explorer into `Loading` mode with a laoding overlay, otherwise it flashes oddly.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/SpinUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+import java.time.Duration
+
+fun spinUntil(duration: Duration, condition: () -> Boolean) {
+    val start = System.nanoTime()
+    runBlocking {
+        while (!condition()) {
+            if (System.nanoTime() - start > duration.toNanos())
+                throw IllegalStateException("Condition not reached within $duration")
+            delay(1)
+        }
```
justinmk3: Just for my own enlightenment: `runBlocking` blocks the current thread, which defeats the typical purpose of a coroutine. What's the advantage of `runBlocking{ delay(1) }` vs ye olde `Thread.sleep` ?

abrooksv: In this case it is the same, since it uses the callee thread for the coroutine

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaLineMarkerTest.kt
```diff
@@ -66,6 +67,11 @@ class JavaLambdaLineMarkerTest {
         )
     }
 
+    @After
+    fun tearDown() {
+        MockProjectAccountSettingsManager.getInstance(projectRule.project).reset()
```
abrooksv: This is added here because `software.aws.toolkits.jetbrains.services.lambda.java.JavaLambdaLineMarkerTest#noCredentialsLeadsToNoMarkerIfNoOtherCriteriaPasses` clears out credentials.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ConnectionSettings.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.credentials
+
+import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
+import software.aws.toolkits.core.region.AwsRegion
+
+data class ConnectionSettings(val credentials: ToolkitCredentialsProvider?, val region: AwsRegion?)
```
kiiadi: Hrmm - was hoping that we could have the properties be 'non-null' and that the connectionsettings itself would be nullable in places where that was required. The whole benefit of having a single object is you no-longer need to do the `if (region == null || credentials == null)` thing anymore.

abrooksv: Okay, but we still need to expose the region/creds for things like switching. 

Got API recommendations?

kiiadi: but do they need to be nullable?

kiiadi: I think this is a step in the right direction - shouldn't block the PR on it - but I'd like to get to a point where we can do a single null-check for the 'connection' being there.

abrooksv: If the user has no creds on their machine, we have no choice but make it nullable.

We could have getConnectionState()? and region?, credentials? methods.

abrooksv: Pushed PR update to address your concerns

PR #1432, opened 2019-12-13
Open files from S3 directly from the explorer

hunterwerlla:
## Description
- Open files as readonly from the editor directly
- Handle unknown file types with a file type selector

Kapture 2019-12-13 at 10 08 20

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -67,10 +74,42 @@ class S3TreeTable(
     private val mouseListener = object : MouseAdapter() {
         override fun mouseClicked(e: MouseEvent) {
             val row = rowAtPoint(e.point).takeIf { it >= 0 } ?: return
+            handleOpeningFile(row, e)
             handleLoadingMore(row, e)
         }
     }
 
+    private fun handleOpeningFile(row: Int, e: MouseEvent) {
+        e.clickCount.takeUnless { it != 2 } ?: return
```
kiiadi: nit: instead of the double-negative maybe `takeIf { it == 2 }` ? or even clearer might be 

```kt
if (e.clickCount != 2) return
```
(which is more consistent with `handleLoadingMore`

What happens if I accidentally triple-click?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -67,10 +74,42 @@ class S3TreeTable(
...
+            return
+        }
+        val fileWrapper = VirtualFileWrapper(File("${FileUtil.getTempDirectory()}${File.separator}${objectNode.key.replace('/', '_')}"))
+        // set the file to not be read only so that the S3Client can write to the file
+        fileWrapper.virtualFile?.isWritable = true
+        val getObjectRequest = GetObjectRequest.builder()
+            .bucket(objectNode.bucketName)
+            .key(objectNode.key)
+            .build()
+        // TODO refactor the download file action to refactor out the common parts and add progress to this
```
kiiadi: when I commit my download util we can use that. PR incoming.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
hunterwerlla: Changed. Good point, changed the logic to check for `clickCount >= 2` since if you click 3+ times we still want to do the action

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -67,12 +74,46 @@ class S3TreeTable(
...
             handleLoadingMore(row, e)
         }
     }
 
+    private fun handleOpeningFile(row: Int, e: MouseEvent) {
+        if (e.clickCount < 2) {
+            return
+        }
+        val objectNode = (tree.getPathForRow(row).lastPathComponent as? DefaultMutableTreeNode)?.userObject as? S3TreeObjectNode ?: return
+        if (objectNode.size > S3TreeObjectNode.MAX_FILE_SIZE_TO_OPEN_IN_IDE) {
```
abrooksv: This should probably use the IDE standards:

com.intellij.openapi.util.io.FileUtilRt#LARGE_FOR_CONTENT_LOADING

PR #1438, opened 2019-12-17
Enable Java 11 runtime support

abrooksv:
Screen Shot 2019-12-17 at 12 33 02 PM

## License


hunterwerlla:
Since we are bumping it to 0.38 we can also remove the 0.35 check for schemas SAM integration

abrooksv:
> Since we are bumping it to 0.38 we can also remove the 0.35 check for schemas SAM integration

Ya, will do after, did not want to conflate PR

PR #1442, opened 2019-12-18
Allow the CloudDebug tests to not clobber each other by making the CF…

abrooksv:
…N stack immutable

Attempt to stabilize the tests by making the CFN stack not be mutated.

## License


justinmk3:
What is the key change that makes this "immutable"[1]? I see that a new task definition was created, but it's used by both java and python integ tests.

[1] This should be mentioned in the commit message.

abrooksv:
> What is the key change that makes this "immutable"[1]? I see that a new task definition was created, but it's used by both java and python integ tests.
> 
> [1] This should be mentioned in the commit message.

We were mutating the stack using a CFN parameter, so each test mutated the stack to change the image, now we just make 2 container definitions.

PR #1448, opened 2019-12-18
Minor tidy-up

kiiadi:


jetbrains-core/src/software/aws/toolkits/jetbrains/core/stack/TreeView.kt
```diff
@@ -86,7 +86,7 @@ internal class TreeViewImpl(private val project: Project, stackName: String) : T
     }
 
     init {
-        val descriptor = StackNodeDescriptor(project, stackName, StatusType.UNKNOWN, message("cloudformation.stack.loading"))
+        val descriptor = StackNodeDescriptor(project, stackName, StatusType.UNKNOWN, message("loading_resource.loading"))
```
abrooksv: Where else is this used?

abrooksv: Since it wasn't deleted in the localized file

kiiadi: fixed - not sure why IDEA wasn't showing it us unsed.

PR #1449, opened 2019-12-18
Adding ability to create a new 'folder'

kiiadi:


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/NewFolderAction.kt
```diff
@@ -0,0 +1,44 @@
...
+    val bucket: S3VirtualBucket
+) : ActionButtonWrapper(message("s3.new.folder"), null, null) {
+
+    override fun isEnabled(): Boolean = treeTable.selectedRows.size <= 1
+
+    override fun doActionPerformed(e: AnActionEvent) {
+
+        val node = treeTable.selectedRows.firstOrNull()?.let { treeTable.getNodeForRow(it) } ?: treeTable.getRootNode()
+
+        Messages.showInputDialog(e.project, message("s3.new.folder.name"), message("s3.new.folder"), null)?.let { key ->
```
abrooksv: Do we want to block `/` in the entered name?

kiiadi: you mean just plain ol "/" or anything with a "/" in it?

abrooksv: Both.

Should we allow foo/bar/file?
Or /foo/ seems odd as well

kiiadi: I think we should leave the validation to S3, I wonder what /folder would do

kiiadi: /folder "works" it just creates a blank node with a sub-folder of "folder"

PR #1450, opened 2019-12-18
Fix CloudDebug stack never able to update.

abrooksv:
The description is a replacement...so change the name so that the replacement can execute

```
GroupDescription
    A description for the security group. **This is informational only.**

    Constraints: Up to 255 characters in length
    Constraints for EC2-Classic: ASCII characters
    Constraints for EC2-VPC: a-z, A-Z, 0-9, spaces, and ._-:/()#,@[]+=&;{}!$*

    Required: Yes

    Type: String

    Update requires: **Replacement**
```

## Testing
Was able to update the stack in personal account

## License


abrooksv:
Fixes stack change made in #1442

kiiadi:
:)

PR #1452, opened 2019-12-19
Refactor object actions so they don't create a new client each time

kiiadi:
Previously each action created a new client instance based on the current credentials in the project. Editor tabs are long-lived, a user may have switched project-level credentials by the time the action is taken - this means that the action would not be made against the correct credential/region.

Also moved to a co-routine model, simplified testing around the `VirtualS3Bucket` so that there's less distribution of `S3Client` knowledge (e.g. each action doesn't need to know about `S3Client`s etc.


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -57,14 +56,14 @@ class S3TreeTable(
...
 
-            virtualFiles.map {
-                s3Client.upload(project, it.inputStream, it.length, bucketVirtual.name, directoryKey + it.name).whenComplete { _, error ->
-                    when (error) {
-                        is Throwable -> error.notifyError(message("s3.upload.object.failed", it.name))
-                        else -> {
-                            invalidateLevel(node)
-                            refresh()
-                        }
+            GlobalScope.launch {
```
rli: "Application code usually should use an application-defined CoroutineScope. Using async or launch on the instance of GlobalScope is highly discouraged."

Should we create a scope for the tree tabel?

kiiadi: OK - good feedback. I'm new to co-routines and was just following tutorials online :P

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3ViewerPanel.java
```diff
@@ -84,23 +85,23 @@ public void actionPerformed(ActionEvent ae) {
...
+            ColumnInfo<Object, String> modified = new S3Column(S3ColumnType.LAST_MODIFIED);
             final ColumnInfo[] COLUMNS = new ColumnInfo[] {key, size, modified};
             model = createTreeTableModel(COLUMNS);
 
             S3TreeCellRenderer treeRenderer = new S3TreeCellRenderer();
             DefaultTableCellRenderer tableRenderer = new DefaultTableCellRenderer();
             tableRenderer.setHorizontalAlignment(SwingConstants.LEFT);
-            /**
+
+            /*
```
hunterwerlla: this comment seems to be hanging out here not attached to anything

kiiadi: right - deleted.

PR #1456, opened 2019-12-19
Upgrade Kotlin to 1.3.61

abrooksv:
This version has fixed the "friend" task system so now we can use internal keyword and access it from tests even with instrumentation enabled

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/iam/CreateIamRoleDialog.kt
```diff
@@ -108,12 +108,12 @@ class CreateIamRoleDialog(
     }
 
     @TestOnly
-    fun createIamRoleForTesting() {
+    internal fun createIamRoleForTesting() {
```
kiiadi: woooooooot!

PR #1458, opened 2019-12-19
Merge master to feature/s3-browser

hunterwerlla:
## License


build.gradle
```diff
@@ -108,6 +108,7 @@ subprojects {
     dependencies {
         compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
         compile "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
+        compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.2.1"
```
kiiadi: can we add the jdk8 dependency here (which allows us to `await` on a `CompletionStage`) and then we can remove the coroutines dependencies from `jetbrains-core` that were introduced as part of the S3 changes.

hunterwerlla: Will do

PR #1469, opened 2019-12-30
Remove the child nodes from CloudFromation Stack Node in th explorer

abrooksv:
* We only supported Lambda and adding new ones slows down the list since each one needs an additional network call per node, so remove it since the benefit of having it was small
* Remove the idea of "immutable" versions of explorer nodes
* This also makes it a leaf node so we can add double click to open stack in the future

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationExplorerNodes.kt
```diff
@@ -44,84 +33,12 @@ class CloudFormationStackNode(
...
+) {
     override fun resourceType() = "stack"
 
     override fun resourceArn() = stackId
 
     override fun displayName() = stackName
 
-    override fun isAlwaysLeaf(): Boolean = false
-
-    override fun isAlwaysShowPlus(): Boolean = true
```
hunterwerlla: Should we set isAlwaysLeaf() to `true` since it is now always a leaf?

abrooksv: The parent class has it as true

PR #1487, opened 2020-01-06
Notify error when trying to upload folder to S3

hunterwerlla:
## Types of changes

## Description
- Right now you can try to upload a folder to S3. It will error. Because of UX issues we need to figure out, we should disallow this for now.

## Screenshot
Screen Shot 2020-01-06 at 1 10 26 PM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -58,6 +58,15 @@ class S3TreeTable(
 
             GlobalScope.launch {
                 virtualFiles.forEach {
+                    if (it.isDirectory) {
```
abrooksv: Can we share this with UploadObjectAction?

hunterwerlla: yep, shared

PR #1492, opened 2020-01-08
Consistent S3 actions

kiiadi:
Introducing some abstract classes for S3 actions so that we can have consistent behaviour about what to enable when.

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/CopyPathAction.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/CopyPathAction.kt
@@ -1,30 +1,23 @@
-// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package software.aws.toolkits.jetbrains.services.s3.objectActions
 
 import com.intellij.icons.AllIcons
-import com.intellij.openapi.actionSystem.AnActionEvent
-import com.intellij.openapi.actionSystem.LangDataKeys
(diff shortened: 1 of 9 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/CopyPathAction.kt
```diff
@@ -1,30 +1,23 @@
...
-            TelemetryService.recordSimpleTelemetry(e.getRequiredData(LangDataKeys.PROJECT), TELEMETRY_NAME, TelemetryResult.Succeeded)
-        }
-    }
+class CopyPathAction(private val project: Project, treeTable: S3TreeTable) : SingleS3ObjectAction(treeTable, message("s3.copy.path"), AllIcons.Actions.Copy) {
 
-    override fun update(e: AnActionEvent) {
-        e.presentation.isEnabled = treeTable.selectedRows.size == 1 && !treeTable.getSelectedNodes().any { it is S3TreeContinuationNode }
+    override fun performAction(node: S3TreeNode) {
+        CopyPasteManager.getInstance().setContents(StringSelection(node.key))
+        TelemetryService.recordSimpleTelemetry(project, TELEMETRY_NAME, TelemetryResult.Succeeded)
```
kiiadi: @hunterwerlla / @rli thoughts on pushing these `TELEMETRY_NAME` constants into a single place where *all* telemetry constants go (for now until we get a generated version) ?

(obviously this would be a separate PR / change).

hunterwerlla: I like that idea

abrooksv: Agree

PR #1496, opened 2020-01-09
S3 UI Tests

kiiadi:
Create initial version of the S3 UI tests - this currently :
- opens an empty-project
- configures credentials/region
- creates an S3 bucket
- uploads a file
- creates a folder
- uploads a file to a folder
- renames a file
- deletes a file
- opens a file
- deletes the S3 bucket

Credentials are provided via an `assume-role` call in the UI CI configuration.

### DIFF
--- a/buildspec/linuxUiTests.yml
+++ b/buildspec/linuxUiTests.yml
@@ -25,9 +25,9 @@ phases:
       - aws sts assume-role --role-arn $ASSUME_ROLE_ARN --role-session-name ui-test > /tmp/.aws/creds.json
       - |
         >/tmp/.aws/credentials echo "[default]
-        aws_access_key_id=$(jq -r '.Credentials.AccessKeyId' /tmp/.aws/creds.json)
-        aws_secret_access_key=$(jq -r '.Credentials.SecretAccessKey' /tmp/.aws/creds.json)
-        aws_session_token=$(jq -r '.Credentials.SessionToken' /tmp/.aws/creds.json)"
+        aws_access_key_id=`jq -r '.Credentials.AccessKeyId' /tmp/.aws/creds.json`
+        aws_secret_access_key=`jq -r '.Credentials.SecretAccessKey' /tmp/.aws/creds.json`
+        aws_session_token=`jq -r '.Credentials.SessionToken' /tmp/.aws/creds.json`"
       - pip install --user aws-sam-cli
 
   build:
--- a/buildspec/linuxUiTests.yml
+++ b/buildspec/linuxUiTests.yml
@@ -23,10 +23,11 @@ phases:
       - apt-get install -y xvfb icewm procps ffmpeg libswt-gtk-3-java msbuild
       - mkdir -p /tmp/.aws
       - aws sts assume-role --role-arn $ASSUME_ROLE_ARN --role-session-name ui-test > /tmp/.aws/creds.json
-      - export KEY_ID=`jq -r '.Credentials.AccessKeyId' /tmp/.aws/creds.json`
-      - export SECRET=`jq -r '.Credentials.SecretAccessKey' /tmp/.aws/creds.json`
-      - export TOKEN=`jq -r '.Credentials.SessionToken' /tmp/.aws/creds.json`
-      - echo -e "[default]\naws_access_key_id=$KEY_ID\naws_secret_access_key=$SECRET\naws_session_token=$TOKEN" > /tmp/.aws/credentials
+      - |
+        >/tmp/.aws/credentials echo "[default]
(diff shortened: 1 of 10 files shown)


buildspec/linuxUiTests.yml
```diff
@@ -21,6 +21,12 @@ phases:
     commands:
       - apt-get update
       - apt-get install -y xvfb icewm procps ffmpeg libswt-gtk-3-java msbuild
+      - mkdir -p /tmp/.aws
+      - aws sts assume-role --role-arn $ASSUME_ROLE_ARN --role-session-name ui-test > /tmp/.aws/creds.json
+      - export KEY_ID=`jq -r '.Credentials.AccessKeyId' /tmp/.aws/creds.json`
+      - export SECRET=`jq -r '.Credentials.SecretAccessKey' /tmp/.aws/creds.json`
+      - export TOKEN=`jq -r '.Credentials.SessionToken' /tmp/.aws/creds.json`
+      - echo -e "[default]\naws_access_key_id=$KEY_ID\naws_secret_access_key=$SECRET\naws_session_token=$TOKEN" > /tmp/.aws/credentials
```
justinmk3: are KEY_ID/SECRET/TOKEN needed outside of this? If not, better not to use `export`, else (if codebuild works like travis) those env vars may show up in the environment of the main build, where because of their generic names they may confuse children, or show in logs, etc. Even if not, it's worth following good hygiene, to avoid copy-paste spread of these habits. 

Something like this:


```
- >
  KEY_ID="$(jq -r '.Credentials.AccessKeyId' /tmp/.aws/creds.json)"
  SECRET="$(jq -r '.Credentials.SecretAccessKey' /tmp/.aws/creds.json)"
  TOKEN="$(jq -r '.Credentials.SessionToken' /tmp/.aws/creds.json)"
  echo -e "[default]\naws_access_key_id=$KEY_ID\naws_secret_access_key=$SECRET\naws_session_token=$TOKEN" > /tmp/.aws/credentials
```

In that example, `export` isn't needed because the env vars are defined before the `echo` command (YAML `>` joins the lines to make a single space-separated line).

- btw: prefer `$(..)` instead of backticks for readability
- `$(..)` (or backticks) should be quoted

kiiadi: thanks!

jetbrains-core-gui/tst/software/aws/toolkits/jetbrains/EmptyProjectTestCase.kt
```diff
@@ -0,0 +1,24 @@
...
+import java.nio.file.Paths
+
+abstract class EmptyProjectTestCase : GuiTestCase() {
+
+    protected val testDataPath: Path = Paths.get(System.getProperty("testDataPath"))
+
+    @Before
+    fun createEmptyProject() {
+        welcomeFrame {
+            createNewProject()
```
abrooksv: Did the openProject(folder) not work?

kiiadi: It did - but then it required going down the road of committing `.idea` files - and they might change per version, had path specific pieces that are going to differ per machine etc - it seemed like a treadmill we didn't need to be on. This was a much simpler implementation than the previous one and it seemed like a better way to go.

abrooksv: Okay, if its stable seems fine. Wonder if we can do something ourself later that doesnt need .idea so its faster but lets worry about that when UI tests become slow to run

jetbrains-core-gui/tst/software/aws/toolkits/jetbrains/ui/s3/S3BrowserTest.kt
```diff
@@ -0,0 +1,232 @@
...
+        ExtendedJTreePathFinder(target().tree).findMatchingPath(paths.toList())
+    } catch (_: Exception) {
+        null
+    }
+
+    private fun IdeFrameFixture.treeTable(block: TreeTableFixture.() -> Unit) {
+        block(TreeTableFixture(robot(), findComponentWithTimeout(this.target(), TreeTable::class.java)))
+    }
+
+    /** Copied from [com.intellij.testGuiFramework.fixtures.TreeTableFixture] **/
```
abrooksv: Why did these need to be copied vs using the TreeTableFixture directly?

kiiadi: it only supported single click.

abrooksv: Ah, i read it as a straight copy-paste

PR #1503, opened 2020-01-14
Credential Management V2

abrooksv:
Or V3? I forget

Experimental changes!

I am taking the nuclear option on our credentials abstractions in prep for multi-partition support.

At the end of the day we need to be able to create AwsCredentials scoped to the `ToolkitCredentialsProvider` (now `ToolkitCredentialsIdentifier`) and the AWS partition.

We want to cache these across projects so the user can be only prompted once (like MFA prompts), but this means we need to get the region data into the flow where we create credentials. This has led to changing method signatures and ripping out useless indirection.

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,95 @@ package software.aws.toolkits.jetbrains.core.credentials
...
-            incModificationCount()
-            ApplicationManager.getApplication().messageBus.syncPublisher(CREDENTIALS_CHANGED).providerModified(provider)
-        }
+    init {
+        Disposer.register(ApplicationManager.getApplication(), this)
 
-        override fun providerRemoved(providerId: String) {
-            incModificationCount()
-            ApplicationManager.getApplication().messageBus.syncPublisher(CREDENTIALS_CHANGED).providerRemoved(providerId)
+        for (providerFactory in EP_NAME.extensionList) {
```
kiiadi: nit: functional style please

abrooksv:
> This looks largely similar to what we have now with the exception of the additional region stuff... what am I missing?

1. `ToolkitCredentialsProvider` is gone (or will be...)
2. Its new identity object `ToolkitCredentialsIdentifier` doesnt implement AwsCredentialProvider
3. The registry (ToolkitCredentialsProviderRegistry) is gone
4. We had 2 managers, ToolkitCredentialsProviderManager is gone
5. We had 2 factories, ToolkitCredentialsProviderFactory will be gone when done
6. Factories no longer store their state of what they created, they will go into CredentialManager. Factory is a factory, it shouldnt have state
7. Shutdown method of factory will go away, leverage Disposer system instead
8. Region is now passed to getCredentialManager, renamed to getAwsCredentialManager().

I think that is a TL;DR of what will be the final changes

abrooksv:
TODO:
1. Finish fixing the profile factory tests
2. Re-add notifications on loading profile
3. ProfileReader dies is format is invalid, we should notify on that

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
abrooksv: Fixed :D

abrooksv:
TODO:
* Handle reloading updating existing ToolkitCredentialProvider
* Notify on invalid profiles

abrooksv:
Ready for review, have fun 😄

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -42,14 +43,16 @@ abstract class ProjectAccountSettingsManager(private val project: Project) : Sim
     protected val recentlyUsedRegions = MRUList<String>(MAX_HISTORY)
 
     // Internal state is visible for AwsSettingsPanel and ChangeAccountSettingsActionGroup
-    internal var selectedCredentials: ToolkitCredentialsProvider? = null
+    internal var selectedCredentialIdentifier: ToolkitCredentialsIdentifier? = null
     internal var selectedRegion: AwsRegion? = null
 
+    private var selectedCredentialsProvider: ToolkitCredentialsProvider? = null
```
hunterwerlla: Why do we store the provider and the identifier? shouldn't we be able to store just the provider?

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -5,10 +5,9 @@ package software.aws.toolkits.core.credentials
 
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
-abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
+abstract class ToolkitCredentialsIdentifier {
```
hunterwerlla: This the ID of a provider, should it be named something less general like `ToolkitCredentialsProviderIdentifier`. That name is probably too long, maybe a comment would suffice.

abrooksv: Ya, I was turned off of that myself due to the length...

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
abrooksv: I believe you are correct, will double check

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -3,26 +3,293 @@
...
+        profileHolder.update(newProfiles.validProfiles)
+        credentialLoadCallback(CredentialIdentifierChange(profilesAdded, profilesModified, profilesRemoved))
+
+        notifyUserOfResult(newProfiles, initialLoad)
+    }
+
+    private fun notifyUserOfLoadFailure(e: Exception) {
+        val loadingFailureMessage = message("credentials.profile.failed_load")
+
+        val detail = e.message?.let {
```
hunterwerlla: nit: why is this not part of `credentials.profile.failed_load`?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,151 @@ package software.aws.toolkits.jetbrains.core.credentials
...
+import software.aws.toolkits.core.credentials.ToolkitCredentialsIdentifier
 import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
+import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.core.utils.getLogger
+import software.aws.toolkits.core.utils.tryOrNull
+import software.aws.toolkits.jetbrains.core.AwsSdkClient
+import java.util.concurrent.ConcurrentHashMap
+import javax.security.auth.login.CredentialNotFoundException
+
+data class CredentialIdentifierChange(
```
hunterwerlla: The name of this class really confused me, potentially something like `CredentialsChangeEvent` would be more clear

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -313,9 +314,9 @@ class DefaultAwsResourceCache(
         cache.clear()
     }
 
-    override fun providerRemoved(providerId: String) = clearByCredential(providerId)
+    override fun providerRemoved(identifier: ToolkitCredentialsIdentifier) = clearByCredential(identifier.id)
 
-    override fun providerModified(provider: ToolkitCredentialsProvider) = clearByCredential(provider.id)
+    override fun providerModified(identifier: ToolkitCredentialsIdentifier) = clearByCredential(identifier.id)
```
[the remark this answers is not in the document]
abrooksv: We have to since the backing hash map is populated by a Provider, so we can:
1. Make Provider have a getIdentnfier() and return the underlyign identifier
2. Keep it this way

If we do (1),I would still keep the .displayName() and .id() to keep the code shorter...

Whats your thoughts?

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -17,11 +16,13 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
      */
     abstract val displayName: String
 
+    abstract val factoryId: String
```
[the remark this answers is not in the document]
abrooksv: The factory id is how CredemtnailManager knows which Factory to go back to and get the Provider for the Identifier.

So an EC2CredentialIdentifer needs to be processed by the ECSCredentialProviderFactory, this tells the CredentialManager which factory to get back from the Extensions map.

With this clarification, what are your thoughts now?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
abrooksv: Agree, will change to your suggestion

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
abrooksv: Its so we don't render a ':' if we have nothing after it.

I think there is a way to do that through localization though

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
abrooksv: If we dont have the region yet, we cant get the provider.

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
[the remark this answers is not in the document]
hunterwerlla: Makes sense, can you add a comment about it?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
hunterwerlla: We could append it in the message but that isn't much better, seems fine as is

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
hunterwerlla: Makes sense, can you add a comment about it? on first glance it seems like duplication

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,151 @@ package software.aws.toolkits.jetbrains.core.credentials
...
+        incModificationCount()
+        ApplicationManager.getApplication().messageBus.syncPublisher(CREDENTIALS_CHANGED).providerRemoved(identifier)
+    }
+
+    /**
+     * Inner class that lazy-ily requests the true AwsCredentialsProvider from the factory when needed. This acts as a middle man that allows for existing
+     * ToolkitCredentialsProvider (like ones passed to existing SDK clients), to keep operating even if the credentials they represent have been updated such
+     * as loading from disk when new values.
+     */
+    private inner class AwsCredentialProviderProxy(private val providerId: ToolkitCredentialsIdentifier, private val region: AwsRegion) :
```
hunterwerlla: nit: this seems like less of a proxy and more of a cache

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
abrooksv: Yep, and next small rev i cleaned up it a bit.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -3,26 +3,293 @@
...
+    override fun setUp(credentialLoadCallback: CredentialsChangeListener) {
+        // Load the initial data, then start the background watcher
+        loadProfiles(credentialLoadCallback, true)
 
-    companion object {
-        val profileWatcher = ProfileWatcher().also {
-            it.start()
-            // TODO: Scope this better in the cred manager refactor
-            Disposer.register(ApplicationManager.getApplication(), it)
+        profileWatcher.start {
```
hunterwerlla: nit: looks like it will call loadProfiles again if you don't dive into `ProfileWatcher` maybe a comment that it's a callback would be good

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileReader.kt
```diff
@@ -0,0 +1,99 @@
...
+            invalidProfiles[it.name()] = e
+        }
+    }
+
+    return Profiles(validProfiles, invalidProfiles)
+}
+
+private fun validateProfile(profile: Profile, allProfiles: Map<String, Profile>) {
+    when {
+        profile.propertyExists(ProfileProperty.ROLE_ARN) -> validateAssumeRoleProfile(profile, allProfiles)
```
hunterwerlla: question: does this normalize case somewhere?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
abrooksv: It doesn't actually cache anything. It's a "pointer" that is capable of replacing its underlying aws credential provider. Thats why I call it a proxy.... Better name?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
abrooksv: How about this?

```
        profileWatcher.start(callback = {
            loadProfiles(credentialLoadCallback, false)
        })
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileReader.kt
abrooksv: Nope, the Java SDK doesn't normalize

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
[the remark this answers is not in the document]
abrooksv: Added

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileReader.kt
hunterwerlla: Oh, true

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
hunterwerlla: looks good

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
hunterwerlla: sounds good, proxy seems fine

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt
```diff
@@ -32,8 +33,8 @@ open class AwsClientManager(project: Project, sdkClient: AwsSdkClient) :
 
         val busConnection = ApplicationManager.getApplication().messageBus.connect(project)
         busConnection.subscribe(CredentialManager.CREDENTIALS_CHANGED, object : ToolkitCredentialsChangeListener {
-            override fun providerRemoved(providerId: String) {
-                invalidateSdks(providerId)
+            override fun providerRemoved(identifier: ToolkitCredentialsIdentifier) {
+                invalidateSdks(identifier.id)
```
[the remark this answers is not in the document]
abrooksv: We can't since the hash map is populated by the Provider, not Identifier. We could make the provider expose its corresponding Identifier though, but I would want to keep the .id and .displayName for code compactness.... Thoughts?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,151 @@ package software.aws.toolkits.jetbrains.core.credentials
...
-import software.aws.toolkits.core.credentials.DefaultToolkitCredentialsProviderManager
 import software.aws.toolkits.core.credentials.ToolkitCredentialsChangeListener
+import software.aws.toolkits.core.credentials.ToolkitCredentialsIdentifier
 import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
+import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.core.utils.getLogger
+import software.aws.toolkits.core.utils.tryOrNull
+import software.aws.toolkits.jetbrains.core.AwsSdkClient
+import java.util.concurrent.ConcurrentHashMap
+import javax.security.auth.login.CredentialNotFoundException
```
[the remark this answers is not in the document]
abrooksv: Correct, it should have been CredentialProviderNotFound

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -3,26 +3,293 @@
...
+            notifyInfo(
+                title = refreshTitle,
+                content = "$refreshBaseMessage $numErrorMessage",
+                notificationActions = listOf(
+                    createShowMoreInfoDialogAction(message("credentials.invalid.more_info"), errorDialogTitle, numErrorMessage, message),
+                    createNotificationExpiringAction(ActionManager.getInstance().getAction("aws.settings.upsertCredentials"))
+                )
+            )
+
+            TelemetryService.recordSimpleTelemetry(null, "aws_credentials_load", false)
```
rli: should we stick this in the telemetry repo?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileReader.kt
```diff
@@ -0,0 +1,99 @@
...
+        profile.propertyExists(ProfileProperty.AWS_SESSION_TOKEN) -> validateStaticSessionProfile(profile)
+        profile.propertyExists(ProfileProperty.AWS_ACCESS_KEY_ID) -> validateBasicProfile(profile)
+        profile.propertyExists(ProfileProperty.CREDENTIAL_PROCESS) -> {
+            // NO-OP Always valid
```
rli: what happens if access keys and credential process is set?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
abrooksv: Its conflicting. I think @hunterwerlla removed these

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,151 @@ package software.aws.toolkits.jetbrains.core.credentials
...
-            ApplicationManager.getApplication().messageBus.syncPublisher(CREDENTIALS_CHANGED).providerRemoved(providerId)
-        }
+class DefaultCredentialManager : CredentialManager() {
+    private val rootDisposable = Disposer.newDisposable()
+
+    private val extensionMap: Map<String, CredentialProviderFactory> by lazy {
+        EP_NAME.extensionList
+            .onEach {
+                if (it is Disposable) {
+                    Disposer.register(rootDisposable, it)
```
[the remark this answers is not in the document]
abrooksv: I actually moved factory def up into core, for next revision. If/when core is made disposable wont exist so we cant reference it in core

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileReader.kt
abrooksv: CRED_PROCESS takes precedence

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProviderManager.kt
```diff
@@ -3,104 +3,13 @@
...
-
-    fun addChangeListener(listener: ToolkitCredentialsChangeListener)
-
-    fun removeChangeListener(listener: ToolkitCredentialsChangeListener)
-
-    /**
-     * Shuts down the manager and all registered factories
-     */
-    fun shutDown()
+class CredentialProviderNotFound : RuntimeException {
```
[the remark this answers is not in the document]
abrooksv: Good call

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -5,10 +5,9 @@ package software.aws.toolkits.core.credentials
 
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
-abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
+abstract class ToolkitCredentialsIdentifier {
     /**
-     * The ID should be unique across all [ToolkitCredentialsProvider].
-     * It is recommended to concatenate the factory type and the display name.
+     * The ID must be unique across all [ToolkitCredentialsIdentifier].
```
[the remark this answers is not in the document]
abrooksv: Default impl now does the concatenate for you, added back comment

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
hunterwerlla: I did, I think this PR should probably merge first though so there doesn't have to be as much rework. I can clean it up in my PR after this merges.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/DeinstrumentResourceAction.kt
```diff
@@ -75,4 +77,14 @@ internal class DeinstrumentAction(
             .withParameters("--service")
             .withParameters(serviceName)
     }
+
+    override fun produceTelemetry(startTime: Instant, result: Result, version: String?) {
```
kiiadi: These telemetry changes make it pretty difficult to see what's related to credential pr

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -6,75 +6,143 @@ package software.aws.toolkits.jetbrains.core.credentials
...
+     * Inner class that lazy-ily requests the true AwsCredentialsProvider from the factory when needed. This acts as a middle man that allows for existing
+     * ToolkitCredentialsProvider (like ones passed to existing SDK clients), to keep operating even if the credentials they represent have been updated such
+     * as loading from disk when new values.
+     */
+    private inner class AwsCredentialProviderProxy(private val providerId: ToolkitCredentialsIdentifier, private val region: AwsRegion) :
+        AwsCredentialsProvider {
+        override fun resolveCredentials(): AwsCredentials = getOrCreateAwsCredentialsProvider(providerId, region).resolveCredentials()
+
+        private fun getOrCreateAwsCredentialsProvider(providerId: ToolkitCredentialsIdentifier, region: AwsRegion): AwsCredentialsProvider {
+            val partitionCache = awsCredentialProviderCache.computeIfAbsent(providerId) { ConcurrentHashMap() }
```
kiiadi: Partition cache cache... :)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/DeinstrumentResourceAction.kt
abrooksv: Master has been merged in to clean that up

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -33,32 +34,22 @@ abstract class ToolkitCredentialsProvider : AwsCredentialsProvider {
...
-    }
+class ToolkitCredentialsProvider(private val identifier: ToolkitCredentialsIdentifier, delegate: AwsCredentialsProvider) : AwsCredentialsProvider by delegate {
+    val id: String = identifier.id
+    val displayName = identifier.displayName
 
-    protected fun remove(provider: T) {
-        providers.remove(provider.id)
-        credentialsProviderManager.providerRemoved(provider.id)
-    }
+    override fun equals(other: Any?): Boolean {
```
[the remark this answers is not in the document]
abrooksv: Ya, its also IDE generated. 2nd line would resolve to 

`if(ToolkitCredentialsProvider.class != null) return false`

PR #1506, opened 2020-01-16
CI: use codebuild dotnet instead of choco

justinmk3:
Attempt to avoid choco.

- codebuild provides dotnet 2.2: https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#runtime-versions-buildspec-file
- Windows build succeeds without `netfx-4.6.1-devpack`. Did the tests silently skip?

Also, [this](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html) says:

>  The base image of the Windows Server Core 2016 contains the following runtimes. 
> ...
> dotnet | 2.2

For reference, https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script provides `dotnet-install.ps1` but that's for dotnet core, which the image already includes. We apparently need .NET framework plus whatever the "devpack" includes?

- powershell alternative to choco (get URL from [here](https://docs.microsoft.com/en-us/dotnet/framework/deployment/deployment-guide-for-developers)): 
  ```
  Invoke-WebRequest -Uri "http://go.microsoft.com/fwlink/?LinkId=863262" -OutFile installer.exe
  .\installer.exe /norestart /quiet
  ```
- other alternative: https://scoop.sh/
  - comparison to choco: https://github.com/lukesampson/scoop/wiki/Chocolatey-Comparison

## License

### DIFF
--- a/buildspec/windowsTests.yml
+++ b/buildspec/windowsTests.yml
@@ -15,8 +15,8 @@ phases:
   install:
     commands:
       - choco install -y --no-progress codecov
-      - choco install -y --no-progress dotnetcore-sdk --version 2.2.104
-      - choco install -y --no-progress netfx-4.6.1-devpack
+    runtime-versions:
+      dotnet: 2.2
 
   build:
     commands:


buildspec/windowsTests.yml
```diff
@@ -15,8 +15,8 @@ phases:
   install:
     commands:
       - choco install -y --no-progress codecov
-      - choco install -y --no-progress dotnetcore-sdk --version 2.2.104
-      - choco install -y --no-progress netfx-4.6.1-devpack
+    runtime-versions:
```
kiiadi: can we use this in the linux builds as well instead of manually installing stuff?

justinmk3: looks like we do: https://github.com/aws/aws-toolkit-jetbrains/blob/3cb1f17f301d0fed5dd2f1732f4f16b2c2b0b7d2/buildspec/linuxUiTests.yml#L17

@abrooksv mentioned that we aren't running Rider tests on CI yet, so I'm not sure if this PR is ok. Do we plan to enable the rider tests on CI?

justinmk3:
obviated by https://github.com/aws/aws-toolkit-jetbrains/pull/1626

PR #1507, opened 2020-01-16
doc: "Guidelines" stub; mention test log location

justinmk3:
Minor documentation.

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ reported the issue. Please try to include as much information as you can. Detail
 
 ## Building From Source
 
-#### Requirements
+### Requirements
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -123,9 +124,14 @@ To test your changes locally, you can run the project from IntelliJ or gradle.
   ```
 
 
+## Guidelines
+
+- AWS Explorer should not have "dependencies" (such as `sam` or `cloud-debug`). It should work without needing to install extra stuff.
```
kiiadi: There's probably a lot more we can put into this area - thanks for getting it started!

justinmk3: Yeah, we may want to move it to the `designs/` folder later, but it's good to have a place to write things down meanwhile.

PR #1522, opened 2020-01-29
Fix 2019.3 gui tests

hunterwerlla:
## Types of changes

## Description
- Do not run s3 UI tests on 2019.3: they never worked on 2019.3
- We have to do a return instead of `Assumption` because the custom test runner does not handle the exception properly

## License

### DIFF
--- a/jetbrains-core-gui/tst/software/aws/toolkits/jetbrains/ui/s3/S3BrowserTest.kt
+++ b/jetbrains-core-gui/tst/software/aws/toolkits/jetbrains/ui/s3/S3BrowserTest.kt
@@ -3,6 +3,7 @@
 
 package software.aws.toolkits.jetbrains.ui.s3
 
+import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.fileEditor.FileEditorManager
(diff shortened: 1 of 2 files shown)


jetbrains-core-gui/tst/software/aws/toolkits/jetbrains/ui/s3/S3BrowserTest.kt
```diff
@@ -148,6 +155,11 @@ class S3BrowserTest : EmptyProjectTestCase() {
 
     @After
     fun cleanUp() {
+        // TODO fix tests on 2019.3
+        val info = ApplicationInfo.getInstance()
+        if (info.majorVersion == "2019" && info.minorVersionMainPart == "2") {
```
justinmk3: the comment says 2019.3 but the code checks for "2" . is that intended?

hunterwerlla: Yep, this was unfortunately because it used to be an `Assume.true`, fixed

PR #1523, opened 2020-01-29
Add JetBrains 2020.1 EAP support

hunterwerlla:
## Types of changes

## Description
Add 2020.1 EAP support
- Still pending are integration tests working, the debugger is throwing an exception internally, 
- Rider backend is throwing corrupted singleton exceptions
- We'll push this until later in February

## License


jetbrains-core/it/software/aws/toolkits/jetbrains/utils/TestUtils.kt
```diff
@@ -28,9 +29,10 @@ fun executeRunConfiguration(
     executorId: String = DefaultRunExecutor.EXECUTOR_ID
 ): Output {
     val executor = ExecutorRegistry.getInstance().getExecutorById(executorId)
+    assertThat(executor).isNotNull
```
kiiadi: If you use `kotlin.test.assertNotNull` you won't need the `!!` on line 35 cos it'll smart-cast (it uses a [contract](https://pspdfkit.com/blog/2018/kotlin-contracts/) to hint to the compiler)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplateIndex.kt
```diff
@@ -54,8 +54,9 @@ class CloudFormationTemplateIndex : FileBasedIndexExtension<String, MutableList<
         val indexedResources = mutableMapOf<String, MutableList<IndexedResource>>()
 
         fileContent.psiFile.acceptNode(object : PsiElementVisitor() {
-            override fun visitElement(element: PsiElement?) {
+            override fun visitElement(element: PsiElement) {
                 super.visitElement(element)
+                // element is nullable in versions prior to 2020.1 FIX_WHEN_MIN_IS_201
```
kiiadi: what's the fix here?

jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/tree/AsyncTreeModelTest.java
```diff
@@ -62,7 +61,7 @@
   public ProjectRule projectRule = new ProjectRule();
 
   @Test
-  public void testAggressiveUpdating() {
+  public void testAggressiveUpdating() throws TimeoutException {
```
kiiadi: Why is this a java file...

abrooksv: That would be my doing. These are taking from IJ for our fork of async tree model and the like. We can remove when base is 2020.1

jetbrains-core/it/software/aws/toolkits/jetbrains/utils/TestUtils.kt
hunterwerlla: I wanted to use the same assertion lib that we were using for everything but that looks better, changed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/CloudFormationTemplateIndex.kt
hunterwerlla: making it not nullable and leaving in `?.`. It seems like it wasn't actually nullable before but the `@NotNull` assertion was just added, so at this point I'm not sure what we should do here

PR #1531, opened 2020-02-05
Autogenerated telemetry

hunterwerlla:
## Types of changes
- Pull in telemetry generator and convert a couple of telemetry calls
- Please comment on anything related to working on this 

## License


.gitignore
```diff
@@ -8,3 +8,4 @@ build/
 Build/
 venv/
 guitest.log
+jetbrains-core/src/software/aws/toolkits/telemetry/TelemetryDefinitions.kt
```
rli: remember to delete this

jetbrains-core/build.gradle
```diff
@@ -21,6 +33,12 @@ configurations {
     testArtifacts
 }
 
+task generateTelemetry(type: GenerateTelemetry) {
+    inputFiles = []
+    outputDirectory = file("src")
+}
+compileJava.dependsOn(generateTelemetry)
```
rli: why not `compileKotlin`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/TelemetryService.kt
```diff
@@ -8,7 +8,7 @@ import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.project.Project
 import com.intellij.util.messages.Topic
-import software.amazon.awssdk.services.toolkittelemetry.model.Unit
+import software.amazon.awssdk.services.toolkittelemetry.model.Unit.MILLISECONDS
```
rli: can we ktlint to only allow this class to be imported to be an alias or something

jetbrains-core/build.gradle
hunterwerlla: Changed

.gitignore
hunterwerlla: Deleted now that it generates into build

jetbrains-core/build.gradle
```diff
@@ -21,6 +33,16 @@ configurations {
     testArtifacts
 }
 
+task generateTelemetry(type: GenerateTelemetry) {
+    inputFiles = []
+    outputDirectory = file("${project.buildDir}/generated-src")
```
justinmk3: why not commit the generated files to the tree? 

- makes it more obvious how generated source changes, which helps with code review, git-bisect, human reasoning
- generated output shouldn't have much "churn" so this won't add much noise to VCS history.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/InstrumentResourceAction.kt
```diff
@@ -12,6 +12,10 @@ import software.aws.toolkits.jetbrains.core.explorer.actions.SingleResourceNodeA
 import software.aws.toolkits.jetbrains.services.ecs.EcsServiceNode
 import software.aws.toolkits.jetbrains.services.ecs.EcsUtils
 import software.aws.toolkits.resources.message
+import software.aws.toolkits.telemetry.ClouddebugTelemetry
```
justinmk3: each telemetry type is a separate class, which must be found, imported, and then called? 

Would it be  more ergonomic to have a single thing that's imported, which then has a bunch of members that can be called? Like:

```
import software.aws.toolkits.telemetry.Telemetry
...
Telemetry.clouddebug.instrument(...)
```

I don't understand why we need all of these classes/types.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
```diff
@@ -44,22 +43,17 @@ import software.aws.toolkits.jetbrains.services.ecs.EcsClusterNode
...
-import software.aws.toolkits.jetbrains.services.telemetry.TelemetryConstants.CLOUDDEBUG_TELEMETRY_PREFIX
-import software.aws.toolkits.jetbrains.services.telemetry.TelemetryConstants.CLOUDDEBUG_VERSION
-import software.aws.toolkits.jetbrains.services.telemetry.TelemetryConstants.RESULT
-import software.aws.toolkits.jetbrains.services.telemetry.TelemetryConstants.TelemetryResult
-import software.aws.toolkits.jetbrains.services.telemetry.TelemetryService
 import software.aws.toolkits.jetbrains.utils.notifyError
 import software.aws.toolkits.jetbrains.utils.notifyInfo
 import software.aws.toolkits.resources.message
-import java.time.Duration
+import software.aws.toolkits.telemetry.Result
```
justinmk3: I keep wondering what this gains instead of plain old true/false :)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
```diff
@@ -308,19 +307,6 @@ class EditFunctionDialog(
...
-            datum(TELEMETRY_NAME) {
-                metadata("update", update)
-                metadata(RESULT, if (succeeded) TelemetryResult.Succeeded.name else TelemetryResult.Failed.name)
-            }
-        }
-    }
-
-    companion object {
-        private const val TELEMETRY_NAME = "lambda_editfunction"
-    }
```
justinmk3: The new approach is definitely a nice improvement compared to this!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/InstrumentResourceAction.kt
hunterwerlla: We can do it that way too, but IMO there is not much difference in ergonomics. You can see what classes are importable by typing `import software.aws.toolkits.telemetry` as well, so I don't have a strong opinion on which way is better.

justinmk3: > You can see what classes are importable by typing `import software.aws.toolkits.telemetry` as well

If  `Telemetry` is where everything lives, then one always just types `Telemetry.<ctrl-space>` anywhere in any file, and gets a lives of options. 

With the current structure, one has to remember a bunch of different class names, or remember the `software.aws.toolkits.telemetry` namespace, or various other half-measures.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
kiiadi: there's actually a triplicate - Success, Failure, Canceled


But I agree - I think there should be an override that allows a "Success/Failure" boolean expression.

jetbrains-core/build.gradle
kiiadi: i don't like the idea of committing generated files - to me VCS is the *source* of files that we own / control / touch / update - it's misleading if you check out a branch and a change in the generator goes and overrides all the of the source-controlled generated files. I only made one change - which is no-where near that - why are all these files changing?!

jetbrains-core/build.gradle
```diff
@@ -1,8 +1,20 @@
...
+
 apply plugin: 'org.jetbrains.intellij'
 
+buildscript {
+    repositories {
+        mavenCentral()
+        maven { setUrl("https://jitpack.io") }
+    }
+    dependencies {
+        classpath("software.aws.toolkits:telemetry-generator:0.0.14")
```
kiiadi: should we put this stuff in the `telemetry-client` module?

jetbrains-core/build.gradle
justinmk3: > it's misleading if you check out a branch and a change in the generator goes and overrides all the of the source-controlled generated files. I only made one change

That should never happen, since we pin to a specific version of the generator. 

Certainly build artifacts should not be committed to VCS, and in general I agree. But this generated code is really just scaffolding. So pretending it's not part of the tree is actually risky IMO.

jetbrains-core/build.gradle
hunterwerlla: I was trying to go in the direction of getting rid of that module, we can move it to shared and reuse it in other products that are still using Pinpoint

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
hunterwerlla: Ask and ye shall receive: https://github.com/aws/aws-toolkit-common/pull/22 . I wanted to add this as well, will update this PR with true/false vs Result where appropriate.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/InstrumentResourceAction.kt
abrooksv: I think if you are adding telemetry, you know what you are looking for already. I'd rather have the smaller files that are easier to grok

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/StartRemoteShellAction.kt
```diff
@@ -120,25 +116,21 @@ class StartRemoteShellAction(private val project: Project, private val container
...
-                            }
-                            metadata(RESULT, result)
-                            unit(Unit.MILLISECONDS)
-                            value(Duration.between(startTime, Instant.now()).toMillis().toDouble())
-                        }
-                    }
+                private fun recordTelemetry(result: Result) {
+                    ClouddebugTelemetry.shell(
+                        project,
+                        version = (cloudDebugExecutable as? ExecutableInstance.Executable)?.version,
```
abrooksv: Can be cleaned u after you merge SAM changes

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/SamProjectWizard.kt
```diff
@@ -147,21 +149,15 @@ abstract class SamProjectTemplate {
...
-                        if (it.schema.registryName == AWS_EVENTS_REGISTRY) {
-                            metadata("awsSchema", it.schema.name)
-                        }
-                    }
-                }
-            }
+            SamTelemetry.init(
+                project,
+                name = getName(),
+                result = if (hasException) Result.CANCELLED else Result.FAILED,
```
abrooksv: this has no cancel ability. This line is wrong

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/CopyPathAction.kt
```diff
@@ -8,19 +8,14 @@ import com.intellij.openapi.ide.CopyPasteManager
...
 class CopyPathAction(private val project: Project, treeTable: S3TreeTable) : SingleS3ObjectAction(treeTable, message("s3.copy.path"), AllIcons.Actions.Copy) {
 
     override fun performAction(node: S3TreeNode) {
         CopyPasteManager.getInstance().setContents(StringSelection(node.key))
-        TelemetryService.recordSimpleTelemetry(project, TELEMETRY_NAME, TelemetryResult.Succeeded)
-    }
-
-    companion object {
-        private const val TELEMETRY_NAME = "s3_copypath"
+        S3Telemetry.copyPath(project)
```
abrooksv: Fyi, I dont remember where this was discussed, but we do have a copy arn action, but its generic and not tied to a single node. It is auto-added in the ExplorerToolWindow class

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/TelemetryService.kt
```diff
@@ -113,7 +85,9 @@ class DefaultTelemetryService(settings: AwsSettings) :
         TelemetryService.subscribe(this)
         TelemetryService.syncPublisher().notify(settings.isTelemetryEnabled)
 
-        record("ToolkitStart").also {
+        // The auto generated telemetry cannot be used here. It tries to get the instance while
+        // constructing it which causes a circular dependency issue.
+        record("session_start").also {
```
abrooksv: We may be able to move this to a postStartupActivity to fix that

abrooksv: Nvm, that looks to be project level. Application level isn't do-able until 193 I think

jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/StartRemoteShellAction.kt
hunterwerlla: yep, will add comment

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/SamProjectWizard.kt
hunterwerlla: Yep this was just wrong, fixed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/CopyPathAction.kt
hunterwerlla: Yep it seems like it's one of the things with telemetry that is still using the wrapper. When I clean up the last of those, will add it, probably as a `aws_copyArn` with the class name as a metadata or something

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/TelemetryService.kt
hunterwerlla: Will add TODO to fix it when min is 193

PR #1547, opened 2020-02-18
Add CloudWatch logs design document

abrooksv:
## License


designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,62 @@
...
+|-------------------------------|---------------|
+| AWS Toolkit for Visual Studio | Not supported |
+| AWS Toolkit for VSCode        | Not supported |
+
+## Experience
+
+### View Log Groups
+
+#### Alternative
+Instead of building a new CloudWatch service node in the AWS Explorer, we instead build a CloudWatch Log Group browser tab in the CloudWatch tool window. This new tab would also be linked to the active AWS connection settings. This would differ from the other tabs which may lead to a confusing user experience and seen as a con.
```
hunterwerlla: I'm worried about the discoverability of the tab vs an explorer node. Agreed that it makes more sense from a UX perspective, especially since most accounts have a ton of groups and streams, but maybe there should be a node that opens it or something?

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,62 @@
...
+
+![ALT_BROWSE_LOG_GROUP]*Possible alternative for browsing Log Groups*
+
+### View Log Streams
+Upon selecting a Log Group, the user would be shown a new tab in the CloudWatch tool window. This new tab will list all the streams for the requested group to allow the user to find the exact stream they are looking for.
```
hunterwerlla: Are we going to add a way to sort them?

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,62 @@
...
+| Product                       | Difference    |
+|-------------------------------|---------------|
+| AWS Toolkit for Visual Studio | Not supported |
+| AWS Toolkit for VSCode        | Not supported |
+
+## Experience
+
+### View Log Groups
+
+#### Alternative
```
hunterwerlla: nit: section title is weird

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,62 @@
...
+Double clicking the Log group in the AWS Explorer will open the Log Stream selection. 
+
+For additional service integrations, such as a Lambda function, where we know what the Log Group is that is associated with the resource, a new context menu action added that will bypass having to search for the Log Group and instead open the Log Stream selection directly.
+
+### View A Log Stream
+Each Log Stream the user selects opens in its own tab to allow for browsing multiple streams at the time. The tab is also be scoped to the region and credentials used to open it so that the user can browse across regions and accounts at the same time.
+
+![VIEW_LOG_STREAM]*View of a Log Stream*
+
+#### "Treat as..."
```
hunterwerlla: This is a super cool feature

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: Oh, ya. I forgot to call it out, but you can see the sort caret in the UI mock. Will call it out

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: Ya, I honestly do not know.... I hope the team can give their ideas on this point

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
+# CloudWatch Logs Experience
+
+* **Type**: UX Design
+* **Status**: Proposal
+
+## Introduction
+CloudWatch Logs enables a user to centralize the logs from all of their systems, applications, and AWS services that they use, in a single, highly scalable service. This document describes adding CloudWatch Logs into the JetBrains toolkit in order to facilitate easily viewing and searching them for specific error codes or patterns.
```
awschristou: Nit:
This document describes adding CloudWatch Logs into ....
to
This document describes adding CloudWatch Logs _functionality_ into ...

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Comparison to other products
+
+| Product                       | Difference    |
+|-------------------------------|---------------|
+| AWS Toolkit for Visual Studio | Not supported |
```
awschristou: The Visual Studio Toolkit allows you to pull down a Lambda Function's Logs. Unsure if logs functionality is elsewhere in the toolkit. @normj ?

image

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
+# CloudWatch Logs Experience
+
+* **Type**: UX Design
+* **Status**: Proposal
```
awschristou: Even with a proposal, avoid language like "would/will". Write as if the system already exists. The approved form of the document lives on to document the actual experience.

Example:
A new AWS Explorer root node would be added for the CloudWatch service.
to
The AWS Explorer gains a root node for the CloudWatch service.

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+Log Streams can also be sorted by name or last event time in order to help facilitate finding the Log Stream of interest.
+
+*View all Log Streams in the selected Log Group:*<br />
+![BROWSE_LOG_STREAMS]
+
+#### Navigate to a Log Stream
+Double clicking the Log group in the AWS Explorer will open the Log Stream selection. 
+
+For additional service integrations, such as a Lambda function, where we know what Log Group is associated with the resource a new context menu action adds the ability to bypass having to search for the Log Group and instead open the Log Stream selection directly.
```
awschristou: I think this is going to be the more common usage pattern when working with systems that have this integration (like Lambda functions).

Run on sentence? How about...

Some services will have a more intuitive and direct association with CloudWatch Logs. For example, a Lambda Function context menu can take users directly to the function's log streams.

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+### View A Log Stream
+Each Log Stream the user selects opens in its own tab to allow for browsing multiple streams at the time. The tab is also be scoped to the region and credentials used to open it so that the user can browse across regions and accounts at the same time.
+
+*View of a Log Stream:*<br />
```
awschristou: I like the breadcrumbs across the top

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+| Product                       | Difference    |
+|-------------------------------|---------------|
+| AWS Toolkit for Visual Studio | Not supported |
+| AWS Toolkit for VSCode        | Not supported |
+
+## Experience
+
+### View Log Groups
+A new AWS Explorer root node would be added for the CloudWatch service. Underneath of that node we would add the list of Log Groups that exist in the selected region for the selected credential profile. Double clicking a Log Group would open the CloudWatch tool window with the tab for the Log Streams that belong to that group. A new tab would be created if needed.
```
awschristou: I'm concerned about the quantity of child nodes here. I prefer this over the alternate proposal though.
What if the root node had some sort of context menu to accomplish search (either as "starts with" or "contains" text style search). 

Possible ideas:
* prompt for a text, then filter the child nodes
* open to a separate tab which then operates as a searchable list

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+For additional service integrations, such as a Lambda function, where we know what Log Group is associated with the resource a new context menu action adds the ability to bypass having to search for the Log Group and instead open the Log Stream selection directly.
+
+### View A Log Stream
+Each Log Stream the user selects opens in its own tab to allow for browsing multiple streams at the time. The tab is also be scoped to the region and credentials used to open it so that the user can browse across regions and accounts at the same time.
+
+*View of a Log Stream:*<br />
+![VIEW_LOG_STREAM]
+
+#### Treat as...
```
awschristou: Can I simply copy one or more entries to clipboard?
Can I select one or more entries and open them in a scratch document (so that I can further consume them however I want)?

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+The View As feature acts as a Log Event post-processor to convert the Log Event into a more readable format. This feature provides a pre-defined list of formats such as JSON, Plain Text, CSV, or [CLF](https://httpd.apache.org/docs/1.3/logs.html#common). The formats would process each Log Event and convert them into a more tabular format where each column maps to an extract format. The filter feature would filter on these extracted values.
+
+*Treat As -> JSON:*<br />
```
awschristou: If one field contains a json blob itself, do you take further action, or treat it as a string?

abrooksv: Treat it as a string, the entire log line would need to be JSON to be trated correctly.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: What do you mean "further consume them however I want"?

I think we need to play with the copy paste to see what feels "natural"

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: So a tree node with no children?

> open to a separate tab which then operates as a searchable list

Open a tab where?

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: Is it logs from cloudwatch or just the tail of the logs from a call to `invokeFunction` ?

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: I think it has to be a log-group list in the explorer as the entry-point.

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Experience
+
+### View Log Groups
+A new AWS Explorer root node would be added for the CloudWatch service. Underneath of that node we would add the list of Log Groups that exist in the selected region for the selected credential profile. Double clicking a Log Group would open the CloudWatch tool window with the tab for the Log Streams that belong to that group. A new tab would be created if needed.
+
+*Browsing Log Groups in AWS Explorer:*<br />
+![BROWSE_LOG_GROUP]
+
+#### Alternative Proposal
```
kiiadi: Along the same lines as 'future' vs 'present' tense - I'd rename this title to be "Alternatives Considered"

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+*View all Log Streams in the selected Log Group:*<br />
+![BROWSE_LOG_STREAMS]
+
+#### Navigate to a Log Stream
+Double clicking the Log group in the AWS Explorer will open the Log Stream selection. 
+
+For additional service integrations, such as a Lambda function, where we know what Log Group is associated with the resource a new context menu action adds the ability to bypass having to search for the Log Group and instead open the Log Stream selection directly.
+
+### View A Log Stream
+Each Log Stream the user selects opens in its own tab to allow for browsing multiple streams at the time. The tab is also be scoped to the region and credentials used to open it so that the user can browse across regions and accounts at the same time.
```
kiiadi: "the tab is also be" doesn't make sense.

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: The title is "treat as.." vs the text saying "View As"

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,76 @@
...
+![VIEW_LOG_STREAM]
+
+#### Treat as...
+The View As feature acts as a Log Event post-processor to convert the Log Event into a more readable format. This feature provides a pre-defined list of formats such as JSON, Plain Text, CSV, or [CLF](https://httpd.apache.org/docs/1.3/logs.html#common). The formats would process each Log Event and convert them into a more tabular format where each column maps to an extract format. The filter feature would filter on these extracted values.
+
+*Treat As -> JSON:*<br />
+![VIEW_LOG_STREAM_JSON]
+
+#### View events around...
+In order to facilitate easier triaging, users should be able to search for events such as "Exception" or "Error". This will filter the currant Log Stream's events to events that contain the specified string. The user can then right click a Log Entry and select `View events around...`. This context menu action provides a list of pre-canned durations (i.e. `30 seconds`, `1 minute`, `5 minutes`, `custom`) that would open up a new Log Stream tab with that duration pre-selected. This allows the user to see events that led up to and resulted from the error they were searching for without losing their triage progress.
```
kiiadi: Usually I'd want a preceding number of lines rather than a time range - is that possible?

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: It is the log stream downloaded to a temp file and opened in Notepad

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: We could keep going back in time until we reach the X amount, switching to that system though forks the UI in a strange way. Not sure best way we should show that we are going to show X instead of time

rli:
Does the search box in all of these filter the existing view or does it call out to CloudWatch to perform an actual search? I think we should have a speed search instead if its the former.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: Changed

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: Altered it

abrooksv:
> Does the search box in all of these filter the existing view or does it call out to CloudWatch to perform an actual search? I think we should have a speed search instead if its the former.

I think the driving factor would be: does it hide things that it does not match? This is how the VCS log works.

Speedsearch is good at finding things, but not removing contextually unrelated items to lessen cognitive load.

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,77 @@
...
+### AWS Toolkit for VSCode  
+Not supported
+
+### AWS Toolkit for Visual Studio  
+Supports listing of Lambda CloudWatch Log Streams and downloading them to a temporary file and opening it in Notepad
+
+## Experience
+
+### View Log Groups
+The AWS Explorer gains a CloudWatch root node for interacting wit CloudWatch service. The node lists all the Log Groups that exist in the selected region for the selected credential profile as child nodes. Double clicking a Log Group opens the CloudWatch tool window with the tab for the Log Streams that belong to that group. A new tab is in the tool window is created if one does not already exist.
```
ckant: Log group names can be very very long. What's the strategy for displaying them? (truncate, group by prefix, etc.)

designs/cloudWatch/cloudWatchIntegration.md
ckant: A supplemental experience could be to add a sub-node to the Lambda function. Logs could be considered something that the Lambda "has".

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,77 @@
...
+*View all Log Streams in the selected Log Group:*<br />
+![BROWSE_LOG_STREAMS]
+
+#### Navigate to a Log Stream
+Double clicking the Log group in the AWS Explorer opens the Log Stream selection. 
+
+For additional service integrations, such as a Lambda function, where we know what Log Group is associated with the resource a new context menu action adds the ability to bypass having to search for the Log Group and instead open the Log Stream selection directly.
+
+### View A Log Stream
+Each Log Stream the user selects opens in its own tab to allow for browsing multiple streams at the time. The Log Stream tab is scoped to the region and credentials used to open it so that the user can browse across regions and accounts at the same time.
```
ckant: So the credentials management supports having multiple simultaneous credentials in use?

It might be a little confusing considering that the main AWS explorer only allows you to view resources for one set of a credentials at a time. Would that change as well?

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,77 @@
...
+
+*View of a Log Stream:*<br />
+![VIEW_LOG_STREAM]
+
+#### Treat as...
+The Treat As feature acts as a Log Event post-processor to convert the Log Event into a more readable format. This feature provides a pre-defined list of formats such as JSON, Plain Text, CSV, or [CLF](https://httpd.apache.org/docs/1.3/logs.html#common). The formats process each Log Event and convert them into a more tabular format where each column maps to an extract format. The filter feature filters on these extracted values.
```
ckant: This seems like it would be a supplemental feature rather than a p0 feature. Would you agree?

IMO, the IDE seems like a quick and convenient way to view logs. But for more powerful features, the user might turn elsewhere. (e.g. Log services like Graylog, Splunk, etc.)

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,77 @@
...
+
+*Treat As -> JSON:*<br />
+![VIEW_LOG_STREAM_JSON]
+
+#### View events around...
+In order to facilitate easier triaging, users should be able to search for events such as "Exception" or "Error". This filters the currant Log Stream's events to events that contain the specified string. The user can then right click a Log Entry and select `View events around...`. This context menu action provides a list of pre-canned durations (i.e. `30 seconds`, `1 minute`, `5 minutes`, `custom`) that opens up a new Log Stream tab with that duration pre-selected. This allows the user to see events that led up to and resulted from the error they were searching for without losing their triage progress.
+
+#### Tail Stream
+TODO
```
ckant: Is this TODO as in "p0 and TBD" or are you saying this will be done post launch?

I see a (the most?) common dev cycle being:

1. Do something locally
2. Check logs

I would rank this as ultra important.

designs/cloudWatch/cloudWatchIntegration.md
```diff
@@ -0,0 +1,77 @@
...
+
+*View all Log Streams in the selected Log Group:*<br />
+![BROWSE_LOG_STREAMS]
+
+#### Navigate to a Log Stream
+Double clicking the Log group in the AWS Explorer opens the Log Stream selection. 
+
+For additional service integrations, such as a Lambda function, where we know what Log Group is associated with the resource a new context menu action adds the ability to bypass having to search for the Log Group and instead open the Log Stream selection directly.
+
+### View A Log Stream
```
ckant: What happens if new logs come in after the user loads the log stream. Does the view refresh automatically? Or is there a button to do this manually?

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: There isn't any harm in long names in the tree, they get truncated to the width of the tool window, and show the full value when moused over. Speedsearch also still searches the "hidden" parts

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: No, but a opened "resource" doesn't stop functioning if you change credentials. Like the Cloudfromation stack viewer keeps showing events even if you change region, or the S3 bucket viewer. This feature enables the users to compare stuff across regions or accounts without losing anything by doing so.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: I disagree, the IDE should improve the DevX as much as it can in order to ease development. We shouldn't force customers to onboard to another external system for this, especially since our own services can and do produce logs in this format.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: It is p0, just I am getting stuck on part of the UX, if the customer is in the middle of a stream, and we have not loaded the data to the end, but is the behavior when they enable tail? Do we:

1. Drop all data and start clean in tail mode as if they "connected" to a live stream?
2. Do we attempt to load all data till now (this seems expensive)
3. Is tail a special mode that acts 100% different than viewing historical data?

I can't decide since everything has pros/cons

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: In non-tail (live?) mode, we would probably operate like the console and when they reach the end have a Check for More row similar to the S3 truncation row.

designs/cloudWatch/cloudWatchIntegration.md
ckant: My fear is that information that shows up (without hover) might just be a long prefix.

e.g. the common paradigm is `/aws/<service-name>`

If the service name is pretty long (like `/aws/kinesisfirehose/`) the user might just see a bunch of `/aws/kinesisfir..` and wouldn't be able to discern which one to choose at a glance.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: > A supplemental experience could be to add a sub-node to the Lambda function. Logs could be considered something that the Lambda "has".

We have no active prior art on this, we used to have Lambdas that belong to a CloudFromation stack underneath of it but we removed that for technical reasons. Your idea interests me and would probably be a good question to ask in a future UX study

designs/cloudWatch/cloudWatchIntegration.md
ckant: Do you think that could be extended to the AWS explorer window? i.e. having multiple independent explorers running simultaneously?

This seems to be the closest analog to having multiple tabs open in the AWS console.

designs/cloudWatch/cloudWatchIntegration.md
ckant: Sure, we're in agreement there. 

The question is the cost to implement *and maintain* these features throughout the life-cycle of the plugin and the future evolution of CloudWatch as a service (as well as components like the CloudWatch Console / Insights).

That needs to be weighted against the opportunity cost of not implementing / maintaining other features.

abrooksv: My hope is this feature is cheap and can be accomplished with a pluggable interface `fun format(str: String): List<Pair<String, String>> (Column, Value)`, so that the ROI is high and cost to implement is low. We can always revisit it during implementation and cut it then, right?.

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: how do they determine which log stream it is in CW for a given invocation?

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: Definitely an interesting concept.

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: the lists are already searchable by just typing. eg

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: should we split on common separators and build a hierarchy? (e.g `/`) similar to s3? if we did that I think that takes us back to the toolwindow for finding a log-group. The names are max 512 characters (which I agree is pretty long) but I think for most use-cases it'll be OK.

designs/cloudWatch/cloudWatchIntegration.md
kiiadi: Tabbing the explorer is an interesting concept actually - it's sort of an alternative to VS Code's multi-region explorer.

designs/cloudWatch/cloudWatchIntegration.md
abrooksv: Doesn't seem to, its just a list of all and you select which to download

kiiadi: ahhh - k - yeah that's a shame.

designs/cloudWatch/cloudWatchIntegration.md
justinmk3: \4. show a "page break" sigil and start tailing data beneath it.

```
line 1
line 2
[...]  // tail enabled
line 5234233
line 5234234
```

PR #1552, opened 2020-02-19
Don't retry 400 errors when publishing to telemetry.

kiiadi:
Also moving TelemetryPublisher to use co-routines.

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
@@ -1,12 +1,13 @@
-// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package software.aws.toolkits.jetbrains.services.telemetry
 
 import com.intellij.openapi.application.ApplicationInfo
-import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.ApplicationNamesInfo
 import com.intellij.openapi.util.SystemInfo
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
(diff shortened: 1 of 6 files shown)


justinmk3:
> Also moving TelemetryPublisher to use co-routines.

Why? We haven't documented anywhere our general aim and guidelines for use of coroutines so it can't be assumed obvious.

At a glance, I see `publish` was marked with `suspend` but it looks like all calls to publish() are wrapped with `runBlocking` so it is not obvious what was gained. Commit message is a good place to call out the notable parts of a changeset.

kiiadi:
There is prior art to moving from Java-style `CompletionStage` / `runOnPooledThread` style asynchronicity to a more Kotlin-style with the use of coroutines, see here: https://github.com/aws/aws-toolkit-jetbrains/commit/2bafa5c662de7912de3b02396450d2a2d6d7651a ultimately I think we'll end up moving most instances of asynchronous stuff to suspended functions - but we haven't reviewed an official stance yet. Do you think it warrants it's own doc? Where would it belong?

FYI - my commit message does mention the move to coroutines.

justinmk3:
> FYI - my commit message does mention the move to coroutines.

Yes, but it doesn't explain why.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
```diff
@@ -35,8 +34,8 @@ class DefaultTelemetryPublisher(
     private val os: String = SystemInfo.OS_NAME,
     private val osVersion: String = SystemInfo.OS_VERSION
 ) : TelemetryPublisher {
-    override fun publish(metricEvents: Collection<MetricEvent>): Boolean = ApplicationManager.getApplication().executeOnPooledThread<Boolean> {
-        try {
+    override suspend fun publish(metricEvents: Collection<MetricEvent>) {
+        withContext(Dispatchers.IO) {
```
hunterwerlla: Style question: should we do `= withContext` since the whole function is withContext?

kiiadi: yes and when the function has an explicit return type, we can - but you can't when the function is effectively a `Unit`.

PR #1568, opened 2020-02-27
CI: skip codecov if $CODE_COV_TOKEN is not defined

justinmk3:
## License

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ script:
 - ./gradlew check -S --info --console plain
 
 after_success:
-- bash <(curl -s https://codecov.io/bash) -F unittest
+- test $TRAVIS_BRANCH = "master" && bash <(curl -s https://codecov.io/bash) -F unittest
 
 branches:
   only:
--- a/buildspec/windowsTests.yml
+++ b/buildspec/windowsTests.yml
@@ -15,7 +15,7 @@ phases:
   install:
     commands:
       - |
-        if(-Not($Env:CODEBUILD_BUILD_SUCCEEDING -eq "0" -Or $Env:CODE_COV_TOKEN -eq $null)) {
+        if(-Not($Env:CODE_COV_TOKEN -eq $null)) {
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -42,7 +42,7 @@ script:
 - ./gradlew check -S --info --console plain
 
 after_success:
-- bash <(curl -s https://codecov.io/bash) -F unittest
+- test $TRAVIS_BRANCH = "master" && bash <(curl -s https://codecov.io/bash) -F unittest
```
justinmk3: Should change this to check for `CODE_COV_TOKEN` later. But it works for our purposes meanwhile.

abrooksv:
I am not seeing the benefit of these changes, what is warranting it?

justinmk3:
> I am not seeing the benefit of these changes, what is warranting it?

Avoids unnecessary downloads to third parties, which avoids potential failed builds.

buildspec/windowsTests.yml
```diff
@@ -14,7 +14,10 @@ env:
 phases:
   install:
     commands:
-      - choco install -y --no-progress codecov
+      - |
+        if(-Not($Env:CODEBUILD_BUILD_SUCCEEDING -eq "0" -Or $Env:CODE_COV_TOKEN -eq $null)) {
```
rli: Since this is the first command in the phase, isn't `CODEBUILD_BUILD_SUCCEEDING` always true?

justinmk3: I wasn't sure where that env var actually gets defined. Will remove it from here.

justinmk3:
Anything blocking this?

.travis.yml
abrooksv: So we no longer will do CodeCov in Travis PRs?

justinmk3: TRAVIS_BRANCH is the target branch

abrooksv: Ah

PR #1570, opened 2020-02-28
Make the SchemaSearchDialogTest asserts work and fix tests, remove mockk

abrooksv:
## Types of changes

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/schemas/search/SchemaSearchDialogTest.kt
```diff
@@ -100,188 +84,228 @@ class SchemaSearchDialogTest {
...
 
-        future.thenApply {
-            assertSearchResultsDialogState(it, emptyList(), listOf(searchError))
+        val dialog = runInEdtAndGet {
+            initSingleRegistryDialog().also {
+                it.searchTextField.text = SEARCH_CRITERIA
+            }
         }
+
+        latch.await(3, TimeUnit.SECONDS)
```
hunterwerlla: nit: since this is copy pasted, why not have a wait function if we need to change the timeout ever

PR #1575, opened 2020-03-03
Add mechanism for users to submit feedback from within the toolkit

rli:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Screenshots (if appropriate)
feedback2

## Checklist

- [x] I have read the **README** document
- [x] I have read the **CONTRIBUTING** document
- [x] Local run of `gradlew check` succeeds
- [x] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] A short description of the change has been added to the **CHANGELOG**

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1581, opened 2020-03-05
Use coroutines instead of threads for feedback submission

rli:
Follow up for #1575
* non-modal popup on submission success
* fixed bug where submit button is still disabled after failure
* coroutine-ify previous usages of threads
    * Please pay special attention to the dispatchers, not sure if there's a better way to do this

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/TelemetryService.kt
```diff
@@ -133,7 +134,9 @@ class DefaultTelemetryService(settings: AwsSettings) :
     }
 
     override fun sendFeedback(sentiment: Sentiment, comment: String) {
-        publisher.sendFeedback(sentiment, comment)
+        runBlocking {
```
kiiadi: Should `sendFeedback` be a blocking call, or should it get queued up with the rest of our metrics and flushed asynchronously?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,14 +37,20 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
 
             val sentiment = panel.sentiment ?: throw IllegalStateException("sentiment was null after validation")
             val comment = panel.comment ?: throw IllegalStateException("comment was null after validation")
-            ApplicationManager.getApplication().executeOnPooledThread {
+            GlobalScope.launch(ApplicationThreadPool) {
+                val edtDispatcher = dispatcherFor(ModalityState.stateForComponent(panel.panel))
                 try {
                     TelemetryService.getInstance().sendFeedback(sentiment, comment)
```
kiiadi: What do we want the behaviour here to be? Do we really want to block the UI on the actual post to the metrics service? Or should we adopt a more asynchronous pattern - they click "Send" we attempt to post the feedback, if it doesn't work they get a toast that says feedback failed, maybe offers them a chance to re-try?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CoroutineDispatchers.kt
```diff
@@ -0,0 +1,22 @@
...
+package software.aws.toolkits.jetbrains.utils
+
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.ModalityState
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Runnable
+import kotlin.coroutines.CoroutineContext
+
+object CoroutineDispatchers {
+    fun dispatcherFor(modalityState: ModalityState): CoroutineDispatcher =
```
kiiadi: What's the intend here? Are there other places we should be using this?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
rli: I think it makes more sense to take the small time penalty of blocking until success/fail than giving the impression of success and making the user context switch back into the feedback dialog. This matches our current behavior for create S3 bucket/IAM policy, though those tie back to a larger wizard-type form

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/TelemetryService.kt
rli: ties into behavior discussion below

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CoroutineDispatchers.kt
rli: intent here was to see what this would look like before committing to converting our other usages of `runInEdt` to this

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CoroutineDispatchers.kt
```diff
@@ -0,0 +1,22 @@
...
+
+package software.aws.toolkits.jetbrains.utils
+
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.ModalityState
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Runnable
+import kotlin.coroutines.CoroutineContext
+
+object CoroutineDispatchers {
```
abrooksv: Should be able to use `AppUIExecutor.onUiThread(state).coroutineDispatchingContext()`

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
kiiadi: Seems reasonable

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CoroutineDispatchers.kt
rli: cool

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,14 +38,20 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
 
             val sentiment = panel.sentiment ?: throw IllegalStateException("sentiment was null after validation")
             val comment = panel.comment ?: throw IllegalStateException("comment was null after validation")
-            ApplicationManager.getApplication().executeOnPooledThread {
+            GlobalScope.launch(ApplicationThreadPool) {
```
ckant: I *think* you want to only group the actions that can safely be cancelled together under a single non-supervisor scope.

And group all actions tied to your application scope to an uber supervisor scope.

I believe GlobalScope is non-supervisor, which means if any of the coroutines are cancelled here, it will cancel all of them which doesn't seem correct.

So at the least, we should create our own supervisor scope and launch coroutines in it. Then we can safely create `coroutineScope { }` stuff to group the actions that can safely be cancelled together without cancelling stuff outside of their scope.

ckant: This is what I'm thinking:

* Instead of `GlobalScope`, create a new scope inside of `FeedbackDialog`
  * `val job = SupervisorJob()`
  * `val scope = job + ApplicationThreadPool + CoroutineName("FeedbackDialog")`

* Change call to `GlobalScope.launch` in `FeedbackDialog` to `scope.launch`

* Change `TelemetryService.sendFeedback` to be `suspend`, since we control it and can say it's a suspendable operation.

* Remove the `runBlocking` from `TelemetryService.sendFeedback`, since it's suspending and we can control the blocking behavior in the caller. Since, in your current implementation, the caller is not doing async stuff, it will be synchronous by default.

kiiadi: should we have a lint rule that stops us from using global scope?

rli: spun up a new class that implements `CoroutineScope` and had `FeedbackDialog` delegate out `CoroutineScope` to that

+1 for globalscope lint, not in this PR though.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
```diff
@@ -38,16 +38,18 @@ class DefaultTelemetryPublisher(
...
-        client.postFeedback {
-            it.awsProduct(clientMetadata.productName)
-            it.awsProductVersion(clientMetadata.productVersion)
-            it.os(clientMetadata.os)
-            it.osVersion(clientMetadata.osVersion)
-            it.parentProduct(clientMetadata.parentProduct)
-            it.parentProductVersion(clientMetadata.parentProductVersion)
-            it.sentiment(sentiment)
-            it.comment(comment)
+    override suspend fun sendFeedback(sentiment: Sentiment, comment: String) {
```
hunterwerlla: Style question, if the function is all `withContext` wouldn't it make more sense to do 
`fun.... (...) = withContext(context) {
}

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,19 +39,31 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
...
+            launch(coroutineContext) {
+                val edtDispatcher = AppUIExecutor.onUiThread(ModalityState.stateForComponent(panel.panel)).coroutineDispatchingContext()
                 try {
                     TelemetryService.getInstance().sendFeedback(sentiment, comment)
-                    ApplicationManager.getApplication().invokeLater({
+                    withContext(edtDispatcher) {
                         close(OK_EXIT_CODE)
-                    }, ModalityState.stateForComponent(panel.panel))
+                    }
+                    notifyInfo(message("aws.notification.title"), message("feedback.submit_success"), project)
```
hunterwerlla: notifyInfo doesn't have to be on edt?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,19 +39,31 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
 
             val sentiment = panel.sentiment ?: throw IllegalStateException("sentiment was null after validation")
             val comment = panel.comment ?: throw IllegalStateException("comment was null after validation")
-            ApplicationManager.getApplication().executeOnPooledThread {
+            launch(coroutineContext) {
+                val edtDispatcher = AppUIExecutor.onUiThread(ModalityState.stateForComponent(panel.panel)).coroutineDispatchingContext()
```
hunterwerlla: I made a function for this in the branch I'm working out of now, it also includes a disposable if the component is disposed:
```
fun getCoroutineUiContext(modalityState: ModalityState = ModalityState.any(), disposable: Disposable? = null): CoroutineContext {
    val uiThread = AppUIExecutor.onUiThread(modalityState)
    return if (disposable == null) {
        uiThread
    } else {
        uiThread.expireWith(disposable)
    }.coroutineDispatchingContext()
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
rli: no, it gets run on edt for you

jetbrains-core/src/software/aws/toolkits/jetbrains/services/telemetry/DefaultTelemetryPublisher.kt
rli: sure

rli: in this case we're not using the response so i'm leaving as is

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,19 +38,31 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
 
             val sentiment = panel.sentiment ?: throw IllegalStateException("sentiment was null after validation")
             val comment = panel.comment ?: throw IllegalStateException("comment was null after validation")
-            ApplicationManager.getApplication().executeOnPooledThread {
+            launch(coroutineContext) {
```
ckant: I think the `coroutineContext` is redundant. That argument is for *additional* context.

> The coroutine context is inherited from a CoroutineScope. Additional context elements can be specified with context argument.

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -32,19 +38,31 @@ class FeedbackDialog(project: Project) : DialogWrapper(project) {
...
+                        Messages.showMessageDialog(panel.panel, message("feedback.submit_failed", e), message("feedback.submit_failed_title"), null)
+                        setOKButtonText(message("feedback.submit_button"))
+                        isOKActionEnabled = true
+                    }
                 }
             }
         }
     }
 
+    override fun doCancelAction() {
```
ckant: Thoughts on create on `init`, destroy on `dispose?`

PR #1582, opened 2020-03-05
Add CloudWatch log group window

hunterwerlla:
## Description
- Add CloudWatch logs group window
  - Opens log streams on double click
  - Sorts by most recently written to
  - The ability to filter and refresh
- Window is in need of design cleanup later, this is a functionality focused PR
- Log streams is the next PR so this just uses our existing window
- See the giant gif below
Kapture 2020-03-05 at 13 21 07
## Questions?
- How do I get the button to just be an icon?

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -0,0 +1,131 @@
...
+        refreshButton.addActionListener { refreshLogStreams() }
+    }
+
+    private fun refreshLogStreams() {
+        runInEdt {
+            groupsPanel.startLoading()
+        }
+        GlobalScope.launch {
+            populateModel()
+            runInEdt {
```
abrooksv: This should be all coroutined by switching contexts to the AppUiExector

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private fun buildMouseListener(tableView: TableView<LogStream>) = object : MouseAdapter() {
+        override fun mouseClicked(e: MouseEvent) {
+            if (e.clickCount < 2 || e.button != MouseEvent.BUTTON1) {
+                return
+            }
+            val row = tableView.rowAtPoint(e.point).takeIf { it >= 0 } ?: return
+            val window = CloudWatchLogWindow.getInstance(project)
+            GlobalScope.launch {
```
abrooksv: Shouldnt we be avoiding the GlobalScope? Should we have a scope to this tab?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -0,0 +1,43 @@
...
+
+class LogGroupTableSorter(model: ListTableModel<LogStream>) : TableRowSorter<ListTableModel<LogStream>>(model) {
+    init {
+        sortKeys = listOf(SortKey(1, SortOrder.DESCENDING))
+        setSortable(0, false)
+        setSortable(1, false)
+    }
+}
+
+private val dateFormatter: DateTimeFormatter = DateTimeFormatterBuilder()
```
rli: Since this is user facing, use the OS's format? [DateTimeFormatter#ofLocalizedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ofLocalizedDateTime-java.time.format.FormatStyle-) Might make sort a bit harder though

rli: Can we use JetBrains' `DateFormatUtil#getDateFormat/getTimeFormat` and then sort independently of the display?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/actions/OpenLogGroups.kt
```diff
@@ -5,13 +5,21 @@ package software.aws.toolkits.jetbrains.services.cloudwatch.logs.actions
...
 
-class OpenLogGroups : SingleResourceNodeAction<CloudWatchLogsNode>(message("cloudwatch.log.open")), DumbAware {
+class OpenLogGroup : SingleResourceNodeAction<CloudWatchLogsNode>(message("cloudwatch.logs.open")), DumbAware {
     override fun actionPerformed(selected: CloudWatchLogsNode, e: AnActionEvent) {
         // TODO subsequent PRs will fill this in
         println("this will open the window")
     }
 }
+
+fun openLogGroup(project: Project, logGroupName: String) = ApplicationThreadPoolScope("openLogGroup").launch {
```
ckant: Should we follow the pattern of delegating this class to the scope?

`OpenLogGroup : CoroutineScope by ApplicationThreadPoolScope("openLogGroup")`

hunterwerlla: This isn't a class, it's a top level function. We could make it an object, but all the object is doing would be holding this function which is fine as well

hunterwerlla: It looks a bit silly but making the companion object extend coroutine scope seems like it makes the most sense in this case since we need to call it from another place and there's not really a point to make a whole new object

ckant: Oh I see now, I was confused by `actionPerformed` and thought that would call this function.

So are you sharing the window-launching logic between this Action and elsewhere? Does this actually belong with the Action? Or should it be factored out?

It looks like you could move this logic into `CloudWatchLogWindow`. Then callers of this function don't have to depend on the Action to open the window.

I'm curious if Action is a concept that exists outside the scope of performing a "standalone Action". Does the framework consider the opening of a panel from another panel to be an Action?

hunterwerlla: An action is just a convenient representation of something that can be done by the IDE with a name and maybe an icon and description. CloudWatchLogWindow is managed by the IDE's service manager so I'm not sure if that would even work properly

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/CoroutineUtils.kt
```diff
@@ -7,14 +7,27 @@ import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.AppUIExecutor
 import com.intellij.openapi.application.ModalityState
 import com.intellij.openapi.application.impl.coroutineDispatchingContext
+import com.intellij.openapi.util.Disposer
+import kotlinx.coroutines.coroutineScope
+
+// FIX_WHEN_MIN_IS_2019_3 this can be removed and the actual runUnlessDisposed can be used which has more features
```
rli: since this is a clone of the actual should this go in CompatabilityUtils instead?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -0,0 +1,155 @@
...
+import javax.swing.event.DocumentEvent
+
+class CloudWatchLogGroup(
+    private val project: Project,
+    private val logGroup: String
+) : CoroutineScope by ApplicationThreadPoolScope("CloudWatchLogsGroup"), Disposable {
+    val title = message("cloudwatch.logs.log_group_title", logGroup.split("/").last())
+    lateinit var content: JPanel
+
+    private lateinit var refreshButton: JButton
```
rli: I think we eventually need some kind of declaration/lint on when we want to prefer the JB swing events over the regular ones

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -0,0 +1,155 @@
...
+                sorter.rowFilter = null
+            } else {
+                sorter.rowFilter = RowFilter.regexFilter(text)
+            }
+        }
+    }
+
+    private fun buildMouseListener(table: JBTable) = object : MouseAdapter() {
+        override fun mouseClicked(e: MouseEvent) {
+            if (e.clickCount < 2 || e.button != MouseEvent.BUTTON1) {
```
rli: :( is there no adapter that does this logic for us?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -0,0 +1,155 @@
...
+                return
+            }
+            val row = table.rowAtPoint(e.point).takeIf { it >= 0 } ?: return
```
rli: why not `JTable#getSelectedRows`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+class CloudWatchLogsStreamsColumn : ColumnInfo<LogStream, String>(message("cloudwatch.logs.log_streams")) {
+    override fun valueOf(item: LogStream?): String? = item?.logStreamName()
+
+    override fun isCellEditable(item: LogStream?): Boolean = false
+}
+
+class CloudWatchLogsStreamsColumnDate : ColumnInfo<LogStream, String>(message("cloudwatch.logs.last_event_time")) {
+    override fun valueOf(item: LogStream?): String? {
+        val timestamp = item?.lastEventTimestamp() ?: return null
+        val date = DateFormatUtil.getDateFormat().format(timestamp)
```
rli: `DateFormatUtil.getDateTimeFormat` instead in case their locale has date/time in a separate order

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/actions/OpenLogGroups.kt
hunterwerlla: Components registered through the registration system have no issues, moved the coroutines stuff to there

jetbrains-core/src/software/aws/toolkits/jetbrains/core/toolwindow/ToolkitToolWindow.kt
```diff
@@ -33,14 +33,15 @@ class ToolkitToolWindowManager(private val project: Project) {
...
-        override fun addTab(title: String, component: JComponent, activate: Boolean, id: String): ToolkitToolWindowTab {
+        override fun addTab(title: String, component: JComponent, activate: Boolean, id: String, disposable: Disposable?): ToolkitToolWindowTab {
             val content = ContentImpl(component, title, true)
             val toolWindow = windowManager.getToolWindow(type.id)
                 ?: windowManager.registerToolWindow(type.id, true, type.anchor, project, true).also {
                     it.setIcon(type.icon)
                     it.stripeTitle = type.title
                 }
             Disposer.register(content, Disposable { closeWindowIfEmpty(toolWindow, type.id) })
+            disposable?.let { Disposer.register(content, it) }
```
abrooksv: This seems inverted, shouldnt we register when we close the tab (content) that we dispose? Not dispose the tab?

hunterwerlla: content is the parent, that's how it already is?

abrooksv: this is right, im a dummy

PR #1584, opened 2020-03-10
Do not attempt to upload 0 files to S3 when 0 are selected

hunterwerlla:
## Description
This had no user facing issues, but messed with our telemetry

## Types of changes

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
@@ -61,11 +61,6 @@ class S3TreeTable(
                 lfs.findFileByIoFile(it)
             }
 
-            if (virtualFiles.isEmpty()) {
-                LOG.info { "0 files were dragged and dropped onto the S3 view, skipping attempted upload" }
-                return
-            }
-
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -60,6 +61,11 @@ class S3TreeTable(
                 lfs.findFileByIoFile(it)
             }
 
+            if (virtualFiles.isEmpty()) {
+                LOG.info { "0 files were dragged and dropped onto the S3 view, skipping attempted upload" }
+                return
+            }
```
justinmk3: since uploadAndRefresh skips itself, this and the other case aren't needed. 

Extra context (fail early) can be helpful in ambiguous cases, but here it probably isn't needed. So less code wins.

Also with the current PR, the changes to `uploadAndRefresh` are actually dead code. So pick one :)

hunterwerlla: I agree, the reason I did both cases is if we add another avenue to `uploadAndRefresh` in the future. I guess we really don't have to know where the 0's are coming from since they aren't really an error so I'll remove the other cases outside of `uploadAndRefresh`

PR #1587, opened 2020-03-11
S3 browser cleanup and small bug fixes

hunterwerlla:
## Types of changes

## Description
- S3 browser code cleanup - remove the janky mouse adapter and replace it with the DoubleClickListener.
- S3 browser open file set max file size based on the IDE property instead of a 5MB flat, see the property that sets that [here](https://www.jetbrains.com/help/idea/tuning-the-ide.html#)
  - Also fix the message that was so long it went outside of the pop up
- Remove a dangling `ApplicationManager.getApplication().executeOnPooledThread` with a `GlobalScope.launch` for now
- S3 browser `clearSelectionOnWhitespace` did not actually work because the table covered the whole panel, fixed and converted to ClickListener

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -119,16 +126,19 @@ class S3TreeTable(
...
 
-    private fun handleLoadingMore(row: Int) {
-        val continuationNode = (tree.getPathForRow(row).lastPathComponent as? DefaultMutableTreeNode)?.userObject as? S3TreeContinuationNode ?: return
-        val parent = continuationNode.parent ?: return
+    private fun handleLoadingMore(row: Int): Boolean {
+        val continuationNode = (tree.getPathForRow(row).lastPathComponent as? DefaultMutableTreeNode)?.userObject as? S3TreeContinuationNode ?: return false
+        val parent = continuationNode.parent ?: return false
 
-        ApplicationManager.getApplication().executeOnPooledThread {
+        GlobalScope.launch {
```
rli: ?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3ViewerPanel.java
```diff
@@ -130,13 +132,18 @@ private void addTreeActions() {
     }
 
     private void clearSelectionOnWhiteSpace() {
-        mainPanel.addMouseListener(new MouseAdapter() {
+        new ClickListener() {
             @Override
-            public void mouseClicked(MouseEvent e) {
-                if (!treeTable.contains(e.getPoint())) {
+            public boolean onClick(@NotNull MouseEvent e, int clickCount) {
+                if(clickCount != 1) {
```
rli: spacing

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -179,7 +179,7 @@ s3.download.object.description=Download
 s3.download.object.progress=Downloading ''{0}''
 s3.load_more=load more...
 s3.name=Name
-s3.open.file_too_big=Cannot open files larger than {0} in the editor!
+s3.open.file_too_big=Cannot open files larger than {0}!
```
rli: Maybe more like: The editor cannot open files larger than x

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
hunterwerlla: The rest of the class uses global scope, this will need to be refactored eventually

resources/resources/software/aws/toolkits/resources/localized_messages.properties
hunterwerlla: That would be as long as the previous one which would still get cut off

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3ViewerPanel.java
hunterwerlla: oops, yay java files. Fixed

resources/resources/software/aws/toolkits/resources/localized_messages.properties
rli: If you set a title for `notifyError` does that solve the cut off issue?

abrooksv: Drop the '!'

Future idea..offer to download locally if https://plugins.jetbrains.com/plugin/11363-large-file-editor is installed

Never used it so no clue how it works

hunterwerlla: Sure, and we can set content on notifyError instead which is fixes the overhang

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -65,14 +66,19 @@ class S3TreeTable(
...
+    private val openFileListener = object : DoubleClickListener() {
+        override fun onDoubleClick(e: MouseEvent?): Boolean {
+            e ?: return false
+            val row = rowAtPoint(e.point).takeIf { it >= 0 } ?: return false
+            return handleOpeningFile(row)
+        }
+    }
+
+    private val loadMoreListener = object : DoubleClickListener() {
+        override fun onDoubleClick(e: MouseEvent?): Boolean {
```
kiiadi: You can safely remove the null check here - `e` can never be null : https://github.com/JetBrains/intellij-community/blob/master/platform/platform-api/src/com/intellij/ui/DoubleClickListener.java#L30

hunterwerlla: Yep but ondoubleclick is not marked @NotNull for some reason:
`protected abstract boolean onDoubleClick(MouseEvent event);`

PR #1604, opened 2020-03-20
Fix S3 bucket viewer speed search and make UI more idiomatic

### LINKED ISSUE #1603: S3 Object table renderer is disabled SpeedSearch rendering
Speed search highlighting is not working in the S3 Object tree due to our renderer.

We need to migrate to allow `com.intellij.ui.speedSearch.SpeedSearchUtil#applySpeedSearchHighlighting` to work or use a different base renderer

abrooksv:
* Use correct type of speed search
* Fix speed search not highlighting matches
* Use more idiomatic icons

Fixes #1603 

Before:
Screen Shot 2020-03-19 at 7 57 45 PM

After:
Screen Shot 2020-03-19 at 7 54 25 PM

Unknown file type:
Screen Shot 2020-03-19 at 8 06 37 PM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeCellRenderer.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.intellij.ui.LoadingNode
+import com.intellij.ui.SimpleTextAttributes
+import com.intellij.ui.speedSearch.SpeedSearchUtil
+import javax.swing.JComponent
+import javax.swing.JTree
+import javax.swing.tree.DefaultMutableTreeNode
+
+class S3TreeCellRenderer(private val speedSearchTarget: JComponent) : ColoredTreeCellRenderer() {
+    init {
+        myUsedCustomSpeedSearchHighlighting = true
```
kiiadi: this is hideous - but whatever works.

abrooksv: Yaaaa.... It uses the wrong jcomponent normally so it never finds anything to highlight. As to the name, I just work here :P

PR #1608, opened 2020-03-20
Clean up CloudWatch group UI

hunterwerlla:
- Breadcrumbs for where we are, later this will have actions to open group from string
- Remove search bar and replace with `SpeedSearch`
- Put refresh in the newly added idiomatic `Toolbar`. This will have an export action added to it later
- Fix sorting of the log streams + re-enable user sort + fix the width of last log stream time

Kapture 2020-03-20 at 12 51 49

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
@@ -7,19 +7,17 @@ import com.intellij.icons.AllIcons
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.actionSystem.ActionPlaces
-import com.intellij.openapi.actionSystem.AnAction
-import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.DefaultActionGroup
 import com.intellij.openapi.application.impl.runUnlessDisposed
 import com.intellij.openapi.project.Project
-import com.intellij.openapi.ui.SimpleToolWindowPanel
 import com.intellij.ui.DocumentAdapter
 import com.intellij.ui.DoubleClickListener
 import com.intellij.ui.PopupHandler
 import com.intellij.ui.ScrollPaneFactory
-import com.intellij.ui.components.JBLabel
 import com.intellij.ui.components.JBTextField
+import com.intellij.ui.components.breadcrumbs.Breadcrumbs
 import com.intellij.ui.table.JBTable
+import com.intellij.util.ui.JBUI
 import com.intellij.util.ui.ListTableModel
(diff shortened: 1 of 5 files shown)


abrooksv:
Moving the toolbar to be same height as the table was the right call. Looks good

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/CloudWatchLogGroup.kt
```diff
@@ -63,46 +59,34 @@ class CloudWatchLogGroup(
...
+        tablePanel.setContent(scroll)
     }
 
     init {
-        locationInformation.text = "${project.activeCredentialProvider().displayName} => ${project.activeRegion().displayName} => $logGroup"
-        filterField.emptyText.text = message("cloudwatch.logs.filter_log_streams")
-        filterField.document.addDocumentListener(buildStreamSearchListener(groupTable))
+        val locationCrumbs = LocationCrumbs(project, logGroup)
+        locationInformation.crumbs = locationCrumbs.crumbs
+        locationInformation.border = JBUI.Borders.customLine(JBColor.border(), 0, 0, 1, 0)
```
abrooksv: Does this look the same?
```
IdeBorderFactory.createBorder(SideBorder.BOTTOM)
```

hunterwerlla: Yep, changed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -23,19 +24,18 @@ class LogStreamsStreamColumn : ColumnInfo<LogStream, String>(message("cloudwatch
...
-        return DateFormatUtil.getDateTimeFormat().format(timestamp)
-    }
+    private val renderer = ResizingDateColumnRenderer()
+    override fun valueOf(item: LogStream?): String? = item?.lastEventTimestamp()?.toString()
 
     override fun isCellEditable(item: LogStream?): Boolean = false
+    override fun getRenderer(item: LogStream?): TableCellRenderer? = renderer
 }
 
 class LogGroupTableSorter(model: ListTableModel<LogStream>) : TableRowSorter<ListTableModel<LogStream>>(model) {
```
abrooksv: What is behavior around scrolling if we change the sort and load more data?

hunterwerlla: This table can't currently load more data: it loads in all of the log groups. In a later PR the table needs to be extended to do that, but for right now I have no idea what would happen.

abrooksv: That makes sense...I'm getting my PRs confused

PR #1612, opened 2020-03-23
Fix Rider compilation in 2020.1

abrooksv:
`@NotNull` was being compared with `@Nullable`

## License

### DIFF
--- a/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
+++ b/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
@@ -80,7 +80,7 @@ class DotNetDebuggerSupport : DebuggerSupport() {
         val useDotnetCoreRuntime
             get() = Registry.`is`(USE_DOTNET_CORE_RUNTIME_FLAG_NAME) && isRider20193OrLater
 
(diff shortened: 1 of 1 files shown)


jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
```diff
@@ -80,7 +80,7 @@ class DotNetDebuggerSupport : DebuggerSupport() {
         val useDotnetCoreRuntime
             get() = Registry.`is`(USE_DOTNET_CORE_RUNTIME_FLAG_NAME) && isRider20193OrLater
 
-        val isRider20193OrLater = ApplicationInfo.getInstance().build >= BuildNumber.fromString("193.0")
+        val isRider20193OrLater = ApplicationInfo.getInstance().build >= BuildNumber("", 193, 0)
```
kiiadi: I just came across this also and my fix was :
`BuildNumber.fromString("193.0")?.let { ApplicationInfo.getInstance().build >= it } == true`

`fromString` seems to have some complex heuristic - didn't want to completely circumvent that.

abrooksv: That seems safe, it falls back to 2019.2 mode if any issue

PR #1615, opened 2020-03-23
CloudWatchLogs fix graphical issues

hunterwerlla:
- Fix speed search and normal selection highlight on group and stream
- Fix trying to load logs backwards every frame when logs load (lol)
- Fix wrapping not properly refreshing the table
- Fix flickering when wrap is turned on
- Use same resizing renderer for timestamp groups in stream
- Remove weird icon from show logs around

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/LogStreamTable.kt
```diff
@@ -72,16 +71,20 @@ class LogStreamTable(
         Disposer.register(this@LogStreamTable, logStreamActor)
         channel = logStreamActor.channel
 
-        component.verticalScrollBar.addAdjustmentListener {
-            if (logsTable.model.rowCount == 0) {
-                return@addAdjustmentListener
+        component.verticalScrollBar.addAdjustmentListener(object : AdjustmentListener {
+            var lastAdjustment = component.verticalScrollBar.minimum
+            override fun adjustmentValueChanged(e: AdjustmentEvent?) {
```
rli: don't think this is actually nullable

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -40,32 +42,49 @@ class LogGroupTableSorter(model: ListTableModel<LogStream>) : TableRowSorter<Lis
...
+
     override fun valueOf(item: LogStreamEntry?): String? = item?.message
     override fun isCellEditable(item: LogStreamEntry?): Boolean = false
     override fun getRenderer(item: LogStreamEntry?): TableCellRenderer? = renderer
 }
 
 private class WrappingLogStreamMessageRenderer : TableCellRenderer {
+    var wrap = false
+    // JBTextArea has a different font from JBLabel (the default in a table) so harvest the font off of it
+    private val font = JBLabel().font
```
abrooksv: This needed with a JBTextArea?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/LogStreamTable.kt
hunterwerlla: I don't think it will ever actually be null but the Java interface doesn't have anything to enforce that so I'd prefer to be defensive

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
hunterwerlla: Tried using a JLabel and HTML but it doesn't adjust its preferred size, we need to reexamine how this is done later

PR #1625, opened 2020-03-25
design: Partition/Region selector UI

rli:


designs/partition-ui/partition-ui.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Overview
+Users want the JetBrains toolkit to support other regions outside the classic partition (i.e. China and GovCloud regions).
+
+When a user wants to switch to a different partition, they will need to change their region in addition to their credentials.
+
+This is a quick, point-in-time document with several directions that we can go. Any decisions made here are not one-way doors and we can adjust the experience as necessary
+if there is any user feedback.
+
+## Mockups
```
awschristou: All of these plans unconditionally show all regions from all partitions. Is that manageable from a user's perspective?

Is this PR intended to be a functional spec for the first release, and then make improvements over time, or is the intention to collaborate towards a first design?

abrooksv:
There is a 5th as well.
Other becomes a popup menu like in 4, but that is another pop out menu that shows the regions in that partiton. So there is no intermediate setting required

designs/partition-ui/partition-ui.md
rli: The desire is to show all partitions/regions only at initial selection phase and if the user wants a region that isn't in their recents list (see screenshot below for this case).

I think this is more of a first release type of thing so customers have something usable and we can make improvements as we get feedback

image

rli:
> There is a 5th as well.
> Other becomes a popup menu like in 4, but that is another pop out menu that shows the regions in that partiton. So there is no intermediate setting required

added

designs/partition-ui/partition-ui.md
```diff
@@ -0,0 +1,37 @@
...
+![3]
+
+1. Same as above, but partition is a submenu
+![4]
+
+1. Expose regions as a nested submenu of the partition submenu
+![5]
+
+ 
+[1]: img/1-always-show-partition.png
```
hunterwerlla: I don't know if we want to make this a long lived doc, but if we do can we add the source for the images? (I'm assuming they are not hand drawn, if not please ignore)

justinmk3:
+1 for partition-as-submenu, which then enables those partitions in the main region-selection menu. 

- The extra friction of a submenu makes sense here, because partition-selection is unusual.
- Once a user "enables" certain regions, they should no longer need to dig into a submenu. Thus it makes sense to list the enabled regions in the main region list.

kiiadi:
My vote is 4) - the partition sub-menu that has AWS Standard selected by default

hunterwerlla:
> +1 for partition-as-submenu, which then enables those partitions in the main region-selection menu.
> 
>     * The extra friction of a submenu makes sense here, because partition-selection is unusual.
> 
>     * Once a user "enables" a partition, they should no longer need to dig into a submenu. Thus it makes sense to list the enabled regions in the main region list.
> 
> 
> Side-note: I agree with @awschristou that we may later want to revisit showing all "standard" regions unconditionally.

I agree with the submenu. I'm not a fan of 1 because it's really hard to have to hunt for regions already, and 2, 3, and 4 seems a bit weird to me that we are showing a filter alongside the rest of the list.

PR #1626, opened 2020-03-25
Use JDK 11 on Windows, change how we install dotnet

abrooksv:
## License


justinmk3:
see also https://github.com/aws/aws-toolkit-jetbrains/pull/1506/

PR #1627, opened 2020-03-25
Attempt to locate dotnet SDK dynamically

abrooksv:
Since we now use runtime versions on Windows, they can change the minor version on us that breaks the path. Attempt to locate it by searching the file system

## License

### DIFF
--- a/jetbrains-rider/tst/base/AwsMarkupBaseTest.kt
+++ b/jetbrains-rider/tst/base/AwsMarkupBaseTest.kt
@@ -3,6 +3,8 @@
 
 package base
 
+import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.execution.util.ExecUtil
 import com.intellij.openapi.util.SystemInfo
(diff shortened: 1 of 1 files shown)


buildspec/windowsTests.yml
```diff
@@ -20,7 +20,7 @@ phases:
 
   build:
     commands:
-      - ./gradlew check coverageReport --info --full-stacktrace --console plain
+      - ./gradlew :jetbrains-rider:check --info --full-stacktrace --console plain
```
hunterwerlla: Why are we only building rider on windows?

abrooksv: Sorry WIP PR that I forgot to mark

abrooksv: Trying to get this PR to build on top of #1613. If I can get all tests to pass here, I will merge the other PR and then merge this right after

abrooksv:
There is a bug in Rider SDK 2020.1, so Windows 2020.1 will fail. All other tests should pass now

PR #1628, opened 2020-03-25
WIP: ECS Debug-enabled Containers as Target Environments

kiiadi:
Hacked together POC that shows we should be able to use containers that have been enabled for cloud-debug as "Target Environments" using the 2020.1 experimental feature.


kiiadi:
oops - meant to open this as a draft.

PR #1630, opened 2020-03-25
CloudWatch Logs Lambda and ECS integration

hunterwerlla:
## Description
- Move off of the CloudWatchLogsAsync client prompted by adding the error handling for log stream not existing. It was throwing an exception wrapped in a `CompletionException` wrapped in a `CompletionException` and since we are only using one at a time there's not a big benefit to using the async client.
- Add action to Lambda to view log group for that lambda
- Fix ECS to not throw errors when the log stream does not actually exist. Currently it will show on the table that it does not exist which I personally prefer to showing an error message, but would like feedback on what behavior we want.
- After looking at Lambda, changed the name of actions to align better with the rest of our code
## Screenshots
Kapture 2020-03-25 at 9 37 52

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/LogStreamActor.kt
```diff
@@ -155,8 +158,9 @@ class LogStreamFilterActor(
         return getSearchLogEvents(request)
     }
 
-    private suspend fun getSearchLogEvents(request: FilterLogEventsRequest): List<LogStreamEntry> {
-        val response = client.filterLogEvents(request).asDeferred().await()
+    private fun getSearchLogEvents(request: FilterLogEventsRequest): List<LogStreamEntry> {
+        // withContext is required to give the execution the same scope as us so if it throws and exception we can catch it
```
abrooksv: Comment seems old and not related to new code

hunterwerlla: That comment was dangling from when it was using the async client, fixed

PR #1654, opened 2020-03-30
Fix localized download text

hunterwerlla:
- Fix some more gratuitous title case
- Set default file name as log stream name
- Set CloudWatch Download text properly
## License


resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -598,7 +598,8 @@ cloud_debug.instrument.production_warning.checkbox_label=I confirm {0} is not a
 cloud_debug.instrument.production_warning.title=Warning!
 lambda.run_configuration.debug_host=Debug Host:
 lambda.sam.debugHost.tooltip=The host to connect the debugger to
-cloudwatch.logs.save_action=Save To A File
+cloudwatch.logs.save_action=Save to a File
+cloudwatch.logs.download=Download
```
abrooksv: I would put a better description like we did for S3.

Also what happens if they select a folder

hunterwerlla: Actually, I don't think this text is used lol

If they select a folder it is saved into that folder as expected

abrooksv: It is used in the non-native browser.

abrooksv: You can force it by calling setForcedToUseIdeaFileChooser(true) on the FileSaverDescriptor to see what it looks like

hunterwerlla: Cool, updated the text to be more descriptive

PR #1663, opened 2020-03-31
Fix two executable manager bugs

hunterwerlla:
- First bug, we were checking for ExecutableWithPath and autoresolved and newer, but this would work for bad paths which would then try to validate and save, eventually saving an invalidexecutable. This invalid path would get spit back out later becasue of bug 2.
- Second bug, an InvalidPath is an ExecutableInstance.Executable, so it would get spit back out without revalidation on the async getExecutable
- Add some more tests for this behavior, any suggestions of more tests to add would be great. updatingExecutableRevalidates tests this change explicitly.

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableManager.kt
```diff
@@ -123,9 +123,11 @@ class DefaultExecutableManager : PersistentStateComponent<ExecutableStateList>,
 
             future.complete(
                 when {
-                    instance is ExecutableWithPath && persisted.autoResolved == true && instance.executablePath.isNewerThan(lastKnownFileTime) ->
+                    // If this is an autoresolved path, and has been updated, remove the autoresolvedness and update
+                    instance is ExecutableInstance.Executable && persisted.autoResolved == true && instance.executablePath.isNewerThan(lastKnownFileTime) ->
```
kiiadi, line 127: what happens to executeables that have bene resolved but are invalid?

hunterwerlla: They are attempted to be loaded again with `load`. Hm looking at this again it seems like the real issue is with the second line, so reverting this line change

PR #1665, opened 2020-04-01
Try to stabilize our S3 integration tests

abrooksv:
* Rewrote our waiter to not returns strings on failures, callee code should worry about that
* Wait until head bucket works after creating a bucket to try to stabilize our tests

## License


core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -3,14 +3,87 @@
...
+                }
+
+                delay(calculateBackOff(attempt))
+            }
+
+            throw WaiterTimeoutException("Condition was not met after $attempt attempts (${Duration.between(start, Instant.now()).seconds} seconds)")
+        }
+    }
+
+    private fun calculateBackOff(attempt: Int) = 250L shl min(attempt - 1, 4) // Max = 250 * 2^4 = 4_000.
```
hunterwerlla, line 77: That's the max, but what strategy is captured here? It looks like exponential backoff? also backoffms?

core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -3,14 +3,87 @@
...
+
+object Waiters {
+    private val LOG = getLogger<Waiters>()
+
+    /**
+     * Generic waiter method to wait for a certain AWS resource to be in a steady status until it reaches in a failed status
+     * or time out for pulling.
+     * @param T The AWS resource type for querying the status from it
+     */
+    fun <T> waitUntil(
```
hunterwerlla, line 23: Do we want to make our waiter suspend then allow callers to runBlocking if they want?

core/src/software/aws/toolkits/core/utils/Waiter.kt
```diff
@@ -3,14 +3,87 @@
...
+        // The success predicate based on the returned resource
+        succeedOn: (T) -> Boolean = { true },
+        // The fail predicate based on the returned resource. Return an error message if it fails, null otherwise
+        failOn: (T) -> Boolean = { false },
+        exceptionsToStopOn: Set<KClass<out Exception>> = emptySet(),
```
hunterwerlla, line 28: it's not immediately clear what this means? shouldn't we stop on all non ignored exceptions?

core/src/software/aws/toolkits/core/utils/Waiter.kt
abrooksv: I stole it from the Java SDK test util waiter...its exponential backoff

core/src/software/aws/toolkits/core/utils/Waiter.kt
abrooksv: Not sure, puts more bloat into the call sites.... seems like the name would indicate it should always block

core/src/software/aws/toolkits/core/utils/Waiter.kt
abrooksv: Ya ill document the params

core/src/software/aws/toolkits/core/utils/Waiter.kt
hunterwerlla: Right but can we add a comment about it + that it's in MS

core/src/software/aws/toolkits/core/utils/Waiter.kt
abrooksv: And no, things like IOExceptions should be retried, its more like a if we trigger a delete we want to keep retrying until we get a ResourceNotFoundException so we should stop then

core/src/software/aws/toolkits/core/utils/Waiter.kt
hunterwerlla: it puts more bloat on non-coroutine call sites, but shouldn't this be called from coroutines like 90%+ of the time?

core/src/software/aws/toolkits/core/utils/Waiter.kt
hunterwerlla: Cool

PR #1666, opened 2020-04-01
CloudWatch UX feedback

hunterwerlla:
- On search, X wasn't doing anything. Fix it by adding another listener.
- Make labels consistent "Show" to "View" to match our other labels
- Fix labels that referred to "log groups" that are actually log streams

## Screenshots
Kapture 2020-04-01 at 9 14 18

## License


resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -561,12 +560,12 @@ ecs.service.logs.cannot_determine_log_stream=Cannot determine log streams for se
 ecs.service.logs.cannot_find_container_log_stream=Cannot determine log stream for container {0}.
 ecs.service.logs.no_running_tasks=Service has no running tasks.
 ecs.service.logs.no_log_stream=ECS has not sent events to CloudWatch
-ecs.service.logs.action_label=Show logs
+ecs.service.container_logs.action_label=View Container Log Stream
```
kiiadi, line 563: We might want to put this back ultimately - because we probably want a 'if there's one task, open the stream directly, else open the log group' kinda deal.

hunterwerlla: Yep, we need to expand the behavior of it

PR #1668, opened 2020-04-01
Add result metric when submitting feedback

rli:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/feedback/FeedbackDialog.kt
```diff
@@ -52,7 +54,9 @@ class FeedbackDialog(private val project: Project) : DialogWrapper(project), Cor
                         setOKButtonText(message("feedback.submit_button"))
                         isOKActionEnabled = true
                     }
+                    FeedbackTelemetry.result(project, success = false)
                 }
+                FeedbackTelemetry.result(project, success = true)
```
hunterwerlla, line 59: uh won't this post feedback success even when it's failed? this should be in the try?

rli: oops

PR #1672, opened 2020-04-02
CloudWatch Logs replace second copy of open in log stream action with copy action

hunterwerlla:
## Screenshots
Before:
Screen Shot 2020-04-02 at 10 33 22 AM

After:
Screen Shot 2020-04-02 at 11 21 55 AM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/actions/CopyFromTableAction.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.ide.CopyPasteManager
+import com.intellij.openapi.project.DumbAware
+import com.intellij.ui.table.TableView
+import software.aws.toolkits.jetbrains.services.cloudwatch.logs.LogStreamEntry
+import software.aws.toolkits.resources.message
+import java.awt.datatransfer.StringSelection
+
+// FIX_WHEN_MIN_IS_201 make anaction text dynamic based on table size
```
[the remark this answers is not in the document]
hunterwerlla: Honestly forgot that was a thing, fixed. Keeping the FIX_WHEN_MIN_IS_201 since that seems like a better way to do it in the future

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/actions/CopyFromTableAction.kt
```diff
@@ -0,0 +1,35 @@
...
+    DumbAware {
+    override fun update(e: AnActionEvent) {
+        e.presentation.text = message("cloudwatch.logs.copy_action", table.selectedRows.size)
+    }
+
+    override fun actionPerformed(e: AnActionEvent) {
+        val copyPasteManager = CopyPasteManager.getInstance()
+        // This emulates the copy that that comes from the jtable which is TSV. We could
+        // pull the action out of the table but it would be more weird
+        val selection = table.selectedRows.joinToString("\n") { row ->
```
abrooksv, line 28: I would use buildString to avoid the create-copy-throw away intermediate strings

hunterwerlla: joinToString uses a string builder so this makes lines+1 string builders. I don't want to append a "\n" or "\t" to the last object so keeping track of which item is the last one in a string builder is probably more trouble than it's worth.

PR #1677, opened 2020-04-03
Refactor how we pass around SdkHttpClient

abrooksv:
* Do not pass it as a constructor arg to prevent starting it up too early
* Use a supplier to the credentials system since it is only needed for STS calls and not always, this prevents spinning it up in tests as much
* Cut down on chances of triggering a thread leak in unit tests
* Tests that do need it should manually create and dispose it

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
@@ -153,7 +153,7 @@ class ProfileCredentialProviderFactory : CredentialProviderFactory, Disposable {
     override fun createAwsCredentialProvider(
         providerId: ToolkitCredentialsIdentifier,
         region: AwsRegion,
-        sdkClient: SdkHttpClient
+        sdkHttpClientSupplier: () -> SdkHttpClient
     ): AwsCredentialsProvider {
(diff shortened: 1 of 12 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -153,15 +153,15 @@ class ProfileCredentialProviderFactory : CredentialProviderFactory, Disposable {
...
+        sdkHttpClientSupplier: () -> SdkHttpClient
     ): AwsCredentialsProvider {
         val profileProviderId = providerId as? ProfileCredentialsIdentifier
             ?: throw IllegalStateException("ProfileCredentialProviderFactory can only handle ProfileCredentialsIdentifier, but got ${providerId::class}")
 
         val profile = profileHolder.getProfile(profileProviderId.profileName)
             ?: throw IllegalStateException("Profile ${profileProviderId.profileName} looks to have been removed")
 
-        return createAwsCredentialProvider(profile, region, sdkClient)
+        return createAwsCredentialProvider(profile, region, sdkHttpClientSupplier.invoke())
```
hunterwerlla, line 164: curious why.invoke vs ()?

abrooksv: No reason

PR #1680, opened 2020-04-03
Fix runUnderRealCredentials failing to switch back to fake creds

abrooksv:
When we reset the cred manager, it causes a switch to dummy creds due to the active has now been removed. This ends up getting cancelled by the explicit switch back to oldActive which is the same (DUMMY). So instead switch the active, before the removal of the real creds

## License


abrooksv:
This should fix some of our flaky cloud debug tests

PR #1684, opened 2020-04-03
Add speed search to logstream

hunterwerlla:
Screen Shot 2020-04-06 at 8 18 15 AM
Screen Shot 2020-04-06 at 8 38 45 AM

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
@@ -3,8 +3,10 @@
 
 package software.aws.toolkits.jetbrains.services.cloudwatch.logs.editor
 
+import com.intellij.ui.SimpleColoredComponent
 import com.intellij.ui.components.JBLabel
 import com.intellij.ui.components.JBTextArea
+import com.intellij.ui.speedSearch.SpeedSearchUtil
 import com.intellij.util.text.DateFormatUtil
 import com.intellij.util.text.SyncDateFormat
 import com.intellij.util.ui.ColumnInfo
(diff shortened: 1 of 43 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -155,3 +167,21 @@ private fun Component.setSelectionHighlighting(table: JTable, isSelected: Boolea
         background = table.background
     }
 }
+
+private class SpeedSearchHighlighter : Highlighter.HighlightPainter {
+    override fun paint(g: Graphics?, p0: Int, p1: Int, bounds: Shape?, component: JTextComponent?) {
```
abrooksv, line 172: What does it do on a wrap boundary like

```
aaaaaaa Fo
o
```

and search for foo

hunterwerlla: it doesn't because we wrap on words (woo dodging that problem)

hunterwerlla: There are a lot more things we need to do to fix wrapping, disabled while wrapping. I think it's still worth doing even if it only works for unwrapped because that is the default state.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -155,3 +167,23 @@ private fun Component.setSelectionHighlighting(table: JTable, isSelected: Boolea
...
```
abrooksv, line 172: Can we fix the bad default param names?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -72,3 +81,51 @@ fun JComponent.validationInfo(message: String) = when {
...
+fun drawSearchMatch(
+    g: Graphics2D,
+    startX: Float,
+    endX: Float,
+    startY: Float,
+    height: Int
+) {
+    val c1: Color =
+        JBColor.namedColor("SearchMatch.startBackground", JBColor.namedColor("SearchMatch.startColor", 0xffeaa2))
+    val c2: Color =
```
abrooksv, line 98: Can we drop the `:Color` part?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
hunterwerlla: Sure, might as well since we have to implement it ourselves

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
hunterwerlla: Yep, removed

Issue #1394, opened 2019-11-26
Support Deployment of Nested Applications (CAPABILITY_AUTO_EXPAND)


dfreemyer:
**Is your feature request related to a problem? Please describe.**
I'm working on a project that I would like to keep organized by nesting applications under a master template. Deployment via "Deploy Serverless Application" fails with the nested applications in status CREATE_FAILED because "Requires Capabilities: [CAPABILITY_AUTO_EXPAND]".

There seems to be no way to modify the command used by the "Deploy Serverless Application" in the context menu to include the additional capability required by CloudFormation.

**Describe the solution you'd like**
In the Deploy Serverless Application dialog, there could be an option to preview the commands that will be run and give the user a chance to edit the arguments as needed. 

Or better yet, there could be a project-wide setting somewhere, kind of like the credential chooser, where users could enable/disable capabilities a-la-carte. 

**Describe alternatives you've considered**
It looks like my only options are to deploy from the command line or reconsider my choice to nest templates. I really like being able to quickly deploy from within the ide (PyCharm in my case)

rli:
Thanks for the feedback. We'll re-evaluate the UX here.

Naturious:
Any advancement on this front?

hunterwerlla:
@Naturious This will be in the next release, and will be available in next weeks EAP release

abrooksv:
We have released support in 1.13, please let us know if there are any issues

PR #1698, opened 2020-04-08
Expose CloudFormation capabilities 

hunterwerlla:
- Expose capabilities as independent checkboxes when deploying, and save them to deployment settings
- Remove very weird spacers from the deployment window that made the template parameters box look compressed

## Related Issue(s)
#1394 

Screen Shot 2020-04-08 at 4 42 55 PM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/CreateCapabilities.kt
```diff
@@ -0,0 +1,41 @@
...
+        true
+    ),
+    NAMED_IAM(
+        "CAPABILITY_NAMED_IAM",
+        message("cloudformation.capabilities.named_iam"),
+        message("cloudformation.capabilities.named_iam.toolTipText"),
+        true
+    ),
+    AUTO_EXPAND(
+        "CAPABILITY_AUTO_EXPAND",
```
abrooksv, line 23: Can we use the actual enum here? (Capability.CAPABILITY_AUTO_EXPAND.value)

abrooksv: Or better yet, the enum itself and convert to the value string when we need it?

hunterwerlla: `value` is actually private, but the toString works, changed. This is really the point where we need it so I don't think we need to store anything other than the string.

kiiadi:
@bryceitoc9 @justinmk3 @awschristou  how are we handling this in vscode?

PR #1703, opened 2020-04-09
Preserve tree state when refreshing all or part of the explorer

hunterwerlla:
## Related issue
#1704
### Refresh whole tree
Kapture 2020-04-09 at 13 30 26
### Delete an object (takes a while because of S3)
Kapture 2020-04-09 at 13 32 49

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
@@ -96,13 +96,14 @@ class ExplorerToolWindow(private val project: Project) : SimpleToolWindowPanel(t
     }
 
     /**
-     * Invalidates tree nodes, causing IntelliJ to redraw the tree
+     * Invalidates tree nodes, causing IntelliJ to redraw the tree. Preserves node state.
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
```diff
@@ -95,18 +96,21 @@ class ExplorerToolWindow(private val project: Project) : SimpleToolWindowPanel(t
...
-     * Invalidates tree nodes, causing IntelliJ to redraw the tree
+     * Invalidates tree nodes, causing IntelliJ to redraw the tree. Preserves node state.
      * Provide an AbstractTreeNode in order to redraw the tree from that point downwards
-     * Otherwise redraws (and collapses) the entire tree
+     * Otherwise redraws the entire tree
      *
      * @param selectedNode AbstractTreeNode to redraw the tree from
      */
     fun invalidateTree(selectedNode: AbstractTreeNode<*>? = null) {
+        // save the state and reapply it after we invalidate (which is the point where the state is wiped)
```
justinmk3, line 106:
```suggestion
        // Save the state and reapply it after we invalidate (which is the point where the state is wiped).
        // Items are expanded again if their user object is unchanged (.equals()).
```

PR #1706, opened 2020-04-10
Paginate Log Group View

hunterwerlla:
- Paginate the Log Group view
- Break out loading on scroll into its own utils file
- Further genericize `LogStreamActor` into `LogActor`
- Prevent user column resize or sort
## Screenshots
Kapture 2020-04-10 at 12 47 39

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1709, opened 2020-04-14
Search from log group

hunterwerlla:
- Part two of #1706 , add search to log group view.
- Comes with bonus cleaned up tests
Kapture 2020-04-14 at 16 00 24

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
```diff
@@ -74,6 +74,14 @@ class LogGroupTableSorter(model: ListTableModel<LogStream>) : TableRowSorter<Lis
     }
 }
 
+class LogGroupFilterTableSorter(model: ListTableModel<LogStream>) : TableRowSorter<ListTableModel<LogStream>>(model) {
```
abrooksv, line 77: Why do we disable sort if its filtered?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/SearchFieldUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+            }
+            lastText = searchFieldText
+            if (text.isEmpty()) {
+                block()
+            }
+        }
+    })
+}
+
+fun SearchTextField.onTextChanged(block: () -> Unit) {
```
abrooksv, line 28: Should we call this onEnter?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/editor/TableUtils.kt
hunterwerlla: We don't, we change sort to column 0 because the search api is limited to sorting by name. Sort is disabled in both cases because otherwise loading more would make no sense.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/SearchFieldUtils.kt
hunterwerlla: Yep that would make more sense

PR #1715, opened 2020-04-19
Add CloudFormation linter

miparnisari:
## Types of changes
- [ ] Bug fix

## Description
Add CloudFormation linter. Among other things, it provides:
-  instance type and resource type regional availability
- IAM action validation
- conditionally required properties and constraints for properties
- warning about approaching template limits
- best practices like properties that shouldn't be hardcoded and resource types that should have `DeletionPolicy / UpdateReplacePolicy`

How this PR works: it's a file annotator. If the file currently open isn't recognized as a CloudFormation template, it doesn't do anything. Otherwise, it tries to run the template through [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint) (a Python package). If errors are found, they are displayed inline.

## Motivation and Context
There are plugins out there that do this same thing, e.g. https://github.com/binxio/cfn-lint-plugin. I would like to have this functionality within the official AWS plugin.

## Related Issue(s)
N/A

## Testing
Tested on Mac OS with IntelliJ Ultimate 2020 and on Windows 10.

## Screenshot

image

image

## Checklist

## License


CONTRIBUTING.md
```diff
@@ -101,6 +101,7 @@ To test your changes locally, you can run the project from IntelliJ or gradle.
   ```
   - Requires valid AWS credentials (take care: it will respect any credentials currently defined in your environmental variables, and fallback to your default AWS profile otherwise).
   - Requires `sam` CLI to be on your `$PATH`.
+  - Requires `cfn-lint` CLI to be on your `$PATH`.
```
PatMyron, line 104:
```suggestion
  - Requires [`sam`](https://github.com/awslabs/serverless-application-model) CLI to be on your `$PATH`.
  - Requires [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/) CLI to be on your `$PATH`.
```

jetbrains-core/resources/META-INF/plugin.xml
```diff
@@ -224,6 +224,14 @@
...
         <webHelpProvider implementation="software.aws.toolkits.jetbrains.core.help.HelpIdTranslator"/>
         <fileEditorProvider implementation="software.aws.toolkits.jetbrains.services.s3.S3ViewerEditorProvider"/>
+
+        <!-- CloudFormation template linting -->
+        <externalAnnotator language="JSON"
+                           id="annotatorJSON"
+                           implementationClass="software.aws.toolkits.jetbrains.services.cloudformation.annotations.ErrorAnnotator"/>
+        <externalAnnotator language="yaml"
+                           id="annotatorYAML"
+                           implementationClass="software.aws.toolkits.jetbrains.services.cloudformation.annotations.ErrorAnnotator"/>
```
PatMyron, line 234: Does this mean [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/) is run on all `.json`/`.yaml` files?
[Could borrow some CloudFormation template recognition logic from the Visual Studio Code extension to check for either `AWSTemplateFormatVersion ` or `Resources` and `Type`](https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/blob/edd51fb86441a1c48d294c4b664189c6f4d11399/server/src/server.ts#L125)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.services.cloudformation.annotations
+
+import com.intellij.execution.configurations.PathEnvironmentVariableUtil
+import com.intellij.openapi.util.SystemInfo
+
+class LinterExecutable {
+
+    companion object {
+        private val PLUGIN_NAME = "cfn-lint"
```
PatMyron, line 11:
```suggestion
        private val EXECUTABLE_NAME = "cfn-lint"
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+
+        @Throws(Exception::class)
+        fun getExecutablePath(): String {
+            var executable: String? = null
+            if (SystemInfo.isWindows) {
+                executable = PathEnvironmentVariableUtil.findExecutableInWindowsPath(PLUGIN_NAME)
```
PatMyron, line 17:
```suggestion
                executable = PathEnvironmentVariableUtil.findExecutableInWindowsPath(EXECUTABLE_NAME)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+            } else {
+                val executableFile = PathEnvironmentVariableUtil.findInPath(PLUGIN_NAME)
```
PatMyron, line 19:
```suggestion
                val executableFile = PathEnvironmentVariableUtil.findInPath(EXECUTABLE_NAME)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+                if (executableFile != null) {
+                    executable = executableFile.absolutePath
+                }
+            }
+            if (executable == null || executable.isEmpty()) {
+                throw Exception("Couldn't find the cfn-lint plugin in the PATH. Please install the cfn-lint plugin.")
```
PatMyron, line 25:
```suggestion
                throw Exception("Couldn't find cfn-lint in the PATH. Please install cfn-lint")
```

This is about the CLI executable rather than the IDE plugin, right? Might be clearer to avoid plugin if so

PatMyron:
It's not well documented, but the [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=kddejong.vscode-cfn-lint) has a `cfnLint.path` setting that actually also accepts the CLI options:

Screen Shot 2020-04-20 at 3 17 35 PM

Something to think about for future additions

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotationTest.kt
```diff
@@ -0,0 +1,55 @@
...
+            "    },\n" +
+            "    \"Message\": \"Invalid or unsupported Type undefined for resource S3Bucket in us-east-1\",\n" +
+            "    \"Rule\": {\n" +
+            "      \"Description\": \"Making sure the basic CloudFormation resources are properly configured\",\n" +
+            "      \"Id\": \"E3001\",\n" +
+            "      \"ShortDescription\": \"Basic CloudFormation Resource Check\",\n" +
+            "      \"Source\": \"https://github.com/aws-cloudformation/cfn-python-lint\"\n" +
+            "    }\n" +
+            "  }\n" +
+            "]\n"
```
PatMyron, line 41: https://realkotlin.com/tutorials/2018-06-26-multiline-string-literals-in-kotlin/

(I think I'm warming up to Kotlin)

miparnisari: TIL!

jetbrains-core/resources/META-INF/plugin.xml
miparnisari: Yes, right now this code runs on all json and yaml files.

We could run some checks beforehand but I'm afraid this will have performance implications, as the plugin would have to read the file every time it changed before sending it over to the linter.

Maybe it would be better to have the linter return as fast as possible once it realises that the file isn't a template. Plus this means that we don't have duplicate checks both here and in cfn-lint.

.travis.yml
```diff
@@ -14,6 +14,8 @@ matrix:
       dist: xenial
```
hunterwerlla, line 14: We are moving away from Travis for tests (it's very unreliable), the tests way pay attention to are in here: https://github.com/aws/aws-toolkit-jetbrains/tree/master/buildspec

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# _1.14_ (2020-04-18)
```
hunterwerlla, line 1: please generate a changelog entry with `./gradlew newChange`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotator.kt
```diff
@@ -0,0 +1,74 @@
...
+     * @param holder a container for receiving annotations
+     */
+    override fun apply(
+        file: PsiFile,
+        annotationResult: List<ErrorAnnotation>,
+        holder: AnnotationHolder
+    ) {
+        val document = FileDocumentManager.getInstance().getDocument(file.virtualFile)
+        annotationResult.forEach { error ->
+            val startOffset = document?.getLineStartOffset(error.location?.start?.lineNumber!! - 1)
```
hunterwerlla, line 64: A good alternitive to !! is to use `lineNumber?.let { ...}` which makes sure `it` is not null or `lineNumber ?: return@forEach`. We are working towards removing all `!!` from prod code.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotator.kt
```diff
@@ -0,0 +1,74 @@
...
+     * @param file a file to annotate
+     * @param annotationResult annotations collected in #doAnnotate
+     * @param holder a container for receiving annotations
+     */
+    override fun apply(
+        file: PsiFile,
+        annotationResult: List<ErrorAnnotation>,
+        holder: AnnotationHolder
+    ) {
+        val document = FileDocumentManager.getInstance().getDocument(file.virtualFile)
```
abrooksv, line 62: You should not need the virtual file here:

`com.intellij.psi.PsiDocumentManager#getDocument` can take a PsiFile

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/Linter.kt
```diff
@@ -0,0 +1,57 @@
...
+        private val LOG = getLogger<Linter>()
+    }
+
+    fun execute(initialAnnotationResults: InitialAnnotationResults): List<ErrorAnnotation> {
+        var errors: List<ErrorAnnotation> = ArrayList()
+
+        try {
+            LOG.info { "Beginning execution of CloudFormation linter" }
+            val executable: String = LinterExecutable.getExecutablePath()
+            val command: MutableList<String> = mutableListOf()
```
abrooksv, line 25: `val command = mutableListOf<String>()`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/Linter.kt
```diff
@@ -0,0 +1,57 @@
...
+            command.add("json")
+            val processBuilder = ProcessBuilder(command)
+            processBuilder.redirectErrorStream(true)
+            val linterProcess = processBuilder.start()
+
+            val reader = BufferedReader(InputStreamReader(linterProcess.inputStream))
+            val linterOutput = StringBuilder()
+            reader.forEachLine {
+                linterOutput.append(it)
+            }
```
abrooksv, line 39: Replace with `ExecUtil#execAndGetOutput`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/Linter.kt
```diff
@@ -0,0 +1,57 @@
...
+import java.io.InputStreamReader
+
+class Linter {
+
+    companion object {
+        private val LOG = getLogger<Linter>()
+    }
+
+    fun execute(initialAnnotationResults: InitialAnnotationResults): List<ErrorAnnotation> {
+        var errors: List<ErrorAnnotation> = ArrayList()
```
abrooksv, line 20: You can avoid the var

you can do 
```
return try {
...
getErrorAnnotations(linterOutput.toString())
} catch(e: Exception) {
log
emptyList()
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/Linter.kt
```diff
@@ -0,0 +1,57 @@
...
+            LOG.error(e) { "Error: " + e.message }
+        }
+        return errors
+    }
+
+    fun getErrorAnnotations(linterOutputJson: String): List<ErrorAnnotation> {
+        val mapper = ObjectMapper()
+        val typeReference = TypeFactory.defaultInstance().constructCollectionType(
+                MutableList::class.java, ErrorAnnotation::class.java
+            )
```
abrooksv, line 53: This should be created once in the Linter companion as a lazy

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+            if (SystemInfo.isWindows) {
+                executable = PathEnvironmentVariableUtil.findExecutableInWindowsPath(EXECUTABLE_NAME)
+            } else {
+                val executableFile = PathEnvironmentVariableUtil.findInPath(EXECUTABLE_NAME)
+                if (executableFile != null) {
+                    executable = executableFile.absolutePath
+                }
+            }
+            if (executable == null || executable.isEmpty()) {
+                throw Exception("Couldn't find cfn-lint in the PATH. Please install cfn-lint")
```
abrooksv, line 25: If this is shown to users, it should be in the localization files

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+            var executable: String? = null
+            if (SystemInfo.isWindows) {
+                executable = PathEnvironmentVariableUtil.findExecutableInWindowsPath(EXECUTABLE_NAME)
+            } else {
+                val executableFile = PathEnvironmentVariableUtil.findInPath(EXECUTABLE_NAME)
+                if (executableFile != null) {
+                    executable = executableFile.absolutePath
+                }
+            }
+            if (executable == null || executable.isEmpty()) {
```
abrooksv, line 24: You can use `executable.isNullOrEmpty()`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterExecutable.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.intellij.openapi.util.SystemInfo
+
+class LinterExecutable {
+
+    companion object {
+        private val EXECUTABLE_NAME = "cfn-lint"
+
+        @Throws(Exception::class)
+        fun getExecutablePath(): String {
+            var executable: String? = null
```
abrooksv, line 15: Use assignment from the if, instead of a var

```
val executable =  if (SystemInfo.isWindows) {
PathEnvironmentVariableUtil.findExecutableInWindowsPath(EXECUTABLE_NAME)
            } else {
                val executableFile = PathEnvironmentVariableUtil.findInPath(EXECUTABLE_NAME)
           
                    executable = executableFile?.absolutePath
                }
            }
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorLocation.kt
```diff
@@ -0,0 +1,21 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.services.cloudformation.annotations
+
+import com.fasterxml.jackson.annotation.JsonCreator
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ErrorLocation @JsonCreator constructor(
```
abrooksv, line 10: By doing this you have made this your API forever due to you have no version checking.

Are you okay with this?

jetbrains-core/resources/META-INF/plugin.xml
abrooksv: The downsides of not using a required marker or file extension :/

Should some form of warning be added to cfn-lint push people back to a better path by specifying AWSTemplateFormatVersion?

This right now will invoke the process for every JSON/YAML file they open (and when they type in them as well)

Merging in support for CFN as a file type would act as the gate in the future though.

jetbrains-core/it/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotatorTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.nhaarman.mockitokotlin2.whenever
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito
+import java.io.File
+import java.io.IOException
+
+class ErrorAnnotatorTest {
```
abrooksv, line 19: We never mock the JB APIs, their behavior is too undocumented and can change between versions.

`JavaCodeInsightTestFixtureRule` can be used to load any file (its not really Java tied, it just has a Java SDK configured in the project) and can be used to run annotators against loaded documents. Though I have never used it against an external annotator, but it should work.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotator.kt
```diff
@@ -0,0 +1,74 @@
...
+ * External annotators are expected to be (relatively) slow and are started
+ * after regular annotators have completed their work.
+ *
+ * Annotators work in three steps:
+ * 1. [.collectInformation] is called to collect some data about a file needed for launching a tool
+ * 2. collected data is passed to [.doAnnotate] which executes a tool and collect highlighting data
+ * 3. highlighting data is applied to a file by [.apply]
+ *
+ */
+class ErrorAnnotator : ExternalAnnotator<InitialAnnotationResults, List<ErrorAnnotation>>() {
```
hunterwerlla, line 23: ErrorAnnotator is a very generic name, since it's loaded into the IDE as an extension can we give it a full name like `CloudFormationLintAnnotator`

jetbrains-core/resources/META-INF/plugin.xml
hunterwerlla: I would personally prefer making a file type, [here's how JB CF plugin does it](https://github.com/shalupov/idea-cloudformation/blob/3195bad973e5b1b4b0913d41d2eb9b6f099ea9cc/src/main/kotlin/com/intellij/aws/cloudformation/CloudFormationFileTypeDetector.kt) . The detection heuristic can probably be improved.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorAnnotator.kt
miparnisari: Ahh but i'm not using `PsiDocumentManager` since it requires the `Project` and I don't know how to access that in this context.

.travis.yml
miparnisari: Not sure how to install cfn-lint for the Windows tests. cfn-lint isn't available in chocolatey.

miparnisari:
> How does a user know their cfn-lint worked? We don't seem to give any indication that cfn-lint could not be found that is visible to the user.

image

abrooksv:
> > How does a user know their cfn-lint worked? We don't seem to give any indication that cfn-lint could not be found that is visible to the user.
> 
> Currently, an IDE error is thrown:
> 
> image
> 
> But ideally something more user friendly would show up...

This will need to be changed to be an opt-in feature that gracefully degrades. A Fatal IDE Error is a blocker to shipping.

miparnisari:
@abrooksv my original idea was to have a configuration window where you can specify the path to the linter, and if you don't specify it, it will silently fail (i.e. no IDE errors, no linting errors, nothing). Do you think this works? If so, do you have any pointers of where in the code I should add this configuration?

abrooksv:
> @abrooksv my original idea was to have a configuration window where you can specify the path to the linter, and if you don't specify it, it will silently fail (i.e. no IDE errors, no linting errors, nothing). Do you think this works? If so, do you have any pointers of where in the code I should add this configuration?
 
We do this for the SAM executable. I would start here: `software.aws.toolkits.jetbrains.services.lambda.sam.SamExecutable`

jetbrains-core/resources/META-INF/plugin.xml
miparnisari: I have updated the code so that the linter only runs if `AWSTemplateFormatVersion` or `Resources` are found.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/ErrorLocation.kt
miparnisari: I added version checking, thanks.

miparnisari:
Latest updates:
- added the linter to the configuration window (plus fixed a tiny bug in the explorer window screen). See the original PR details for a screenshot.
- improved the template detection. The linter only runs if `AWSTemplateFormatVersion` or `Resources` are found. @PatMyron maybe we should add a "suggestion" rule to the linter so that people start using `AWSTemplateFormatVersion` more
- i'm checking that `cfn-lint`'s version is from 0.2.0 to 1.0.0 (exclusive). @PatMyron this means that if we change the format of the linter's JSON output, we _have_ to do a major bump to 1.x, or we'll break this plugin

Some things missing:
1. `cfn-lint --version` sometimes takes a while to run and I see an error in the logs (not in the IDE) about a `TimeoutException`. @abrooksv is it safe to ignore this?

    I saw that SAM CLI had this problem and they set a timeout value but it's not being used: https://github.com/aws/aws-toolkit-jetbrains/blob/0b9642bccfb48b83324bdf931182f92a98edfc31/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamVersionCache.kt#L37-L40

1. From the contributing guide:

   >Dependencies (such as sam or cloud-debug) should fetch/install lazily, when the user interacts with a feature that requires them.

   I am not sure how to do a lazy install of `cfn-lint`.

1. @hunterwerlla please see my comment on the Windows tests. 

1. I noticed something weird when IntelliJ has auto-save enabled: unless you specifically press ctrl+s to save the file, the linter isn't invoked, which means the errors you see may go stale. I haven't looked into whether I can change when the annotator runs when this setting is enabled. I don't think this is a blocker.

I'd appreciate if you guys downloaded this PR and tested it. :)

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/AwsSettingsConfigurable.java
```diff
@@ -84,6 +93,9 @@ private void createUIComponents() {
         samHelp = createHelpLink("lambda.sam.cli.install_url");
         samHelp.setEnabled(false);
         samExecutablePath = createCliConfigurationElement(getSamExecutableInstance(), SAM);
+        cfnLintHelp = createHelpLink("cloudformation.linter.install_url");
+        cfnLintHelp.setEnabled(false);
```
miparnisari, line 97: Doing this disables the link. I am not sure why the other help links are disabled as well. 

image

If I set this to `true` they appear as proper links.

image

abrooksv:
>     1. `cfn-lint --version` sometimes takes a while to run and I see an error in the logs (not in the IDE) about a `TimeoutException`. @abrooksv is it safe to ignore this?
>        I saw that SAM CLI had this problem and they set a timeout value but it's not being used: https://github.com/aws/aws-toolkit-jetbrains/blob/0b9642bccfb48b83324bdf931182f92a98edfc31/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamVersionCache.kt#L37-L40

Ya, looks like it is left over from the migration of SAM one-off management to Executable manager. It's dead code.

ExecutableManager is async and does not have timeouts that I know of.

> 
>     2. From the contributing guide:
>        > Dependencies (such as sam or cloud-debug) should fetch/install lazily, when the user interacts with a feature that requires them.
>        
>        
>        I am not sure how to do a lazy install of `cfn-lint`.

That statement is and/or. CloudDebug auto-installs. SAM detection runs when it is needed.

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/AwsSettingsConfigurable.java
abrooksv: Welp. That's bad.

hunterwerlla:
> > ```
> > 1. `cfn-lint --version` sometimes takes a while to run and I see an error in the logs (not in the IDE) about a `TimeoutException`. @abrooksv is it safe to ignore this?
> >    I saw that SAM CLI had this problem and they set a timeout value but it's not being used: https://github.com/aws/aws-toolkit-jetbrains/blob/0b9642bccfb48b83324bdf931182f92a98edfc31/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamVersionCache.kt#L37-L40
> > ```
> 
> Ya, looks like it is left over from the migration of SAM one-off management to Executable manager. It's dead code.

However we are using the same timeout for SAM, it's just applied to all executables:
`private val DEFAULT_TIMEOUT = Duration.ofSeconds(5)`

miparnisari:
OK. I'm not sure why `cfn-lint --version` is taking so long to run but I assume this error isn't a blocker.

Same goes for this

> Basically what we are trying to avoid is installing 15 tools up front just because someone installs the Toolkit. Instead opt-in to the download by triggering the logic somehow.

This PR doesn't install the linter. It expects people to install it on their own.

So I think that unless you guys have any other comments the only thing left is the Windows tests?

jetbrains-core/it/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotatorTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        """.trimIndent()
+            ) as YAMLFile
+        }
+        val initialAnnotationResults = cloudFormationLintAnnotator.collectInformation(psiFile)
+        assertThat(initialAnnotationResults).isNotNull
+        val errors = cloudFormationLintAnnotator.doAnnotate(initialAnnotationResults)
+        assertThat(errors).isNotNull
+        assertThat(errors.size).isEqualTo(1)
+
+        // TODO test errorAnnotator.apply(), but I would need to be able to mock a static method
```
kiiadi, line 42: What static method would you need to mock?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotator.kt
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Collects full information required for annotation. This method is intended for long-running activities
+     * and will be called outside a read action; implementations should either avoid accessing indices and PSI or
+     * perform needed checks and locks themselves.
+     *
+     * @param initialAnnotationResults initial information gathered by [.collectInformation]
+     * @return annotations to pass to [.apply]
+     */
+    override fun doAnnotate(initialAnnotationResults: InitialAnnotationResults): List<CloudFormationLintAnnotation> {
+        val linter = Linter()
```
kiiadi, line 43: Why does this need to be instantiated every time? Can we make it stateless?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotator.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import com.intellij.lang.annotation.AnnotationHolder
+import com.intellij.lang.annotation.ExternalAnnotator
+import com.intellij.openapi.fileEditor.FileDocumentManager
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiFile
+
+/**
+ *
+ * Process the files in a language by an external annotation tool ("linter").
```
kiiadi, line 13: Do we feel like all these comments are required?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    private val TEMPLATE_VERSION_KEY = "AWSTemplateFormatVersion"
+    private val RESOURCES_KEY = "Resources"
+
+    var isCloudFormationTemplate = false
+
+    val pathToTemplate = psiFile.virtualFile.path
+
+    init {
+        if (psiFile.fileType is JsonFileType) {
```
kiiadi, line 25: Can we rebased this off [feature/cloudformation](https://github.com/aws/aws-toolkit-jetbrains/tree/feature/cloudformation) and use the newly introduced CFN FileTypes (https://github.com/aws/aws-toolkit-jetbrains/commit/3a60a54539686559c9819eb74fa88b6f57c028d9)?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotation.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @JsonProperty(value = "Rule")
+    val linterRule: LinterRule? = null
+
+    @JsonProperty(value = "Message")
+    val message: String? = null
+
+    val severity: HighlightSeverity
+        get() {
+            if (level.toLowerCase().contains("error")) {
```
kiiadi, line 26: this is a pretty classic case for a `when`:

```kt
val lowered = level.toLowerCase()
return when {
  lowered.contains("error") -> HighlightSeverity.ERROR
  lowered.contains("warn") -> HighlightSeverity.WARNING
  else -> HighlightSeverity.INFORMATION
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/Linter.kt
```diff
@@ -0,0 +1,54 @@
...
+import software.aws.toolkits.core.utils.getLogger
+import software.aws.toolkits.core.utils.info
+import software.aws.toolkits.jetbrains.core.executables.CfnLintExecutable
+import software.aws.toolkits.resources.message
+
+class Linter {
+
+    companion object {
+        private val LOG = getLogger<Linter>()
+        private val mapper = ObjectMapper()
```
kiiadi, line 19: might make more sense to use the `com.fasterxml.jackson.module.kotlin.jacksonObjectMapper` which is more Kotlin aware - then you can model your objects as data classes

It's already on the classpath, example : https://github.com/aws/aws-toolkit-jetbrains/blob/63b04161077b7fd9f9405075b93f45607ee95445/jetbrains-core/src/software/aws/toolkits/jetbrains/services/RoleValidation.kt#L11

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/LinterRule.kt
```diff
@@ -0,0 +1,13 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.services.cloudformation.annotations
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
kiiadi, line 8: We should configure the `ObjectMapper` to ignore unknown properties - rather than doing it for every actual type.

jetbrains-core/it/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotatorTest.kt
miparnisari: ` FileDocumentManager.getInstance()`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
PatMyron: While `AWS::Serverless-2016-10-31` and `AWSTemplateFormatVersion` definitely indicate CloudFormation templates, I suspect most CloudFormation templates don't contain either

It'd be nice if other CloudFormation templates could be recognized using something like [`Resources` and `Type`](https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/blob/edd51fb86441a1c48d294c4b664189c6f4d11399/server/src/server.ts#L132-L133)

PatMyron: Agree with keeping CloudFormation template recognition in one place though

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotator.kt
miparnisari: They are by no means required but they did help me a lot to better structure the code

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
miparnisari: > It'd be nice if other CloudFormation templates could be recognized using something like Resources and Type

I included this in the latest commit

jetbrains-core/it/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotatorTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test
+import software.aws.toolkits.jetbrains.utils.rules.JavaCodeInsightTestFixtureRule
+
+class CloudFormationLintAnnotatorTest {
+
+    @Rule
+    @JvmField
+    val projectRule = JavaCodeInsightTestFixtureRule()
+
+    var cloudFormationLintAnnotator: CloudFormationLintAnnotator = CloudFormationLintAnnotator()
```
kiiadi, line 23: this can be a `val`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
kiiadi: see my commend above - I'm not sure this resolves it.

jetbrains-core/it/software/aws/toolkits/jetbrains/services/cloudformation/annotations/CloudFormationLintAnnotatorTest.kt
kiiadi: per @abrooksv 's comment https://github.com/aws/aws-toolkit-jetbrains/pull/1715#discussion_r413301623 - we shouldn't need to mock the `FileDocumentManager` - that puts us on a treadmill.

miparnisari: Ok I fixed this, though I'm not sure why code coverage doesn't show this as covered :(

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
miparnisari: @kiiadi, not sure how to proceed. The code I wrote handles CFN templates but not SAM templates. If @abrooksv incorporates pat's comment into this branch, I can definitely rebase on top of that. 

So the logic should look something like

```java
if (contains "Transform: AWS::Serverless-2016-10-31")
   return true
if (contains "AWSTemplateFormatVersion: "2010-09-09"")
   return true
if (contains "Resources" block AND each Resource has a "Type" key) 
   return true

return false
```

abrooksv: The last case is too unstable to use. As a user is trying to add a new resource the file will "blink" between being a CFN file and not being a CFN file. This has repercussions of things like auto-complete.

We will probably allow people to mark arbitrary files as CFN files and cache that as a VirtualFile attribute, but I have not made that change yet.

PatMyron: > The code I wrote handles CFN templates but not SAM templates

wouldn't hurt to add `Transform: AWS::Serverless-2016-10-31` since that definitely indicates a template, but [SAM templates](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) should already be detected by the existing logic

> ```java
>    if (psiFile contains "Resources" block AND each Resource has a "Type" key) 
> ```
> The last case is too unstable to use. As a user is trying to add a new resource the file will "blink" between being a CFN file and not being a CFN file. This has repercussions of things like auto-complete



don't think this necessarily needs to be checked for every resource. [The Visual Studio Code extension only checks for one instance](https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/blob/edd51fb86441a1c48d294c4b664189c6f4d11399/server/src/server.ts#L133) and seems to be sufficient:

```java
   if (psiFile contains "Resources" block AND a Resource has a "Type" key) 
```

miparnisari:
@kiiadi we need to solve this discussion https://github.com/aws/aws-toolkit-jetbrains/pull/1715#discussion_r421074217

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
PatMyron: also might want to filter out https://www.serverless.com/ files like the VSCode extension does if we include the `Resources` and `Type` logic:
https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/pull/25#issuecomment-642416464

.travis.yml
PatMyron: should be able to get Python and `pip` installed on Windows and install through that

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/annotations/InitialAnnotationResults.kt
miparnisari: > We will probably allow people to mark arbitrary files as CFN files and cache that as a VirtualFile attribute, but I have not made that change yet.

This doesn't sound very user-friendly.

Would it make sense to only look out for `"Transform: AWS::Serverless-2016-10-31"` and `"AWSTemplateFormatVersion: \"2010-09-09\""`? Plus introducing a new warning rule in `cfn-lint` that using these are good practices? (I don't understand why these were not made required tags from the very beginning)

PatMyron: > Plus introducing a new warning rule in `cfn-lint` that using these are good practices? (I don't understand why these were not made required tags from the very beginning)

The [Linter does recommend `Transform: AWS::Serverless-2016-10-31`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/ServerlessTransform.py) for the [Serverless transform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) if the template contains [`AWS::Serverless` resource types](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-resources-and-properties.html)

`Transform: AWS::Serverless-2016-10-31` should definitely not be recommended for every CloudFormation template. It requires an additional capability, `CAPABILITY_AUTO_EXPAND`, to be explicitly granted for stack operations to continue succeeding, [Stack sets](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/what-is-cfnstacksets.html) [does not support transforms](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/207), and many other features like [`aws cloudformation validate-template` and `aws cloudformation get-template-summary` do not understand templates with transforms](https://i.amazon.com/CFN-DE-1693?selectedConversation=f0a390bc-664c-445a-b72f-752023e3738e)

miparnisari: > > Plus introducing a new warning rule in `cfn-lint` that using these are good practices? (I don't understand why these were not made required tags from the very beginning)
> 
> The [Linter does recommend `Transform: AWS::Serverless-2016-10-31`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/ServerlessTransform.py) for the [Serverless transform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) if the template contains [`AWS::Serverless` resource types](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-resources-and-properties.html)

Ok, what if it also recommended adding `AWSTemplateFormatVersion: "2010-09-09"` to the templates so that in this plugin we only look out for that?

PatMyron: > in this plugin we only look out for that?

We should still look out for more signals because [`AWSTemplateFormatVersion`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/format-version-structure.html) can never be required by CloudFormation itself at this point. No reason to discard the other signals even if we were to start encouraging one of them

abrooksv: The API for detecting file types by content is limited to the first 1024 bytes, and you don't have access to a pre-parsed file. This is what is needed if you want to support `foo.template` files which the existing plugin does.

For when you know the file extension, you do get a PSI file for it, but you still need this action to be as quick as possible due to it will be called for every single file of that type and multiple times. You need to balance speed for accuracy

miparnisari: @kiiadi @abrooksv what's the status of https://github.com/aws/aws-toolkit-jetbrains/tree/feature/cloudformation ?

kiiadi:
@miparnisari this branch is getting pretty out of date - would you mine rebasing off the [`feature/cloudformation`](https://github.com/aws/aws-toolkit-jetbrains/tree/feature/cloudformation) branch? This is our oldest PR - I'd like to get it closed out.

miparnisari:
As per conversation with Kyle, Austin & others, i'm putting this PR on hold.

PR #1718, opened 2020-04-20
Change credentials reloaded message wording so it does not get cut off

hunterwerlla:
- Change credentials reloaded message wording so it does not get cut off
- Update the message to handle plurals properly
## Screenshots
Screen Shot 2020-04-20 at 12 10 57 PM
Screen Shot 2020-04-20 at 12 12 43 PM

## License


abrooksv:
What about the error case with invalid profile?

hunterwerlla:
> What about the error case with invalid profile?

Honestly forgot that was the impetus for this, message updated and screenshot added

PR #1728, opened 2020-04-23
Add the option to delete log groups

hunterwerlla:
- Deletes the log group
- Closes open windows

## Screenshots
Screen Shot 2020-04-23 at 3 09 37 PM
Screen Shot 2020-04-23 at 3 10 57 PM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/CloudWatchLogWindow.kt
```diff
@@ -89,8 +89,16 @@ class CloudWatchLogWindow(private val project: Project) : CoroutineScope by Appl
         }
     }
 
+    fun closeLogGroup(logGroup: String) = launch {
+        toolWindow.findPrefix(logGroup).forEach {
```
abrooksv, line 93: We should find a better route, this will close FooBar if you delete Foo, or mutli-account with FooBar open

hunterwerlla: I know this happens but I didn't want to fix it because it's a rare case, will do a quick fix. We should probably write a toolwindowmangerV2 that makes stuff like this easier as we have added a lot more expectations on top of this vs just opening a log or cloudformation stack.

abrooksv: Lots of V2 stuff we need to write....

PR #1737, opened 2020-04-30
Changing default IDE profile to 2020.1

kiiadi:


hunterwerlla:
I disagree with this change because it means we will have less trying out on our min version, thoughts?

kiiadi:
Either-way we have CI jobs that cover all of them.

I think the biggest risk with this change is that we start using APIs that don't exist in older versions.

Maybe you're right.

abrooksv:
> Either-way we have CI jobs that cover all of them.
> 
> I think the biggest risk with this change is that we start using APIs that don't exist in older versions.
> 
> Maybe you're right.

I am less concerned about that given the IDE tells you now:

```
'com.intellij.ui.scale.JBUIScale' is available only since 192.5118 but the module is targeted for 192.0 - 193.SNAPSHOT. It may lead to compatibility problems with IDEs prior to 192.5118. Note that this type might have had a different fully qualified name in the previous IDEs. 
```

hunterwerlla:
> > Either-way we have CI jobs that cover all of them.
> > I think the biggest risk with this change is that we start using APIs that don't exist in older versions.
> > Maybe you're right.
> 
> I am less concerned about that given the IDE tells you now:
> 
> ```
> 'com.intellij.ui.scale.JBUIScale' is available only since 192.5118 but the module is targeted for 192.0 - 193.SNAPSHOT. It may lead to compatibility problems with IDEs prior to 192.5118. Note that this type might have had a different fully qualified name in the previous IDEs. 
> ```
> 
> As long as we don't change the plugin.xml from min version, we will keep getting these warnings

If these warnings don't fail build though I think something will slip through. We already ignore them because our min is not our true min (which is the version released).

abrooksv:
> > > Either-way we have CI jobs that cover all of them.
> > > I think the biggest risk with this change is that we start using APIs that don't exist in older versions.
> > > Maybe you're right.
> > 
> > 
> > I am less concerned about that given the IDE tells you now:
> > ```
> > 'com.intellij.ui.scale.JBUIScale' is available only since 192.5118 but the module is targeted for 192.0 - 193.SNAPSHOT. It may lead to compatibility problems with IDEs prior to 192.5118. Note that this type might have had a different fully qualified name in the previous IDEs. 
> > ```
> > 
> > 
> > As long as we don't change the plugin.xml from min version, we will keep getting these warnings
> 
> If these warnings don't fail build though I think something will slip through. We already ignore them because our min is not our true min (which is the version released).

That is true, but it would fail in CI. I do not have a strong opinion either way. I currently have mine set to 2020.1 locally, but I also switch it out a lot when testing so no strong preference either way.

hunterwerlla:
> > > > Either-way we have CI jobs that cover all of them.
> > > > I think the biggest risk with this change is that we start using APIs that don't exist in older versions.
> > > > Maybe you're right.
> > > 
> > > 
> > > I am less concerned about that given the IDE tells you now:
> > > ```
> > > 'com.intellij.ui.scale.JBUIScale' is available only since 192.5118 but the module is targeted for 192.0 - 193.SNAPSHOT. It may lead to compatibility problems with IDEs prior to 192.5118. Note that this type might have had a different fully qualified name in the previous IDEs. 
> > > ```
> > > 
> > > 
> > > As long as we don't change the plugin.xml from min version, we will keep getting these warnings
> > 
> > 
> > If these warnings don't fail build though I think something will slip through. We already ignore them because our min is not our true min (which is the version released).
> 
> That is true, but it would fail in CI. I do not have a strong opinion either way. I currently have mine set to 2020.1 locally, but I also switch it out a lot when testing so no strong preference either way.

I guess that's true it would fail CI because at that point it pulls the actual version. This is probably fine to put in then

abrooksv:
Actually I change my mind. Right now `/run[Ui|Integration]tests` use the default profile. Until we make that test matrix aware, I would keep it at 192 so the ui tests do something 😄

kiiadi:
fair points, I guess I'll just go back to changing it locally - I just had having modified files in my working dir - but that's my cross to bear.

PR #1753, opened 2020-05-13
Add refresh to S3 browser and rework UI

hunterwerlla:
- Add a refresh action to the S3 browser
- Rework the S3 browser UI to remove redundant information in the big box at the top. Name is already at the top and can be copied from the node, ARN can be copied from the node, and create date is not really important during development.
- Add a toolbar that compliments right click actions
- Feedback on toolbar at bottom/side (it's based on platform) vs top?

### Before
Screen Shot 2020-05-12 at 2 36 07 PM

### After
Screen Shot 2020-05-15 at 9 00 00 AM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3ViewerPanel.java
```diff
@@ -3,88 +3,91 @@
...
+        DefaultActionGroup group = makeActionGroup(table);
+        group.addAction(new AnAction(message("explorer.refresh.title"), null, AllIcons.Actions.Refresh) {
+            @Override
+            public boolean isDumbAware() {
+                return true;
+            }
+
+            @Override
+            public void actionPerformed(@NotNull AnActionEvent e) {
+                table.setModel(createTreeTableModel(columns, new S3TreeDirectoryNode(virtualBucket, null, "")));
```
abrooksv, line 66: Why do we need a new model? This is using the same AsyncTreeModel as the explorer, why can't we invalidate it the same way?

hunterwerlla: Yeah this was due to it not working, looked at it again today and it works fine.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeNode.kt
```diff
@@ -55,7 +55,7 @@ class S3TreeDirectoryNode(private val bucket: S3VirtualBucket, parent: S3TreeDir
 
         val continuation = listOfNotNull(response.nextContinuationToken()?.let {
             // Spaces are intentional
```
abrooksv, line 57: Out of date comment

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/S3ObjectAction.kt
```diff
@@ -23,7 +23,17 @@ abstract class S3ObjectAction(protected val treeTable: S3TreeTable, title: Strin
...
     override fun actionPerformed(e: AnActionEvent) = performAction(selected().filter { it !is S3TreeContinuationNode })
 
-    private fun selected(): List<S3TreeNode> = treeTable.getSelectedNodes().takeIf { it.isNotEmpty() } ?: listOf(treeTable.getRootNode())
+    private fun selected(): List<S3TreeNode> {
+        val selectedNodes = treeTable.getSelectedNodes()
+        if (selectedNodes.isNotEmpty()) {
+            return selectedNodes
+        }
+        val rootNode = treeTable.getRootNode()
+        if (rootNode != null) {
```
abrooksv, line 32: `listOfNotNull(rootNode)`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
```diff
@@ -207,7 +207,8 @@ class S3TreeTable(
         return (path.lastPathComponent as DefaultMutableTreeNode).userObject as? S3TreeNode
     }
 
-    fun getRootNode(): S3TreeDirectoryNode = (tableModel.root as DefaultMutableTreeNode).userObject as S3TreeDirectoryNode
+    // This is nullable because tableModel.root is nullable. It is null sometimes, like when we invalidate the tree on a refresh
+    fun getRootNode(): S3TreeDirectoryNode? = (tableModel.root as? DefaultMutableTreeNode)?.userObject as? S3TreeDirectoryNode
```
abrooksv, line 211: We never actually allow root node to be null / replaced, Why not just keep a reference to it in this class instead of going through the model?

abrooksv:
Let's allow users to refresh subfolders so they don't have to do a full refresh/expand.

Also should we re-expand on refresh?

hunterwerlla:
> Let's allow users to refresh subfolders so they don't have to do a full refresh/expand.
> 
> Also should we re-expand on refresh?

I like the first idea. re-expanding on refresh is doable, but would not really work as expected since it won't load all of the currently loaded pages (because we are trying to paginate something that isn't meant to be paginated). It would probably still be a better experience so I'll try it out and see how it looks.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeNode.kt
hunterwerlla: Removed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/editor/S3TreeTable.kt
hunterwerlla: That makes a lot of sense, changed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/S3ObjectAction.kt
hunterwerlla: Now that rootNode is not nullable again this change can be reverted

hunterwerlla:
Added the two types of refresh, now just need to add tests

abrooksv:
> > Let's allow users to refresh subfolders so they don't have to do a full refresh/expand.
> > Also should we re-expand on refresh?
> 
> I like the first idea. re-expanding on refresh is doable, but would not really work as expected since it won't load all of the currently loaded pages (because we are trying to paginate something that isn't meant to be paginated). It would probably still be a better experience so I'll try it out and see how it looks.

Good call out on pagination.... We can re-look if people ask enough, something like keep track of number of pages it took to get that far.... I am fine punting until we get people asking to warrant the complexity

hunterwerlla:
These tests are actually horrible to write and getting a `java.awt.HeadlessException`, I see now why most of these don't have tests around them. This is a place that we should add UI tests when we get the new framework merged in.

kiiadi:
There was UI tests around the S3 browser before ...

hunterwerlla:
> There was UI tests around the S3 browser before ...

There still are, they just don't work with 2020.1

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -449,7 +449,6 @@ notice.message.jetbrains.minimum.version=Support for version {0} of {1} is being
 notice.suppress=Don't show this again
 notice.title.jetbrains.minimum.version=AWS Toolkit Deprecation Notice
 resource_selector.create=Create
-s3.bucket.arn.label=Bucket ARN:
 s3.bucket.creation_date.label=Creation Date:
```
abrooksv, line 452: This can be deleted

hunterwerlla: Bucket name is still used but creation date can be deleted, updated

PR #1761, opened 2020-05-15
Update deprecation notice to warn users that the next release requires 2019.3

hunterwerlla:
## 2019.2
Screen Shot 2020-05-15 at 4 14 09 PM

## License


abrooksv:
> I think this is fine - tho @abrooksv had an alternate solution that hung off the explorer and was more in your face.

Let's discuss this this afternoon to see if we want to change it, or ship what we got

PR #1762, opened 2020-05-16
Adding credential drop-downs to explorer toolbar

kiiadi:
More obvious surfacing of AWS credentials / region selection

The next step will be surfacing the connection state more obviously - you'll notice there's a delay between when credentials are selected and when the explorer loads - this is while the credential is being validated (e.g. can we call STS `get-caller-identity` successfully).


ckant:
Looks good! Would it be possible to remove the `profile:` prefix? Perhaps group by `Profile` rather than `Credentials` in the dropdown? 

If it's really needed, perhaps we could make it a suffix instead. I think it's a little noisy having to scan past the prefix to read the current profile name. Especially noisy if the profile name itself is short.

kiiadi:
@ckant quite like the idea of grouping by type to reduce the noise - not sure how that should work in 'recent' credentials though - where it might be pertinent to know the type of credential (e.g. no guarantee that id's will be consistent across types).

right now it's not really even an issue since *everything* is a `Profile` so we could probably safely strip it out and defer the decision until we support other credential types (e.g. ecs / ec2).

kiiadi:
@ckant reckon something like this?
Screen Shot 2020-05-20 at 1 38 39 PM 

prly not as part of this PR - but could be a separate UX.

kiiadi:
OK - new experience

ckant:
> @ckant reckon something like this?
> Screen Shot 2020-05-20 at 1 38 39 PM
> 
> prly not as part of this PR - but could be a separate UX.

That looks really good. I think it could coincide with `demo (Profile)` showing in the selected dropdown. It's consistent with the region which is `Oregon (us-west-2)`. More important information coming first. Additional details in parenthesis.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/nodes/AwsExplorerErrorNode.kt
```diff
@@ -23,11 +25,13 @@ class AwsExplorerErrorNode(project: Project, exception: Throwable) :
...
-                val awsErrorDetails = exception.awsErrorDetails()
-                "${awsErrorDetails.serviceName()}: ${awsErrorDetails.errorCode()}"
-            } else {
-                message("explorer.error_loading_resources_default_details")
+            val errorDetails = when {
+                exception is AwsServiceException -> {
+                    val awsErrorDetails = exception.awsErrorDetails()
+                    "${awsErrorDetails.serviceName()}: ${awsErrorDetails.errorCode()}"
+                }
+                !accountSettings.isValidConnectionSettings() -> message("explorer.error_loading_resources_not_connected")
```
abrooksv, line 33: Is this still possible to hit?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/NotificationUtils.kt
```diff
@@ -80,7 +81,7 @@ fun notifyNoActiveCredentialsError(
         title = title,
         content = content,
         project = project,
-        action = SettingsSelectorAction(showRegions = false)
+        action = SettingsSelectorAction(ChangeAccountSettingsMode.CREDENTIALS)
```
abrooksv, line 84: Can't comment on the function...

Where is this used? Will this lead to confusion if its conflated with no region?

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/credentials/SettingsSelectorComboBoxActionTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Test
+import software.aws.toolkits.core.credentials.aToolkitCredentialsIdentifier
+import software.aws.toolkits.core.region.anAwsRegion
+
+class SettingsSelectorComboBoxActionTest {
+    @Rule
+    @JvmField
+    val projectRule = ProjectRule()
+
+    private val DUMMY_REGION = anAwsRegion()
```
abrooksv, line 19: The fuzzy-ing is nice

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/nodes/AwsExplorerErrorNode.kt
kiiadi: good point - shoudl remove.

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/credentials/SettingsSelectorComboBoxActionTest.kt
kiiadi: yeah i think this is a pattern we should use more.

core/tst/software/aws/toolkits/core/region/AwsRegionTest.kt
```diff
@@ -40,3 +42,12 @@ class AwsRegionTest(private val region: AwsRegion, private val expectedCategory:
         assertThat(region.category, equalTo(expectedCategory))
     }
 }
+
+fun anAwsRegion(id: String = aRegionId(), name: String = aString(), partitionId: String = aString()) = AwsRegion(id, name, partitionId)
```
hunterwerlla, line 46: Shouldn't this also go in credential mocks? (or its own mocks file?)

core/tst/software/aws/toolkits/core/utils/test/TestUtils.kt
```diff
@@ -5,7 +5,11 @@ package software.aws.toolkits.core.utils.test
 
 import java.time.Duration
 import java.time.Instant
+import java.util.UUID
 import java.util.concurrent.atomic.AtomicInteger
+import kotlin.random.Random
+
+fun aString(length: Int = Random.nextInt(5, 30)): String = UUID.randomUUID().toString().substring(length)
```
hunterwerlla, line 12: I like these randomized tests

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
```diff
@@ -64,8 +68,15 @@ class ExplorerToolWindow(private val project: Project) : SimpleToolWindowPanel(t
     }
 
     init {
+        val group = DefaultActionGroup(
+            SettingsSelectorComboBoxAction(project, ChangeAccountSettingsMode.CREDENTIALS),
+            SettingsSelectorComboBoxAction(project, ChangeAccountSettingsMode.REGIONS)
```
hunterwerlla, line 73: I'm curious how this works, here we use credentials and regions but we show both below? It seems like the enum is doing too many things or I'm just confused on how it should be used?

core/tst/software/aws/toolkits/core/region/AwsRegionTest.kt
kiiadi: They're not really mock's they're just test data generators

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
kiiadi: You're *probably* right the enum might be doing too many things, but the logic is all encapsulated together - found myself doing a bunch of 

kt
```
when(mode) {
  BOTH -> logic
  REGIONS -> logic
}
```

stuff- and figured we could collapse it down.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/NotificationUtils.kt
kiiadi: this is specifically `nonotifyNoActiveCredentialsError`

kiiadi: but looking at where it's used, it could be either - will fix###

kiiadi: actually - i'm going to do this separately - because it relies on logic that is already in the next PR

.github/workflows/mac.yml
```diff
@@ -24,4 +24,4 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Build with Gradle
-      run: ./gradlew check
+      run: ./gradlew --no-daemon check coverageReport --info --debug --full-stacktrace --console plain
```
abrooksv, line 27: Revert before final merge

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ChangeConnectionSettingsMenu.kt
```diff
@@ -44,26 +47,43 @@ class ChangeAccountSettingsActionGroup(private val project: Project, private val
...
         }
 
         CachedValueProvider.Result.create(actions.toTypedArray(), accountSettingsManager)
     }
 }
 
+enum class ChangeAccountSettingsMode(
+    internal val showRegions: Boolean,
+    internal val showCredentials: Boolean,
+    internal val nonSelected: String,
```
abrooksv, line 75: Can we fold this and selectedDisplayFormat together, it seems are are mutually exclusive and that its usage implies `selectedDisplayFormat ?: nonSelected`

PR #1767, opened 2020-05-18
Update run configs

abrooksv:
* Add idea.log for ultimate run configs
* Disable Gradle debug by default so it can re-use Gradle daemons

## License


.run/Run IDE - Ultimate [2019.2].run.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectRunConfigurationManager">
```
Comment (rli) [inline .run/Run IDE - Ultimate [2019.2].run.xml:1]: why change this file if it's about to be deleted?

PR #1772, opened 2020-05-19
Do not default to positive in feedback dialog

abrooksv:
## License

### DIFF
--- a/jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/feedback/FeedbackTest.kt
+++ b/jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/feedback/FeedbackTest.kt
@@ -27,13 +27,13 @@ class FeedbackTest {
     class NonParameterizedTests {
         @Test
         fun panelInitiallyNegative() {
-            val panel = SubmitFeedbackPanel(initiallyPositive = false)
+            val panel = SubmitFeedbackPanel(Sentiment.NEGATIVE)
             assertThat(panel.sentiment).isEqualTo(Sentiment.NEGATIVE)
         }
 
         @Test
         fun panelInitiallyPositive() {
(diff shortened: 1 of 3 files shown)


jetbrains-core/tst/software/aws/toolkits/jetbrains/ui/feedback/FeedbackTest.kt
```diff
@@ -43,9 +43,6 @@ class FeedbackTest {
                 val dialog = FeedbackDialog(projectRule.project)
                 val panel = dialog.getViewForTesting()
 
-                assertThat(panel.sentiment).isEqualTo(Sentiment.POSITIVE)
-
-                panel.clearSentimentSelection()
                 assertThat(panel.sentiment).isEqualTo(null)
```
hunterwerlla, line 46: Does posting feedback with it null work, I don't think it would with how the auto-generated code is:
```
        public final Builder sentiment(Sentiment sentiment) {
            this.sentiment(sentiment == null ? null : sentiment.toString());
```

abrooksv: No, there is a validation that prevents it from being sent, we just never hit it before:
https://github.com/aws/aws-toolkit-jetbrains/pull/1772/files#diff-4d65ad53aa6a658ac3d9c1983ec36d43R74

hunterwerlla: Ah OK, didn't expect that to exist since we didn't have that scenario before, looks good

abrooksv: You. commented on the unit test that validates it 😄

PR #1778, opened 2020-05-20
Revert release that didn't really happen and change build.gradle to publish to multiple channels

hunterwerlla:
## License

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -243,7 +243,7 @@ prepareSandbox {
 
 publishPlugin {
     token publishToken
-    channels publishChannel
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -243,7 +243,7 @@ prepareSandbox {
 
 publishPlugin {
     token publishToken
-    channels publishChannel
+    channels publishChannel ? publishChannel.split(',') : []
```
abrooksv, line 246: Should we trim?

hunterwerlla: Sure, added

PR #1786, opened 2020-05-22
Add S3 Bucket Icon

hunterwerlla:
Screen Shot 2020-05-22 at 3 33 51 PM

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1788, opened 2020-05-22
Make all codecov checks informational

hunterwerlla:
## License

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -25,7 +25,7 @@ coverage:
       default:
         threshold: 1
         only_pulls: true
-        informational: false
+        informational: true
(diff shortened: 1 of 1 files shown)


kiiadi:
Why? I like the break on the patch. We should be adding tests when we add code.

hunterwerlla:
> Why? I like the break on the patch. We should be adding tests when we add code.

I was mostly annoyed that PRs like this https://github.com/aws/aws-toolkit-jetbrains/pull/1786 "fail" when a test doesn't really test much. Agree it should probably exist to be annoying for the other 90% of PRs, closing this one

PR #1796, opened 2020-05-27
Surfacing various connection states

kiiadi:
Improving visibility to current AWS connection state, surfacing it with consistent messaging.
Untitled


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/AwsSettingsPanel.kt
```diff
@@ -57,18 +48,17 @@ private class AwsSettingsPanel(private val project: Project) : StatusBarWidget,
 
     override fun install(statusBar: StatusBar) {
         this.statusBar = statusBar
-        project.messageBus.connect().subscribe(ProjectAccountSettingsManager.CONNECTION_SETTINGS_CHANGED, this)
+        project.messageBus.connect().subscribe(ProjectAccountSettingsManager.CONNECTION_SETTINGS_STATE_CHANGED, this)
```
abrooksv, line 51: This should actually be `connect(this)`

kiiadi: fixed

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -621,8 +621,14 @@ settings.credentials=Credentials
 settings.credentials.none_selected=No credentials selected
 settings.credentials.profile_sub_menu=All Credentials
 settings.credentials.recent=Recent Credentials
+settings.none_selected=No region or credentials selected
 settings.partitions=Partitions
 settings.regions.none_selected=No region selected
 settings.regions.recent=Recent Regions
 settings.regions.region_sub_menu=All Regions
+settings.states.initializing=Toolkit initializing...
+settings.states.invalid=Unable to connect to AWS using selected credentials ({0}) / region ({1}): {2}
```
abrooksv, line 630: Can we put a new line after the colon

abrooksv:
So here is a question:

Do we consider the validating / error states to be considered empty states of the explorer?

If so, should we follow their guidance? https://jetbrains.github.io/ui/principles/empty_state/

The state machine changes I like, and have no concerns with.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
kiiadi: done

PR #1804, opened 2020-05-29
Implement ability to refresh connection

kiiadi:
Connection state (and thus events that drive it) can be refreshed in several ways:
- Refresh button in the explorer
- Retry action on the "invalid connection" notification
- If the selected profile is refreshed (and it was previously marked invalid) a refresh occurs automatically (e.g. if the profile file is corrected).


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -64,6 +64,12 @@ abstract class ProjectAccountSettingsManager(private val project: Project) : Sim
                         changeConnectionSettings(null, selectedPartition, selectedRegion)
                     }
                 }
+
+                override fun providerModified(identifier: ToolkitCredentialsIdentifier) {
+                    if (selectedCredentialIdentifier == identifier && connectionState is ConnectionState.InvalidConnection) {
```
abrooksv, line 69: Should this always validate? If they modified the provider at all we should probably check again. Right now this fixes going from bad->good, but not good->bad

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/RefreshConnectionAction.kt
```diff
@@ -0,0 +1,28 @@
...
+        when (ProjectAccountSettingsManager.getInstance(project).connectionState) {
+            is ConnectionState.InitializingToolkit, is ConnectionState.ValidatingConnection -> e.presentation.isEnabled = false
+            else -> e.presentation.isEnabled = true
+        }
+    }
+
+    override fun actionPerformed(e: AnActionEvent) {
+        val project = e.project ?: return
+        ProjectAccountSettingsManager.getInstance(project).refreshConnectionState()
+        AwsResourceCache.getInstance(project).clear()
```
abrooksv, line 26: I would clear the cache first since technically this will go and notify the tree to refresh. Ya, likely to never happen but let's make it so it can't

kiiadi: yes, you're right.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
kiiadi: I did think about it - my thought was do we want to disrupt an apparently stable connection? We would temporarily remove the explorer while we do the validate - and if they're off the network or something it'll prevent them from seeing the resources they could see before.

I can be convinced though - maybe it would be less confusing - and probably also makes sense in an MFA world.

abrooksv: I (believe) we validate the profile actually has changes so editing Profile Foo will not trigger Bar to be modified.

If that is true, it would imply to me that they are making an active decision to mess with it.

(We may have a gap here though in source_profile chains, I am not sure we notify the consumers of the source_profile) has been modded if the source is altered)

kiiadi: Will test the source profile

PR #1805, opened 2020-05-29
Adding region sub-menu for partitions

kiiadi:


kiiadi:
Alternative menu that shows "Other Partitions" and only allows selection of the non-selected partitions.

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/credentials/ChangeAccountSettingsActionGroupTest.kt
```diff
@@ -61,9 +59,8 @@ class ChangeAccountSettingsActionGroupTest {
             .filterIsInstance<ChangeRegionActionGroup>().first().getChildren(null)
             .filterIsInstance<ChangePartitionActionGroup>().first().getChildren(null)
 
-        val selectedAction = partitionActions.filterIsInstance<ToggleAction>().first()
-        assertNotNull(selectedAction)
-        assertThat(selectedAction.templateText).isEqualTo(selectedRegion.partitionId)
+        val selectedAction = partitionActions.firstOrNull { it.templateText == selectedRegion.partitionId }
```
abrooksv, line 62: Probably reads better to say `.none {  }` and assert true

kiiadi: good call.

PR #1808, opened 2020-06-01
Clean up MFA, introduce a connection requires user interaction state

abrooksv:
* Upon starting, if the creds require user interaction, enter terminal state that user must trigger to prevent prompts on startup

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -290,13 +305,6 @@ fun Project.activeRegion(): AwsRegion = ProjectAccountSettingsManager.getInstanc
  */
 fun Project.activeCredentialProvider(): ToolkitCredentialsProvider = ProjectAccountSettingsManager.getInstance(this).activeCredentialProvider
 
-/**
- * The underlying AWS account for current active credential provider of the project. Return null if credential provider is not set.
- * Calls of this member should be in non-UI thread since it makes network call using an STS client for retrieving the
- * underlying AWS account.
- */
-fun Project.activeAwsAccount(): String? = tryOrNull { AwsResourceCache.getInstance(this).getResourceNow(StsResources.ACCOUNT) }
```
kiiadi, line 298: Could probably just kill `activeAwsAccountIfKnown` altogether - it's only used in the telemetry service.

abrooksv: Ya, I put it here due to activeAwsAccount is here. We should move it into telemetry service itself.

abrooksv: I'll do that on a PR to master

abrooksv: #1810

abrooksv:
Split how we deep of an inspection we do for cloud debug so that it does not trigger MFA during shallow validation.

Remote Lambda does not appear to trigger any lockup because we do not validate the resource actually exists every time

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+    yield(currentProfile)
+
+    while (currentProfile.propertyExists(ProfileProperty.ROLE_ARN)) {
+        val currentProfileName = currentProfile.name()
+        if (!profileChain.add(currentProfileName)) {
+            val chain = profileChain.joinToString("->", postfix = "->$currentProfileName")
+            throw IllegalArgumentException(message("credentials.profile.circular_profiles", topLevelProfile.name(), chain))
+        }
+
+        val sourceProfile = currentProfile.requiredProperty(ProfileProperty.SOURCE_PROFILE)
```
kiiadi, line 24: So we don't support `credential_source` chains? 

From https://docs.aws.amazon.com/cli/latest/topic/config-vars.html
> credential_source - The credential provider to use to get credentials for the initial assume-role call. This 
  parameter cannot be provided alongside source_profile. Valid values are:
  Environment to pull source credentials from environment variables.
  Ec2InstanceMetadata to use the EC2 instance role as source credentials.
  EcsContainer to use the ECS container credentials as the source credentials.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/execution/EcsCloudDebugRunConfiguration.kt
```diff
@@ -80,7 +80,7 @@ class EcsCloudDebugRunConfiguration(project: Project, private val configFactory:
     }
 
     override fun checkSettingsBeforeRun() {
-        val immutableSettings = resolveEcsServiceCloudDebuggingRunSettings()
+        val immutableSettings = resolveEcsServiceCloudDebuggingRunSettings(true)
```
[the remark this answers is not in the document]
abrooksv: Ya, I try to do that sometimes, but I forget

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -33,6 +33,11 @@ abstract class ToolkitCredentialsIdentifier {
      */
     open val defaultRegionId: String? get() = null
 
+    /**
+     * Return true if the credential may require prompting users for more information
+     */
+    open val requiresUserAction: Boolean = false
```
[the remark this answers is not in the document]
abrooksv: Would we want to wrap and using Kotlin interface delegation then?

I'm not really a fan of ProfileCredentialProvider, ProfileCredentialProviderMFA, and ProfileCredentialProviderSSO

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/MfaSupport.kt
```diff
@@ -0,0 +1,21 @@
...
+import software.aws.toolkits.jetbrains.utils.getCoroutineUiContext
+import software.aws.toolkits.resources.message
+
+fun promptForMfaToken(name: String, mfaSerial: String): String = runBlocking {
+    withContext(getCoroutineUiContext(ModalityState.any())) {
+        Messages.showInputDialog(
+            message("credentials.profile.mfa.message", mfaSerial),
+            message("credentials.profile.mfa.title", name),
+            null
+        ) ?: throw IllegalStateException("MFA challenge is required")
```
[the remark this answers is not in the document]
abrooksv: Shows up as the exception from the SDK call

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -143,15 +144,24 @@ abstract class ProjectAccountSettingsManager(private val project: Project) : Sim
...
+        val isInitial = connectionState is ConnectionState.InitializingToolkit
         connectionState = ConnectionState.ValidatingConnection
 
         // Clear existing provider
         selectedCredentialsProvider = null
 
         fieldUpdateBlock()
 
-        validationJob = GlobalScope.launch(Dispatchers.IO) {
+        if (isInitial) {
```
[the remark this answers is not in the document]
abrooksv: We don't have selectedCredentialIdentifier till after fieldUpldateBlock(), so we either move into Validating state after that or do this way, right?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -143,15 +144,24 @@ abstract class ProjectAccountSettingsManager(private val project: Project) : Sim
         incModificationCount()
```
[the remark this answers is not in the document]
abrooksv: Ya, it should probably be synchronized tbh.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ProjectAccountSettingsManager.kt
```diff
@@ -274,6 +285,10 @@ sealed class ConnectionState(val displayMessage: String, val isTerminal: Boolean
     class InvalidConnection(val cause: Exception) : ConnectionState(message("settings.states.invalid", cause.localizedMessage), isTerminal = true) {
         override val shortMessage = message("settings.states.invalid.short")
     }
+
+    object RequiresUserAction : ConnectionState(message("settings.states.requiresUser"), isTerminal = true) {
```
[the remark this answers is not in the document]
abrooksv: True

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/execution/EcsCloudDebugRunConfiguration.kt
```diff
@@ -64,7 +64,7 @@ class EcsCloudDebugRunConfiguration(project: Project, private val configFactory:
 
     override fun getState(executor: Executor, environment: ExecutionEnvironment): CloudDebugRunState? =
         try {
-            val cloudDebugRunSettings = resolveEcsServiceCloudDebuggingRunSettings()
+            val cloudDebugRunSettings = resolveEcsServiceCloudDebuggingRunSettings(true)
```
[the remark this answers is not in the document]
abrooksv: Not sure I follow.

Ideally it shouldn't need to do any validation, but we conflated validation with building the immutable state. So I guess it could be shallow since it should be guarded by checkConfigBeforeRun()

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileUtils.kt
abrooksv: Nope, we don't today since we don't support any of the other providers.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/execution/EcsCloudDebugRunConfiguration.kt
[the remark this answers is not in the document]
kiiadi: as long as that doesn't affect the cascading drop-down loads.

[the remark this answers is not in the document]
abrooksv: Switched it to false, but also that loads data in the background so yay

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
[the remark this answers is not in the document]
kiiadi: Well they're *all* "ProfileCredentialProviders" - but they *do* have different ways of actually resolving the credentials. There's also "Static", "AssumeRole" and "CredentialProcess" types if you think about it. Wouldn't it simplify our code ultimately if the different types were broken out? We know when we create the `ToolkitCredentialsIdentifier` in the factory's `setUp` method what type of profile it is.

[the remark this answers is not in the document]
abrooksv: Created a version of this, interactive block is kept out because it may be deeper than the identifier can get to it like in a multi-link credential chain

Root creds -> Assume role into dev account using mfa -> assume role into beta account.

Identifier itself sees the assume role into beta, so won't have the data for MFA at the identifier level, but we can identify that it will need MFA at some point in the chain.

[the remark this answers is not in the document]
abrooksv: Also this interface lives in JetBrains core, due to using `AnAction`

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -5,22 +5,24 @@ package software.aws.toolkits.core.credentials
...
 
     /**
-     * A user friendly display name shown in the UI.
+     * An optional shortened version of the name to display in the UI where space is at a premium
      */
-    abstract val displayName: String
+    val shortName: String get() = displayName
+}
 
+abstract class ToolkitCredentialsIdentifier : CredentialIdentifier {
```
kiiadi, line 20: Maybe rename `CredentialIdentifierBase` ? I'm kinda following the same pattern with the `ProjectAccountSettingsManager` refactor in that I'm thinking:

```kt

interface AwsConnectionManager

abstract class AwsConnectionManagerBase : AwsConnectionManager

class DefaultAwsConnectionManager(delegate: AwsConnectionManagerBase) : AwsConnectionManager by delegate
```

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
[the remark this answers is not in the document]
kiiadi: Yeah - that's fair - maybe your point an where it lives due to being `AnAction` is an indicator that we should have gone for the `() -> Unit` approach for the action representation so we could just de-couple it. Ultimately that's what we'd do when this gets pushed down.

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -6,6 +6,12 @@ package software.aws.toolkits.core.credentials
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
 interface CredentialIdentifier {
+    /**
+     * The ID must be unique across all [CredentialIdentifier].
```
kiiadi, line 10: feels like this sentence is missing "instances" at the end.

core/tst/software/aws/toolkits/core/credentials/Mocks.kt
```diff
@@ -8,7 +8,7 @@ import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 import software.aws.toolkits.core.utils.test.aString
 
 fun aToolkitCredentialsProvider(
-    identifier: ToolkitCredentialsIdentifier = aToolkitCredentialsIdentifier(),
+    identifier: CredentialIdentifier = aToolkitCredentialsIdentifier(),
```
kiiadi, line 11: the `aToolkitCredentialsIdentifier` generator should be renamed to `aCredentialIdentifier` also.

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -6,6 +6,12 @@ package software.aws.toolkits.core.credentials
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
 interface CredentialIdentifier {
```
kiiadi, line 8: Should probably make a comment that this should *not* be extended directly but that implementers should extend the base.

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
```diff
@@ -15,31 +21,25 @@ interface CredentialIdentifier {
...
 
     /**
      * Some ID types (e.g. Profile) have a concept of a default region, this is optional.
      */
-    open val defaultRegionId: String? get() = null
+    val defaultRegionId: String? get() = null
+}
 
+abstract class CredentialIdentifierBase : CredentialIdentifier {
     override fun equals(other: Any?): Boolean {
```
kiiadi, line 38: we may want to make this explicitly `final` so that it cannot be overridden in an implementer - just did a quick test even though typically `fun` is `final` when we're overriding a base-class method you have to be explicit to stop extenders being able to trump the implementation.

*or* maybe we do want future implementers to be able to have their own equals - I suspect not tho.

abrooksv: Good call out, will fix

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
abrooksv: Fixed

core/tst/software/aws/toolkits/core/credentials/Mocks.kt
abrooksv: Changed

core/src/software/aws/toolkits/core/credentials/ToolkitCredentialsProvider.kt
abrooksv: Added

PR #1829, opened 2020-06-08
RDS draft PR 

hunterwerlla:


jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/rds/actions/CreateConfigurationAction.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    private fun createDatasource(username: String, database: String) {
+        val endpoint = node.dbInstance.endpoint()
+        val url = "${endpoint.address()}:${endpoint.port()}"
+        val source = LocalDataSourceManager.getInstance(node.nodeProject)
+        val dataSource = LocalDataSource().also {
+            // turn on url only config since we don't need any additional config
+            it.isConfiguredByUrl = true
+            it.authProviderId = IamAuth.providerId
+            it.additionalJdbcProperties[CREDENTIAL_ID_PROPERTY] = node.nodeProject.activeCredentialProvider().id
```
kassak, line 73: Just wanted to note that, these properties won't be passed to jdbc (despite it's name), they are to be used by interceptor (I see that they are used, but anyway).

PR #1839, opened 2020-06-10
Change build specs to match standard 4.0

abrooksv:
Note: This PR will NOT build due to we need these in place for CI tests on the pipeline to pass to get the image upgrades out.

## License


abrooksv:
Tested so far is linux unit 2019.3

PR #1845, opened 2020-06-12
Add Redshift explorer nodes

hunterwerlla:
Screen Shot 2020-06-12 at 9 49 42 AM

## License


jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/redshift/RedshiftUtils.kt
```diff
@@ -0,0 +1,17 @@
...
+package software.aws.toolkits.jetbrains.services.redshift
+
+import com.intellij.openapi.project.Project
+import software.amazon.awssdk.services.redshift.model.Cluster
+import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.core.utils.tryOrNull
+import software.aws.toolkits.jetbrains.core.AwsResourceCache
+import software.aws.toolkits.jetbrains.services.sts.StsResources
+
+fun Project.clusterArn(cluster: Cluster, region: AwsRegion): String {
```
kiiadi, line 13: We should probably make a generic 'arn' util, I could see something like:

```kt
Project.arn(
    serviceName: String, 
    resourceType: String, 
    resourceIdentifier: String, 
    connection: ConnectionSettings? = null
)
```

Could be something we do in the future tho - could take a leaf out of the CDK book : https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/core/lib/arn.ts

jetbrains-ultimate/tst-201/software/aws/toolkits/jetbrains/services/redshift/RedshiftExplorerNodeTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerEmptyNode
+import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerErrorNode
+import software.aws.toolkits.jetbrains.core.explorer.nodes.RedshiftExplorerRootNode
+
+class RedshiftExplorerNodeTest {
+    @JvmField
+    @Rule
+    val projectRule = ProjectRule()
+
+    private fun resourceCache() = MockResourceCache.getInstance(projectRule.project)
```
kiiadi, line 24: Would you mind creating a new `ResourceCacheRule` test rule similar to [`MockClientManager`](https://github.com/aws/aws-toolkit-jetbrains/blob/d75ddd35d87e108f03ce19a1f6ea5c602f4e5aed/jetbrains-core/tst/software/aws/toolkits/jetbrains/core/MockClientManager.kt#L59) - I'm hoping to gradually deprecate direct use of the "Mock" instances - and replace them with rules so that we can better control their lifecycle.

jetbrains-ultimate/tst-201/software/aws/toolkits/jetbrains/services/redshift/RedshiftExplorerNodeTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun `Exception thrown makes error node`() {
+        resourceCache().addEntry(RedshiftResources.LIST_CLUSTERS, CompletableFutureUtils.failedFuture(RuntimeException("Simulated error")))
+        val serviceRootNode = rootNode.buildServiceRootNode(projectRule.project)
+        assertThat(serviceRootNode.children).hasOnlyOneElementSatisfying {
+            it is AwsExplorerErrorNode
+        }
+    }
+
+    private companion object {
+        val rootNode = RedshiftExplorerRootNode()
```
kiiadi, line 62: I'm guessing this is your "subject under test" consider renaming to `sut` - it's a pattern I've started using to make it more obvious what is the object being tested.

hunterwerlla: Sure, renamed to `sut`

jetbrains-ultimate/tst-201/software/aws/toolkits/jetbrains/services/redshift/RedshiftExplorerNodeTest.kt
hunterwerlla: Makes a lot of sense, we have this duplicated around way too many places. Added

jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/redshift/RedshiftUtils.kt
hunterwerlla: Yep I mostly included this to get comments, we have to do the same thing for S3 (and probably others) so a util in master would make sense

PR #1852, opened 2020-06-15
Trying to get around JVM bug

kiiadi:


.github/workflows/mac.yml
```diff
@@ -17,11 +17,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 11
+    - run: curl -OL "https://corretto.aws/downloads/latest/amazon-corretto-11-x64-macos-jdk.tar.gz"
```
hunterwerlla, line 20: Can you add a comment in the file on why this is here so we can remove it when they upgrade the jvm version?

kiiadi: why remove it? I figure we could continue using corretto...

hunterwerlla: Speed? But, fair it should be fine either way

justinmk3: If not a source comment,  mentioning the actual bug in the commit message is at least worthwhile.

abrooksv:
Do we still need this/ It seems the string that Coretto team gave us to disable C2 is working to prevent segfaults: https://github.com/aws/aws-toolkit-jetbrains/blob/master/jetbrains-core/build.gradle#L48

PR #1864, opened 2020-06-24
Redo build system - Changelog, misc cleanup

abrooksv:
This is the first part of redo-ing the build system

It removes the unused singing stuff, cleans up change log generation, and moves to the new task avoidance APIs in a few areas

## License


buildSrc/build.gradle
```diff
@@ -26,10 +26,11 @@ repositories {
 }
 
 apply plugin: 'kotlin'
+apply plugin: "java-gradle-plugin"
 
 sourceSets {
-    main.java.srcDir 'src'
-    test.java.srcDir 'tst'
+    main.kotlin.srcDir 'src'
```
hunterwerlla, line 32: What does this actually change versus the old version?

buildSrc/src/toolkits/gradle/changelog/ReleaseCreator.kt
```diff
@@ -6,7 +6,7 @@ package toolkits.gradle.changelog
 import java.io.File
 import java.time.LocalDate
 
-class ReleaseCreator(private val unreleasedFiles: List<File>, private val nextReleaseFile: File) {
+class ReleaseCreator(private val unreleasedFiles: Collection<File>, private val nextReleaseFile: File) {
```
hunterwerlla, line 9: What does this change?

abrooksv: The call site switch to returning a Set<File> instead, but since we dont care about the actual list vs set, I went up a level in the inheritance tree

buildSrc/build.gradle
abrooksv: Nothing, just felt more clear since all of this code is Kotlin and not java.

PR #1887, opened 2020-06-30
Allow proxy and IP RDS urls

hunterwerlla:
- Use instance ID to get the correct inputs to the presigner
- Cleanup and simplify code

## Screenshots
Kapture 2020-06-30 at 10 35 35

## License


jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/rds/auth/IamAuthWidget.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.ui.components.JBTextField
+import org.jetbrains.annotations.TestOnly
+import software.aws.toolkits.jetbrains.services.rds.RdsResources
+import software.aws.toolkits.jetbrains.ui.AwsAuthWidget
+import software.aws.toolkits.resources.message
+import javax.swing.JPanel
+
+const val INSTANCE_ID_PROPERTY = "AWS.RdsInstanceId"
+
+class RdsAwsAuthWidget : AwsAuthWidget() {
```
abrooksv, line 21: Make class name match file name since only one class

hunterwerlla: Ah refactor didn't work properly because there's that floating const val, fixed

PR #1890, opened 2020-07-01
Fixing feedback comment length text

kiiadi:


kiiadi:
> Suggestion: "X character(s) over the limit" for excess

See https://github.com/aws/aws-toolkit-vscode/pull/1171#issuecomment-652532499

PR #1894, opened 2020-07-06
Clean up DotNet CloudDebug code

abrooksv:
* Remove code paths from < 193
* Remove ability to run debugger under mono
* Remove custom dbgshim
* Fix getExecutionRole not respecting the run config credentials

## License


jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
```diff
@@ -494,44 +396,14 @@ class DotNetDebuggerSupport : DebuggerSupport() {
...
-                    // remove existing linux Mono distribution since we run debugger on container .net core
-                    linuxMonoSubdirectory.deleteRecursively()
-                } catch (e: Throwable) {
-                    logger.trace(e) { "Error while trying to delete unused linux Mono directory ${linuxMonoSubdirectory.absolutePath}" }
-                }
+        val linuxMonoSubdirectory = File(targetPath, linuxSubdirectoryName)
+        if (linuxMonoSubdirectory.isDirectory) {
+            try {
+                // remove existing linux Mono distribution since we run debugger on container .net core
+                linuxMonoSubdirectory.deleteRecursively()
```
kiiadi, line 403: assume this is to save space? is this destructive to the users local setup?

jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
```diff
@@ -247,7 +191,7 @@ class DotNetDebuggerSupport : DebuggerSupport() {
 
         val debugArgs = StringBuilder()
             .append("RESHARPER_HOST_LOG_DIR=$remoteDebuggerLogPath ")
```
hunterwerlla, line 193: nit: we could change it to a list then `.joinToString(' ')` then we wouldn't have to have the trailing spaces

jetbrains-rider/src/software/aws/toolkits/jetbrains/services/clouddebug/DotNetDebuggerSupport.kt
abrooksv: It copies everything into our own directory and then removes mono

abrooksv: Space is not the concern, it is the amount we have to rsync over. Copying mono slows down the copy by a few minutes if I remember right...

abrooksv: This logic also has not hanged, it was the else branch of not using dotnet core. Using dotnet core debugger has been the default for us for 193+, but we had a registry option to disable in case of instability was found. This registry option has been removed since no issues have been found

kiiadi: > it is the amount we have to rsync over.

this is what I meant

PR #1901, opened 2020-07-09
Remove the unnecessary Netty dependency so we can trim our plugin size

kiiadi:
Saves 3m on the packaged plugin.zip


abrooksv:
Do we need software.amazon.awssdk:netty-nio-client?

kiiadi:
> Do we need software.amazon.awssdk:netty-nio-client?

Nope - changed - and that gets rid of all the reactive streams netty dependencies too - doesn't really affect the size tho - still 19MB.

build.gradle
```diff
@@ -68,6 +68,7 @@ allprojects {
         runtimeClasspath.exclude group: "org.slf4j"
         runtimeClasspath.exclude group: "org.jetbrains.kotlin"
         runtimeClasspath.exclude group: "org.jetbrains.kotlinx"
+        runtimeClasspath.exclude group: "software.amazon.awssdk", module: "netty-nio-client"
```
hunterwerlla, line 71: Did you accidentally remove the change?

kiiadi: No - there's no need to explicitly exclude the netty dependency - by removing the `nett-nio-client` - its dependencies don't get resolved.

PR #1915, opened 2020-07-14
Fix OOM when running tests

abrooksv:
* Fix project leak by not passing a disposable to the message bug in the resource cache
* Add hack to clear the inline mock cache that gets used when using Kotlin + Mockito
* Fix ResourceCache test using project rule incorrectly leading to leak

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
@@ -4,6 +4,7 @@
 package software.aws.toolkits.jetbrains.core
 
 import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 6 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -390,4 +391,7 @@ class DefaultAwsResourceCache(
             return removed
         }
     }
+
+    override fun dispose() {
```
kiiadi, line 395: Why do we need this? If we are going to dispose should we not clear out the cahce?

abrooksv:
> how

How what?

Mockito kotlin uses inline mocks, this was leading to a bunch of mocks not getting cleaned up and leaking references to Projects which eventually leads to OOMs.

The message bus issue was that the ResourceCache registers itself to the message bus as a listener. When a project is closed, we never disconnected it so it would also leak. We now use the project service as the parent disposable, so when the project is closed and the project service gets disposed the resource cache disconnects itself off the message bus.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
abrooksv: We need this because the Disposable interface that we are using to register to the message bus has no default impl, but good call out, we should clear the cache hash map just in case here as well

kiiadi:
> > how
> 
> How what? 😄
> 
> Mockito kotlin uses inline mocks, this was leading to a bunch of mocks not getting cleaned up and leaking references to Projects which eventually leads to OOMs. According to hprof files, the first test failure already had 113 project references leaked 😦
> 
> The message bus issue was that the ResourceCache registers itself to the message bus as a listener. When a project is closed, we never disconnected it so it would also leak. We now use the project service as the parent disposable, so when the project is closed and the project service gets disposed the resource cache disconnects itself off the message bus.
> 
> Any projectService can be marked Disposable and handled by the framework, which is what we are doing here: https://jetbrains.org/intellij/sdk/docs/basics/disposers.html#automatically-disposed-objects

yeah no sorry - that was a hold over from when I was still scrolling through. the implementation makes sense.

PR #1919, opened 2020-07-15
Added SQS tool window and actions

ranyoo2367:
Added SQS tool window where the user can view message sampling and send a message.

## Types of changes

## Description

- Added SQS tool window. Each queue window has two tabs: "Message Sampling" and "Send a Message". 
- Two actions are added to the context menu in the AWS Explorer: "View Sample Messages" and "Send a Message"
- Double-clicking the queue or selecting "View Sample Messages" action open the queue window and direct the user to the "Message Sampling" tab. 
- Selecting "Send a Message" action directs the user to the "Send a Message" tab.
- For now, opening a queue that is already exists in the tool window will cause the existing tab to dispose and reopen. 

## Testing

There are no functional tests for two actions because their current functionalities are to open the tool window. Tests will be implemented in the future.

SqsToolWindow

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+            LOG.error(e) { "Exception thrown while trying to open queue '${queue.queueName}'" }
+            throw e
+        }
+    }
+
+    companion object {
+        internal val SQS_TOOL_WINDOW = ToolkitToolWindowType(
+            "AWS.Sqs",
+            message("sqs.toolwindow"),
+            EmptyIcon.ICON_0 // TODO: Get and change icons
```
hunterwerlla, line 61: nit: would prefer `AllIcons.Nodes.EmptyNode` even though it's 16x16 and it's getting replaced immedetly

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+        fun getInstance(project: Project) = ServiceManager.getService(project, SqsWindow::class.java)
+        private val LOG = getLogger<SqsWindow>()
+    }
+}
+
+// Will add more parameters once window is populated
+class SqsWindowUI(private val project: Project) {
```
hunterwerlla, line 70: This should go into its own file, also should make some empty form files for the panels

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+    private val toolWindow = ToolkitToolWindowManager.getInstance(project, SQS_TOOL_WINDOW)
+    private val edtContext = getCoroutineUiContext()
+
+    fun openQueue(queueUrl: String) {
+        showQueue(queueUrl, SqsWindowUI(project))
+    }
+
+    fun sendMessage(queueUrl: String) {
+        val component = SqsWindowUI(project)
+        component.mainPanel.selectedIndex = 1
```
hunterwerlla, line 33: This should be encapsulated in `SqsWindowUI` as a function. Also, I'm not a fan of that magic number, but it should be fine if it is a defined constant like:
`val SEND_MESSAGE_PANE=1`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+
+class SqsWindow(private val project: Project) : CoroutineScope by ApplicationThreadPoolScope("openQueue") {
+    private val toolWindow = ToolkitToolWindowManager.getInstance(project, SQS_TOOL_WINDOW)
+    private val edtContext = getCoroutineUiContext()
+
+    fun openQueue(queueUrl: String) {
+        showQueue(queueUrl, SqsWindowUI(project))
+    }
+
+    fun sendMessage(queueUrl: String) {
```
hunterwerlla, line 31: Same comment about passing in queue, and it should be stored in `SqsWindowUI`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsWindow.kt
```diff
@@ -0,0 +1,75 @@
...
+        try {
+            val existingWindow = toolWindow.find(queueUrl)
+            if (existingWindow != null) {
+                withContext(edtContext) {
+                    existingWindow.dispose()
+                }
+            }
+
+            withContext(edtContext) {
+                toolWindow.addTab(queue.queueName, component.mainPanel, true, queueUrl)
```
hunterwerlla, line 49: nit: Not sure if we do this on the other ones, but should use named arguments for the bool like `active = true`

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -668,4 +668,8 @@ settings.regions.none_selected=No region selected
 settings.regions.recent=Recent Regions
 settings.regions.region_sub_menu=All Regions
 settings.title=AWS Connection Settings
+sqs.queue.message.sampling=Message Sampling
```
justinmk3, line 671: is "Sampling" the same word use by SQS in the web console? I wonder if there is a more common name for this.

ranyoo2367: It is not used in the web console - I added "sampling" to make it clear that the user can view only a sample of the messages in the queue. Perhaps "Messages received" or "Polled messages" would be more familiar.

hunterwerlla: How about the wording used on the console "Poll for Messages"

justinmk3: Makes sense,  the idea is to convey that this is a (partial) snapshot at an instant in time. I wonder if there is precedent in tools like WireShark. Or CloudWatch...

I don't have much better suggestions: "snapshot", "glance". Alterntiavely,  UIs like CloudWatch tend to explicitly say actual numbers: something like `X (out of Y) messages polled at 05:42`.

But here I see you just need a reasonable title for the editor tab. `Message Sample` instead of `Message Sampling` might be slightly more familiar IMO. Or if the API supports it, consider `Messages (4 of 73 @ 05:52)`.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindowUI.kt
```diff
@@ -0,0 +1,33 @@
...
+import javax.swing.JPanel
+
+// Will add more parameters once window is populated
+class SqsWindowUI(private val project: Project, val queue: Queue) {
+    val mainPanel = JBTabbedPane().apply {
+        this.add(message("sqs.queue.message.sampling"), JPanel())
+        this.add(message("sqs.queue.send.message"), JPanel())
+    }
+
+    fun openMessage(): SqsWindowUI {
```
hunterwerlla, line 19: This returning `SqsWindowUI` is kind of weird, it's like mixing a builder and a thing to select. instead of doing `showQueue(queue, SqsWindowUI(project, queue).openMessage())`, I think a more Kotliny way to do it is `showQueue(queue, SqsWindowUI(project, queue)).apply { openMessage()}`, then openMessage is just the set

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindowUI.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.ui.components.JBTabbedPane
+import software.aws.toolkits.jetbrains.services.sqs.Queue
+import software.aws.toolkits.resources.message
+import javax.swing.JPanel
+
+// Will add more parameters once window is populated
+class SqsWindowUI(private val project: Project, val queue: Queue) {
+    val mainPanel = JBTabbedPane().apply {
+        this.add(message("sqs.queue.message.sampling"), JPanel())
```
hunterwerlla, line 15: nit: use the panes made above (they will need a JPanel added to them which we usually name `component`)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindow.kt
```diff
@@ -0,0 +1,61 @@
...
+            LOG.error(e) { "Exception thrown while trying to open queue '${queue.queueName}'" }
+            throw e
+        }
+    }
+
+    companion object {
+        internal val SQS_TOOL_WINDOW = ToolkitToolWindowType(
+            "AWS.Sqs",
+            message("sqs.toolwindow"),
+            AllIcons.Nodes.EmptyNode // TODO: Get and change icons
```
hunterwerlla, line 55: nit: the icon should be in the base branch now

resources/resources/software/aws/toolkits/resources/localized_messages.properties
justinmk3: > How about the wording used on the console "Poll for Messages"

For the tab title this could be `Polled Messages`.

ranyoo2367: I think changing the name to what console uses would work. In the future, I plan on adding a warning dialog about polling messages whenever the user opens the queue. I can clarify that the messages are only a snapshot in the dialog.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.services.sqs.toolwindow
+
+import javax.swing.JPanel
+
+class PollMessagePane {
+    var component = JPanel()
```
hunterwerlla, line 8: Ah this should be bound from the form file, checkout `CreateBucketPanel` as an example

PR #1943, opened 2020-07-22
Fix RIDER-34717. Fix completion text replacement with an exact handler name

sdubov:
## Types of changes

## Description
- Make sure selected handler completion variant fully replace the value in handler text field in Local Lambda run configuration when completing
- Fix frontend prefix matcher for handler to allow ':' and '.' symbols and correctly set a matcher instance
- Add tests for checking if handler completion feature is supported for added runtimes
- Unify the fix for missing VFS roots in Rider tests

## Motivation and Context
Handler text field in Local lambda run configuration is a single value text field. It makes sense to replace the matched text with selected completion item in full.

## Related Issue(s)
https://youtrack.jetbrains.com/issue/RIDER-34717

## Testing
Verified in Rider (since completion is supported only in Rider IDE). Checked that resolved completion item is replace the matched text in full.

None.

## Checklist

## License


jetbrains-rider/tst/base/AwsReuseSolutionTestBase.kt
```diff
@@ -38,10 +36,14 @@ abstract class AwsReuseSolutionTestBase : BaseTestWithSolutionBase() {
     override val testCaseNameToTempDir: String
         get() = getSolutionDirectoryName()
 
+    // TODO: Remove when https://youtrack.jetbrains.com/issue/RIDER-47995 is fixed FIX_WHEN_MIN_IS_203
```
abrooksv, line 39: FYI, this is not public

jetbrains-rider/tst/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletionTest.kt
```diff
@@ -36,6 +36,8 @@ class DotNetHandlerCompletionTest : BaseTestWithSolution() {
     }
 
     // TODO this test only works on 2019.2. Which we don't support anymore. Fix the test
+    // TODO: This test is failing due to handlers detection logic. I assume it need to be fixed if test is correct.
```
abrooksv, line 39: I think the first TODO was wrong, is this still failing?

jetbrains-rider/tst/base/AwsReuseSolutionTestBase.kt
sdubov: Agh. Sorry, this is an internal task that are internal by default. This is a ticket to handle VFS roots in Rider Test Framework in case we use local environment. I've made it visible for you. Please let me know if you prefer to delete the comment.

jetbrains-rider/tst/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletionTest.kt
sdubov: Yes. This is because of the lambda detection logic. The main question is where is the problem: in test or in backend logic.

jetbrains-rider/tst/base/AwsReuseSolutionTestBase.kt
abrooksv: No, I like the comment, but needed public so we could track this TODO

sdubov: Done.

jetbrains-rider/tst/software/aws/toolkits/jetbrains/services/lambda/completion/DotNetHandlerCompletionTest.kt
sdubov: @abrooksv, I think we can handle this as a separate issue if there are any changes required from any side. What do you think? I can do it in this PR, but need to figure out what to fix here. Could you please help with this?

abrooksv: Ya, we can follow up on this one.

PR #1947, opened 2020-07-23
Remove html based label and use actual characters instead of escaped

abrooksv:
Still renders correctly, but does not require specific Label implementation

Screen Shot 2020-07-23 at 1 58 27 PM

## License


abrooksv:
> Can you also confirm this renders correctly for the run configuration error (I think this message also shows up there)

Screen Shot 2020-07-23 at 2 45 22 PM

PR #1950, opened 2020-07-23
SQS: Added Polled Messages Pane

ranyoo2367:
Added features to the polled messages pane in the tool window for the users to view the polled messages.

## Types of changes

## Description

- Messages table added to the Polled Messages pane:
   - Receives up to 10 messages (set as default)
   - Table displays message ID, message body, sender ID, sent timestamp
   - Message body size is limited to 1KB when displayed. A feature to view the entire message will be implemented separately.
   - Polls new sample set of messages when the user clicks on the refresh button. Updates the number of messages available.
- Displays the total number of messages available in the queue under the table
- Moved `ResizingDateColumnRenderer` from TableUtils.kt to UiUtils.kt:
   - Renamed it to `ResizingColumnRenderer` and takes optional Boolean argument. Adjusted the name change for instances where it was called in TableUtils.kt.
   - Generalized it so it can be used for different types of columns.
   - Only change involves the logic for setting the component text based on the argument.

## Testing

- Added functional tests for PollMessagesPane to check the behavior when the user receives one message, no messages, or error polling messages. 
- Tested that the messages available text updates accordingly. If there is error requesting queue attributes, it checks that the correct error message is displayed.
- Ran tests for message table displaying received messages as expected. Added tests to check that message bodies are correctly displayed for messages shorter, equal to, and greater than 1KB of characters.
- Ran all local tests for services to check that `ResizingColumnRenderer` remains functional. 

PolledMessages@

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/MessagesTable.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.intellij.ui.ScrollPaneFactory
+import com.intellij.ui.TableSpeedSearch
+import com.intellij.ui.table.TableView
+import com.intellij.util.ui.ListTableModel
+import software.amazon.awssdk.services.sqs.model.Message
+import software.aws.toolkits.resources.message
+import javax.swing.JComponent
+import javax.swing.JTable
+
+class MessagesTable : Disposable {
```
abrooksv, line 16: This does not appear to be registered to a parent disposable, but also I do not see it acting as a parent either. Looks like you can remove the Disposable interface

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,101 @@
...
-class PollMessagePane {
+class PollMessagePane(
+    private val project: Project,
+    private val queue: Queue
+) : Disposable {
     lateinit var component: JPanel
+    lateinit var messagesAvailableLabel: JLabel
+    lateinit var tablePanel: SimpleToolWindowPanel
+
+    private val client: SqsClient = project.awsClient()
```
abrooksv, line 31: This should not be creating a new client, but instead pass it along from the explorer so that it continues to operate even after they switch regions / creds

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,101 @@
...
+import software.aws.toolkits.jetbrains.services.sqs.Queue
+import software.aws.toolkits.resources.message
+import javax.swing.JLabel
 import javax.swing.JPanel
 
-class PollMessagePane {
+class PollMessagePane(
+    private val project: Project,
+    private val queue: Queue
+) : Disposable {
```
abrooksv, line 26: This does not appear to be registered to a parent disposable, but also I do not see it acting as a parent either. Looks like you can remove the Disposable interface

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,101 @@
...
+        }
+        requestMessages()
+        addTotal()
+    }
+
+    override fun dispose() {}
+
+    private companion object {
+        const val MAX_NUMBER_OF_MESSAGES = 10
+        const val MESSAGES_AVAILABLE = "Messages Available: "
```
abrooksv, line 100: This should be moved to localized messages file.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -197,3 +204,53 @@ class WrappingCellRenderer(private val wrapOnSelection: Boolean, private val tog
         return component
     }
 }
+
+// Generalized the renderer
+class ResizingColumnRenderer(showSeconds: Boolean? = null) : TableCellRenderer {
+    private val defaultRenderer = DefaultTableCellRenderer()
+    private val formatter: SyncDateFormat? =
+        if (showSeconds != null) {
+            if (showSeconds) {
```
abrooksv, line 213: Combine this to be `if(showSeconds == true)`, it will fail if null

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -197,3 +204,53 @@ class WrappingCellRenderer(private val wrapOnSelection: Boolean, private val tog
         return component
     }
 }
+
+// Generalized the renderer
+class ResizingColumnRenderer(showSeconds: Boolean? = null) : TableCellRenderer {
```
abrooksv, line 209: When would show seconds be nullable?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -197,3 +204,53 @@ class WrappingCellRenderer(private val wrapOnSelection: Boolean, private val tog
         return component
     }
 }
+
+// Generalized the renderer
```
abrooksv, line 208: This comment does not add value here, I would remove it

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessageUtilsTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        val column2 = MessageBodyColumn().valueOf(message2)
+        // Message 3 has length of 1025
+        val column3 = MessageBodyColumn().valueOf(message3)
+
+        assertThat(column1?.length).isEqualTo(message1.body().length)
+        assertThat(column2?.length).isEqualTo(MAX_LENGTH)
+        assertThat(column3?.length).isEqualTo(MAX_LENGTH)
+    }
+
+    private companion object {
```
abrooksv, line 45: FYI, if the companion object is private, you do not need to make all the variables in it private too.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
ranyoo2367: It would be nullable for columns that do not handle date/time. If you look at lines 231-238, this will handle the component text differently based on `formatter`.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
ranyoo2367: Yes, that would be optimal. I do have a concern though because from my understanding about service classes like `SqsWindow`, I am not allowed to pass any arguments other than Project. If that's the case, I would not be able to pass it along from explorer, but instead, I can move this line to SqsWindow so I wouldn't have to create a new client every time `PollMessagePane` is called. Please correct me if I am wrong.

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
hunterwerlla: We can make this `abstract` then have a getText or something that is implemented by the child class, then we would not have to worry about this unrelated field

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessageUtilsTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        val MAX_LENGTH = 1024
+        val message1: Message = Message.builder()
+            .body("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
+            .messageId("AAAAAAAAAAA")
+            .attributes(mapOf(Pair(MessageSystemAttributeName.SENDER_ID, "1234567890:test1"), Pair(MessageSystemAttributeName.SENT_TIMESTAMP, "111111111")))
+            .build()
+        val message2: Message = Message.builder()
+            .body("VJBIGiHnyegmcoGOEqS7EfHB8xU2CTyUHuJOPDap1ThwWFJ3WbdBlj73YGJtWQM8FWYeBxxUzeZOq7kW9WuYQGoFZ923Gm1jz78N9RnX8Z7V8mLMbA32CjBopsmUYKACH3" +
```
hunterwerlla, line 53: To make a multiline string, you can use """<string>""". However, in this case it looks like you are making random text, so you can use `RuleUtils.randomName` then pass it the length you want

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindowUI.kt
```diff
@@ -5,13 +5,21 @@ package software.aws.toolkits.jetbrains.services.sqs.toolwindow
...
+class SqsWindowUI(
+    private val project: Project,
+    private val client: SqsClient,
+    val queue: Queue
+) {
     val mainPanel = JBTabbedPane().apply {
-        this.add(message("sqs.queue.polled.messages"), PollMessagePane().component)
+        tabComponentInsets = JBUI.emptyInsets()
+        border = JBUI.Borders.empty()
+        this.add(message("sqs.queue.polled.messages"), PollMessagePane(project, client, queue).component)
```
hunterwerlla, line 22: nit: it's weird to have `this` here but not above. With apply it's obvious what it's being applied to so `this` should be removed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindowUI.kt
```diff
@@ -5,13 +5,21 @@ package software.aws.toolkits.jetbrains.services.sqs.toolwindow
...
 import com.intellij.ui.components.JBTabbedPane
+import com.intellij.util.ui.JBUI
+import software.amazon.awssdk.services.sqs.SqsClient
 import software.aws.toolkits.jetbrains.services.sqs.Queue
 import software.aws.toolkits.resources.message
 
 // Will add more parameters once window is populated
-class SqsWindowUI(private val project: Project, val queue: Queue) {
+class SqsWindowUI(
+    private val project: Project,
```
hunterwerlla, line 15: project is now no longer used? remove?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,95 @@
...
+        // Remove all entries
+        while (messagesTable.tableModel.rowCount != 0) {
+            messagesTable.tableModel.removeRow(0)
+        }
+        requestMessages()
+        addTotal()
+    }
+
+    private companion object {
+        const val MAX_NUMBER_OF_MESSAGES = 10
```
hunterwerlla, line 94: This is an api limit, can we stick this in a more central file?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessageUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+
+class MessageBodyColumn : ColumnInfo<Message, String>(message("sqs.message.message_body")) {
+    private val renderer = WrappingCellRenderer(wrapOnSelection = true, toggleableWrap = false)
+    override fun valueOf(item: Message?): String? = StringUtils.abbreviate(item?.body(), MAX_LENGTH_OF_MESSAGES)
+    override fun isCellEditable(item: Message?): Boolean = false
+    override fun getRenderer(item: Message?): TableCellRenderer? = renderer
+
+    private companion object {
+        // Truncated the message body to show up to 1KB, as it can be up to 256KB in size. Cannot limit the retrieved message size through API.
+        const val MAX_LENGTH_OF_MESSAGES = 1024
```
hunterwerlla, line 31: Would prefer this in a more central location like `SQSUtils.kt` or something along those lines?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/MessagesTable.kt
```diff
@@ -0,0 +1,49 @@
...
+            autoResizeMode = JTable.AUTO_RESIZE_LAST_COLUMN
+            setPaintBusy(true)
+            emptyText.text = message("loading_resource.loading")
+        }
+
+        TableSpeedSearch(table)
+        component = ScrollPaneFactory.createScrollPane(table)
+    }
+
+    fun showBusy(busy: Boolean) {
```
hunterwerlla, line 41: `setBusy`? showBusy implies it shows busy only

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,95 @@
...
+import software.amazon.awssdk.services.sqs.model.Message
+import software.amazon.awssdk.services.sqs.model.QueueAttributeName
+import software.aws.toolkits.jetbrains.services.sqs.Queue
+import software.aws.toolkits.resources.message
+import javax.swing.JLabel
 import javax.swing.JPanel
 
-class PollMessagePane {
+class PollMessagePane(
+    private val project: Project,
```
hunterwerlla, line 22: project seems to be unused, remove?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,95 @@
...
+        } catch (e: Exception) {
+            messagesAvailableLabel.text = message("sqs.failed_to_load_total")
+        }
+    }
+
+    private fun addToolbar() {
+        val actionGroup = DefaultActionGroup()
+        actionGroup.addAction(object : AnAction(message("general.refresh"), null, AllIcons.Actions.Refresh), DumbAware {
+            override fun actionPerformed(e: AnActionEvent) {
+                refreshTable()
```
hunterwerlla, line 75: This will also be run on the UI thread

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,95 @@
...
+        tablePanel.setContent(messagesTable.component)
+
+        requestMessages()
+        addTotal()
+        addToolbar()
+    }
+
+    private fun requestMessages() {
+        try {
+            val polledMessages: List<Message> = client.receiveMessage {
```
hunterwerlla, line 45: This will end up making a network call on the UI thread (since it's called in `init`), this needs to be put in the background (preferentially using coroutines)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessageUtils.kt
ranyoo2367: Sounds like a good solution, I will create a new SqsUtils file.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -699,6 +699,14 @@ settings.states.invalid.short=Unable to connect
 settings.states.validating=Validating connection to AWS...
 settings.states.validating.short=Validating connection
 settings.title=AWS Connection Settings
+sqs.failed_to_load_total=Failed to load approximate total number of messages available
+sqs.failed_to_poll_messages=Failed to poll messages
+sqs.message.message_body=Message Body
+sqs.message.message_id=Message ID
+sqs.message.no_messages= No messages polled
```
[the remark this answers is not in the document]
ranyoo2367: I agree it is a little weird

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,105 @@
...
+            override fun actionPerformed(e: AnActionEvent) {
+                refreshTable()
+            }
+        })
+        tablePanel.toolbar = ActionManager.getInstance().createActionToolbar("PollMessagePane", actionGroup, false).component
+    }
+
+    // TODO: Add message table actions
+
+    private fun refreshTable() = launch {
```
hunterwerlla, line 97: Can you add a TODO to add debounce or throttling or something

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
```diff
@@ -2,8 +2,105 @@
...
+        messagesTable.setBusy(busy = true)
+        // Remove all entries
+        while (messagesTable.tableModel.rowCount != 0) {
+            messagesTable.tableModel.removeRow(0)
```
hunterwerlla, line 101: nit: the table model can just be set to empty

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindowUI.kt
```diff
@@ -3,20 +3,38 @@
...
-class SqsWindowUI(private val project: Project, val queue: Queue) {
+class SqsWindowUI(
+    private val client: SqsClient,
+    val queue: Queue
+) {
+    private val pane = PollMessagePane(client, queue)
+    // This makes sure that PollMessagePane is properly set up when switched to the Poll Message tab
+    private val changeListener = ChangeListener { e ->
+        val selected = e?.source as JBTabbedPane
+        if ((selected.selectedIndex == POLL_MESSAGE_PANE) && (!pane.isSetup)) {
```
hunterwerlla, line 22: What is the isSetup for? can this be a function called by the thing that creates the table after it is created/in init?

core/tst/software/aws/toolkits/core/utils/RuleUtils.kt
```diff
@@ -11,6 +11,11 @@ object RuleUtils {
         return "${prefix.toLowerCase()}-${userName.toLowerCase()}-${Random().nextInt(10000)}".take(length)
     }
 
+    fun randomName(length: Int): String {
```
hunterwerlla, line 14: This should replace the existing randomName

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePaneTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    private lateinit var label: JLabel
+
+    private val message: Message = Message.builder()
+        .body("ABC")
+        .messageId("XYZ")
+        .attributes(mapOf(Pair(MessageSystemAttributeName.SENDER_ID, "1234567890:test1"), Pair(MessageSystemAttributeName.SENT_TIMESTAMP, "111111111")))
+        .build()
+
+    @Before
+    fun loadVariables() {
```
hunterwerlla, line 44: nit: maybe reset or something

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/PollMessagePane.kt
ranyoo2367: Please correct me if I'm wrong, but I think that's possible with `DefaultTableModel`, but not `ListTableModel`.

PR #1958, opened 2020-07-29
Convert most of the rest of the gradle to Kotlin dsl

hunterwerlla:
## License


resources/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import de.undercouch.gradle.tasks.download.Download
+
+plugins {
+    id("de.undercouch.download")
```
abrooksv, line 7: Add version number to this (4.1.1)

hunterwerlla: It's a bit weird we didn't before, added

hunterwerlla: Oh that causes an error no wonder we didn't do it before, added a comment why we can't

PR #1960, opened 2020-07-30
Make the Lambda sub-system dynamic plugin compatible

abrooksv:
* Nuke the RuntimeGroup enum
* Merge RuntimeGroup with RuntimeGroupInfo
* Migrate vals to funs so that it always queries the EP system
* Use IDs to reference RuntimeGroups now, this could use more clean up in the future

This is draft, because I am not 100% happy with that `RuntimeGroup.findById("JAVA") != JavaRuntimeGroup.INSTANCE`, but the APIs should be considered stable

## License

### DIFF
--- a/jetbrains-core/resources/META-INF/ext-python.xml
+++ b/jetbrains-core/resources/META-INF/ext-python.xml
@@ -7,11 +7,11 @@
     </extensions>
 
     <extensions defaultExtensionNs="aws.toolkit.lambda">
-        <runtimeGroup runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonRuntimeGroup"/>
-        <builder runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaBuilder"/>
-        <handlerResolver runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaHandlerResolver"/>
-        <sam.debugSupport runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamDebugSupport"/>
-        <sam.projectWizard runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamProjectWizard"/>
+        <runtimeGroup implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonRuntimeGroup"/>
+        <builder id="PYTHON" implementationClass="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaBuilder"/>
+        <handlerResolver id="PYTHON" implementationClass="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaHandlerResolver"/>
+        <sam.debugSupport id="PYTHON" implementationClass="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamDebugSupport"/>
+        <sam.projectWizard id="PYTHON" implementationClass="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamProjectWizard"/>
     </extensions>
     <extensions defaultExtensionNs="aws.toolkit.clouddebug">
         <debuggerSupport implementation="software.aws.toolkits.jetbrains.services.clouddebug.python.PythonDebuggerSupport"/>
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
@@ -51,9 +51,7 @@ interface RuntimeGroup {
         @JvmStatic
         fun find(predicate: (RuntimeGroup) -> Boolean): RuntimeGroup? = registeredRuntimeGroups().firstOrNull(predicate)
 
-        fun findById(id: String): RuntimeGroup? = find { it.id == id }
-
-        fun getById(id: String): RuntimeGroup = findById(id) ?: throw IllegalStateException("No RuntimeGroup with $id is registered")
+        fun getById(id: String): RuntimeGroup = find { it.id == id } ?: throw IllegalStateException("No RuntimeGroup with $id is registered")
 
         fun determineRuntime(project: Project?): Runtime? = project?.let { _ ->
             registeredRuntimeGroups().asSequence().mapNotNull { it.determineRuntime(project) }.firstOrNull()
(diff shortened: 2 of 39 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
```diff
@@ -17,108 +16,55 @@ import com.intellij.openapi.projectRoots.SdkType
...
-enum class RuntimeGroup {
-    JAVA,
-    PYTHON,
-    NODEJS,
-    DOTNET;
-
-    private val info by lazy {
-        RuntimeGroupInformation.getInstances(this)
-    }
+interface RuntimeGroup {
```
abrooksv, line 27: To fix the equality issue, this will become an abstract class. I am going to model it after `Language` if curious

hunterwerlla:
What was your fix for the comment in the description "his is draft, because I am not 100% happy with that RuntimeGroup.findById("JAVA") != JavaRuntimeGroup.INSTANCE, but the APIs should be considered stable"?

jetbrains-core/resources/META-INF/ext-python.xml
```diff
@@ -7,11 +7,11 @@
     </extensions>
 
     <extensions defaultExtensionNs="aws.toolkit.lambda">
-        <runtimeGroup runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonRuntimeGroup"/>
-        <builder runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaBuilder"/>
-        <handlerResolver runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaHandlerResolver"/>
-        <sam.debugSupport runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamDebugSupport"/>
-        <sam.projectWizard runtimeGroup="PYTHON" implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonSamProjectWizard"/>
+        <runtimeGroup implementation="software.aws.toolkits.jetbrains.services.lambda.python.PythonRuntimeGroup"/>
+        <builder id="PYTHON" implementationClass="software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaBuilder"/>
```
hunterwerlla, line 11: Why do these need ID? Can the be children of runtimeGroup? If we need them across multiple files can we merge them together? It's weird that above the runtime group defines the ID, but you'd need to go to the runtime group to know that. Then it is used below but they don't necessarily look connected.

abrooksv: I can also make runtimeGroup's ID be here too for consistency sake

abrooksv:
> What was your fix for the comment in the description "his is draft, because I am not 100% happy with that RuntimeGroup.findById("JAVA") != JavaRuntimeGroup.INSTANCE, but the APIs should be considered stable"?

Ended up moving away from `.INSTANCE` methods. Always query for it by ID if you need it

jetbrains-core/resources/META-INF/ext-python.xml
hunterwerlla: Thinking about this more I'm not sure what is preferable, I think this is fine

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/LambdaRunConfigurationType.kt
```diff
@@ -20,10 +20,10 @@ class LambdaRunConfigurationType :
         AwsIcons.Resources.LAMBDA_FUNCTION
     ) {
     init {
-        // Although it should work, isApplicable doesn't seem to work for locallambdarunconfigurationfactory
-        // and it still shows up when it is not applicalbe. So we have to decide in the configuration to add it or not.
+        // Although it should work, isApplicable doesn't seem to work for LocalLambdaRunConfigurationFactory
+        // and it still shows up when it is not applicable. So we have to decide in the configuration to add it or not.
         // TODO see if this is resolvable
```
hunterwerlla, line 25: Is this resolvable now that it's dynamic? :P

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
```diff
@@ -17,108 +16,67 @@ import com.intellij.openapi.projectRoots.SdkType
...
-        /**
-         * Lazily apply the predicate to each [RuntimeGroup] and return the first match (or null)
-         */
         @JvmStatic
-        fun find(predicate: (RuntimeGroup) -> Boolean): RuntimeGroup? = RuntimeGroup.values().asSequence().filter(predicate).firstOrNull()
+        fun find(predicate: (RuntimeGroup) -> Boolean): RuntimeGroup? = registeredRuntimeGroups().firstOrNull(predicate)
+
+        fun findById(id: String): RuntimeGroup? = find { it.id == id }
+
+        fun getById(id: String): RuntimeGroup = findById(id) ?: throw IllegalStateException("No RuntimeGroup with $id is registered")
```
hunterwerlla, line 56: The distinction between find/get follows convention but it's not immediately obvious from the name. Can this be `get` and `getOrNull`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
```diff
@@ -139,38 +97,24 @@ fun AnActionEvent.runtime(): Runtime? {
...
+ * See [software.aws.toolkits.jetbrains.services.lambda.LambdaBuilder]
  */
-class RuntimeGroupExtensionPoint<T> : AbstractExtensionPointBean(), KeyedLazyInstance<T> {
+abstract class RuntimeGroupExtensionPointObject<T>(private val extensionPointName: ExtensionPointName<IdBasedExtensionPoint<T>>) {
+    private val collector = KeyedExtensionCollector<T, String>(extensionPointName.name)
 
-    @Attribute("implementation")
-    lateinit var implementation: String
+    fun getInstance(runtimeGroup: RuntimeGroup): T? = collector.findSingle(runtimeGroup.id)
+    fun getInstanceOrThrow(runtimeGroup: RuntimeGroup): T = getInstance(runtimeGroup)
```
hunterwerlla, line 107: Same comment about making it `getInstance`/`getInstanceOrNull`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/LambdaRunConfigurationType.kt
abrooksv: Don't think so, they check the factory is applicable, if I remember right. Not the children in the factory.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
abrooksv: Ya

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/RuntimeGroup.kt
abrooksv: Agree, we should prioritize `getInstanceOrThrow` since that is what we use more

PR #1961, opened 2020-07-31
Convert core and ultimate gradle to kotlin

hunterwerlla:
## License


jetbrains-core/build.gradle.kts
```diff
@@ -0,0 +1,106 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import groovy.lang.Closure
```
kiiadi, line 4: what the heck is this?

hunterwerlla: This is so that we can call into functions defined in groovy (it's the type of the top level funcitons), we will be able to remove it when the whole thing is kotlin

PR #1983, opened 2020-08-12
Add resource filtering service + tag based filtering

hunterwerlla:
- Add resource filtering service along with tag based filtering
- We apply the same filtering logic client side so that we can cache all of the tagged resources of a resource type so we don't have to make more calls as filters change
- The next PRs will add CloudFormation stack based filtering and the UI

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/filtering/ResourceFilterManager.kt
```diff
@@ -0,0 +1,62 @@
...
+                }
+        }
+    }
+
+    companion object {
+        fun getInstance(project: Project): ResourceFilterManager = ServiceManager.getService(project, ResourceFilterManager::class.java)
+    }
+}
+
+sealed class ResourceFilter(open val enabled: Boolean)
```
abrooksv, line 50: Can we get unit tests on the serialization of these like our other managers

jetbrains-core/src/software/aws/toolkits/jetbrains/core/filtering/FilterManager.kt
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.filtering
+
+import com.intellij.openapi.project.Project
+
+interface FilterManager {
```
abrooksv, line 8: This is confusing how it differs from Filter.

The lines between this, Filter, FilterManagerService are a bit too blurred from the API definitions.

This seems to duplicate all the fields, so its not clear why Filter exists

jetbrains-core/src/software/aws/toolkits/jetbrains/core/filtering/Filter.kt
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.filtering
+
+interface Filter {
+    val name: String
+    val enabled: Boolean
```
abrooksv, line 8: This is my main comment, I stopped reviewing and wrote this after going only half way since it seemed important to discuss this due to ramifications on the overall PR.


Not every filter should worry about if its enabled or not, that should be handled by the "framework" of filters since the UI for rendering them and toggling them is in the framework.

if we REALLY want to go down this path of making it an EP I think we need to revist a few things, and they all seem to bring me to thinking about this like how run configs are managed.

I'll post the serialized XML since I think it will make the following more clear:

```xml
<resourceFilters>
    <filter type="tag" enabled="true">
        <state>
            <setting name="key" value="MyTag"/>
            <setting name="value" value="Value1,Value2"/>
        </state>
    </filter>
    <filter type="tag" enabled="false">
        <state>
            <setting name="key" value="YourTag"/>
            <setting name="value" value="Value1,Value2"/>
        </state>
    </filter>
    <filter type="cfnStack" enabled="true">
        <state>
            <setting name="stackName" value="App"/>
            <setting name="includeChildren" value="true"/>
        </state>
    </filter>
</resourceFilters>
```

All state should be managed by `FilterManagerService` (This kills current FilterManager becomes FilterFactory FYI, so rename FilterManagerService to FilterManagerService, but I'll use FilterManagerService for clarity since that is current state)

FilterManagerService will keep track of all filters and what are enabled. The job of a filter is to just worry about how to filter, not if it should.

FilterManagerService has the state:
`Map<Name, Pair<Filter, Boolean>>`

Filter is now has something along the lines of added to its interface:
```
    fun loadState(element: Element) {

    }

    fun saveState(element: Element) {

    }
```

When FilterManagerService wants to load it loops over the filters and grabs each type. We then grab the FilterFactory of the matching type. It creates an instance of the filter and then calls loadState on it. Saving does the opposite.

In summary:
FilterManager tracks name, enabled, and the created Filter
FilterFactory can create filter instances (empty, default state), as well as the UI to edit/create them.
Filter has its internal state, a `matches)project: Project, arn, resource, service)` method, load & save methods.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/filtering/FilterManager.kt
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * An ID to uniquely identify the filter
+     */
+    val type: String
+
+    /**
+     * Determine if a resource should be filtered out.
+     * @return true if the resource is not filtered by the currently applied filters or if there are
```
abrooksv, line 16: "return true if the resource matches the current filter"?

PR #1985, opened 2020-08-13
Add support for java8.al2

abrooksv:
## License


jetbrains-core/build.gradle.kts
```diff
@@ -1,15 +1,12 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+// Cannot be removed or else it will fail to compile
```
hunterwerlla, line 4: old comment still left over, also how did this work? It would not import properly without the (actually unused) import

jetbrains-core/build.gradle.kts
```diff
@@ -19,6 +16,7 @@ apply(from = "../intellijJVersions.gradle")
 buildscript {
     val telemetryVersion: String by project
     repositories {
+        mavenLocal()
```
hunterwerlla, line 19: errant inclusion

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfigurationTest.kt
```diff
@@ -103,6 +103,7 @@ class LocalLambdaRunConfigurationTest {
     @Test
     fun samIsNotSet() {
         val fakeSamPath = "NotValid"
+        preWarmLambdaHandlerValidation(projectRule.project)
```
hunterwerlla, line 106: Why is this now needed? add a comment?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
```diff
@@ -77,25 +77,36 @@ class LocalLambdaRunConfiguration(project: Project, factory: ConfigurationFactor
...
+                    runtimeGroup.validateSamVersion(runtime, semVer)
+                } catch (e: Exception) {
+                    throw RuntimeConfigurationError(e.message)
+                }
+            }
+        }
     }
 
-    private fun checkLambdaHandler() {
+    private fun checkLambdaHandler(): Pair<Runtime, String> {
```
hunterwerlla, line 109: Do we use the second return anywhere? what is it for?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfigurationTest.kt
abrooksv: Sam is validated after the handler is validated (so we can get the runtime)

jetbrains-core/build.gradle.kts
abrooksv: Removed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
abrooksv: Nah, I can remove it

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
```diff
@@ -77,25 +77,36 @@ class LocalLambdaRunConfiguration(project: Project, factory: ConfigurationFactor
     }
 
     override fun checkConfiguration() {
-        checkSamVersion()
         resolveRegion()
         resolveCredentials()
-        checkLambdaHandler()
+        val runtime = checkLambdaHandler()
```
hunterwerlla, line 82: I understand this is where we currently get runtime, but it being returned by check is just weird, can we grab runtime in checkSamVersion as well?

abrooksv: we should probably extract it then pass it in if we do that

PR #1995, opened 2020-08-17
Finish the conversion of Gradle to Kotlin

hunterwerlla:
- Move build.gradle to build.gradle.kts
- Move `IntelliJVersions` into buildSrc so we get type checking and completion
- This attempts to be as 1:1 as possible

## License


build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+    val kotlinVersion: String by project
+    val ideaPluginVersion: String by project
+    dependencies {
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
+        classpath("gradle.plugin.org.jetbrains.intellij.plugins:gradle-intellij-plugin:$ideaPluginVersion")
+        classpath("com.adarshr:gradle-test-logger-plugin:1.7.0")
+    }
+}
+
+val ideVersions = IdeVersions(project)
```
abrooksv, line 34: Is there no better way than this?

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+            excludes = listOf("software.aws.toolkits.ktlint.*")
+        }
+
+        reports {
+            junitXml.isEnabled = true
+            html.isEnabled = true
+        }
+    }
+
+    plugins.withType<IdeaPlugin> {
```
abrooksv, line 175: Is this still needed with latest IDE versions + Gradle?

hunterwerlla: Maybe, I don't want this to turn into the change everything about our build PR though. We can make further improvements later.

build.gradle.kts
hunterwerlla: This is the way to do it, otherwise the value does not exist and the compiler fails (see docs: https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:properties)

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+val mockitoKotlinVersion: String by project
+val assertjVersion: String by project
+val junitVersion: String by project
+val remoteRobotPort: String by project
+val ktlintVersion: String by project
+val remoteRobotVersion: String by project
+val ideaPluginVersion: String by project
+// publishing
+val publishToken: String by project
+val publishChannel: String by project
```
kiiadi, line 48: is it possible to do this once in `settings.gradle.kts` and have it available everywhere? In which case I'd probably just specify the values themselves in there instead of being in the `gradle.properties` (with the exception of `toolkitVersion` that needs to be able to be written by CI.

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+
+    java.sourceCompatibility = JavaVersion.VERSION_1_8
+    java.targetCompatibility = JavaVersion.VERSION_1_8
+
+    tasks.withType(JavaExec::class.java) {
+        systemProperty("aws.toolkits.enableTelemetry", false)
+    }
+
+    tasks.withType(RunIdeTask::class.java) {
+        val ijExt = project.extensions.getByName("intellij") as IntelliJPluginExtension
```
kiiadi, line 82: You can do `getByType` eg: `project.extensions.getByType<IntelliJPluginExtension>()` saves you having to do the cast and you don't need to figure out the name.

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+
+    sourceSets {
+        getOrCreate("integrationTest") {
+            java.srcDir("it")
+        }
+    }
+}
+
+subprojects {
+    group = parent!!.group
```
kiiadi, line 115: I know this is necessary but I *hate* the double bangs :) - just a pet peev - don't mind me.

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+    version = parent!!.version
+
+    apply(plugin = "java")
+    apply(plugin = "idea")
+    apply(plugin = "com.adarshr.test-logger")
+
+    sourceSets {
+        main.get().java.srcDirs(findFolders(project, "src", ideVersion))
```
kiiadi, line 123: this greedily gets it right? is there a better way - when I was last poking about with this it was discouraged to do the greedy get - something like 

```
main {
  java.srcDir(findFolders(project, "src", ideVersion))
}
```

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+
+        mustRunAfter(tasks.test)
+    }
+
+    project.plugins.withId("org.jetbrains.intellij") {
+        tasks.withType(org.jetbrains.intellij.tasks.DownloadRobotServerPluginTask::class.java) {
+            this.version = remoteRobotVersion
+        }
+
+        tasks.withType(org.jetbrains.intellij.tasks.RunIdeForUiTestTask::class.java).all {
```
kiiadi, line 216: replace with `tasks.withType<RunIdeForUiTestTask>()` - it's already been imported so you shouldn't need the fully qualified class name.

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+            systemProperty("jb.privacy.policy.text", "<!--999.999-->")
+            if (System.getenv("CI") != null) {
+                systemProperty("aws.sharedCredentialsFile", "/tmp/.aws/credentials")
+            }
+        }
+
+        extensions.getByType<JacocoPluginExtension>().applyTo(tasks.getByName<RunIdeForUiTestTask>("runIdeForUiTests"))
+    }
+
+    tasks.withType(KotlinCompile::class.java).all {
```
kiiadi, line 230: prefer the `<>` style over using the `::class.java`

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+    project.afterEvaluate {
+        removeTask(tasks, org.jetbrains.intellij.tasks.PublishTask::class.java)
+        removeTask(tasks, org.jetbrains.intellij.tasks.VerifyPluginTask::class.java)
+        removeTask(tasks, org.jetbrains.intellij.tasks.BuildSearchableOptionsTask::class.java)
+    }
+}
+
+val ktlint: Configuration by configurations.creating
+
+inline fun <reified T : Task> removeTask(tasks: TaskContainer, takeType: Class<T>) {
```
kiiadi, line 259: I think you can make this into an extension function and remove some of the boiler plate:

```
inline fun <reified T : Task> Project.removeTask() {
    tasks.withType<T> { 
        enabled = false
    }
}
```

Then at the call-site we can change it to :

`removeTask<PublishTask>()`

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+    tasks.withType(takeType).configureEach {
+        enabled = false
+    }
+}
+
+apply(plugin = "org.jetbrains.intellij")
+apply(plugin = "toolkit-change-log")
+
+(project.extensions.getByName("intellij") as IntelliJPluginExtension).apply {
```
kiiadi, line 268: again could replace this with `project.extensions.getByType`

build.gradle.kts
```diff
@@ -0,0 +1,372 @@
...
+        from(sourceSets.getByName("integrationTest").output)
+    }
+
+    artifacts {
+        add("testArtifacts", testJar)
+    }
+
+    // Remove the tasks added in by gradle-intellij-plugin so that we don"t publish/verify multiple times
+    project.afterEvaluate {
+        removeTask(tasks, org.jetbrains.intellij.tasks.PublishTask::class.java)
```
kiiadi, line 251: Generally I think we should favour imports over fully qualified classes.

buildSrc/src/software/aws/toolkits/gradle/SourcesUtils.kt
```diff
@@ -12,9 +14,9 @@ import java.io.FileFilter
  * [type] is the type of the source folder (e.g. 'src', 'tst', 'resources')
  * [ideVersion] is the 3 digit numerical version of the JetBrains SDK (e.g. 192, 201 etc)
  */
-fun findFolders(project: Project, type: String, ideVersion: String): List<String> = project.projectDir.listFiles(FileFilter {
+fun findFolders(project: Project, type: String, ideVersion: String): MutableSet<File> = project.projectDir.listFiles(FileFilter {
     it.isDirectory && includeFolder(type, ideVersion, it.name)
-})?.map { it.name } ?: emptyList()
+})?.map { File(it.name) }?.toMutableSet() ?: mutableSetOf()
```
kiiadi, line 19: why does this need to be mutable?

build.gradle.kts
hunterwerlla: Nope, we don't have access to the get methods of any project or root project. We can put them in buildSrc, but would prefer to do that in a different PR

build.gradle.kts
hunterwerlla: I agree, changed to throwing an exception with a message

build.gradle.kts
hunterwerlla: Ah yeah, changing all of them to that

build.gradle.kts
hunterwerlla: The auto import is wonky, fixed

build.gradle.kts
hunterwerlla: Agree, changed

build.gradle.kts
hunterwerlla: 100%, see wonky imports

build.gradle.kts
hunterwerlla: I was trying to make this as 1:1 as possible, but that's a good improvement, changed

build.gradle.kts
hunterwerlla: replacing with a buildSrc function

buildSrc/src/software/aws/toolkits/gradle/SourcesUtils.kt
hunterwerlla: I was doing something wrong in the conversion and it complained about them being immutable, they can be immutable again.

build.gradle.kts
kiiadi: makes sense - and the `buildSrc` option is probably a decent compromise.

build.gradle.kts
```diff
@@ -79,11 +82,12 @@ allprojects {
     }
 
     tasks.withType(RunIdeTask::class.java) {
-        val ijExt = project.extensions.getByName("intellij") as IntelliJPluginExtension
         val alternativeIde = System.getenv("ALTERNATIVE_IDE")
         if (alternativeIde != null) {
             if (File(alternativeIde).exists()) {
-                ijExt.alternativeIdePath = System.getenv("ALTERNATIVE_IDE")
+                intellij {
+                    alternativeIdePath = System.getenv("ALTERNATIVE_IDE")
```
kiiadi, line 90: nice! much better.

build.gradle.kts
```diff
@@ -112,18 +116,22 @@ configure(subprojects.filter { it.name != "telemetry-client" }) {
 }
 
 subprojects {
-    group = parent!!.group
-    version = parent!!.version
+    group = parent?.group ?: throw IllegalStateException("Subproject $name parent is null!")
```
kiiadi, line 120: massive nit but you could do :

```
parent?.let {
  group = it.group
  version = it.version
} ?: throw IllegalStateException("Subproject $name parent is null!") 
```

hunterwerlla: Sure, changed

PR #1996, opened 2020-08-19
Improve rendering performance of wrapped text in CW logs table

abrooksv:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -169,31 +168,35 @@ private fun JTextArea.speedSearchHighlighter(speedSearchEnabledComponent: JCompo
...
+class WrappingCellRenderer(private val wrapOnSelection: Boolean, private val toggleableWrap: Boolean) : CellRendererPanel(), TableCellRenderer {
     var wrap: Boolean = false
 
-    // JBTextArea has a different font from JBLabel (the default in a table) so harvest the font off of it
-    private val jLabelFont = JBLabel().font
+    private val textArea = JBTextArea()
+
+    init {
+        textArea.font = UIUtil.getLabelFont()
+        textArea.border = JBUI.Borders.empty(2, 2)
```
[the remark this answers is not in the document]
abrooksv: @hunterwerlla Where did these come from?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -169,31 +168,35 @@ private fun JTextArea.speedSearchHighlighter(speedSearchEnabledComponent: JCompo
...
-
-        component.setSize(table.columnModel.getColumn(column).width, component.preferredSize.height)
-        if (table.getRowHeight(row) != component.preferredSize.height) {
-            table.setRowHeight(row, component.preferredSize.height)
+        if (table == null) {
+            return this
+        }
+
+        textArea.lineWrap = (wrapOnSelection && isSelected) || (toggleableWrap && wrap)
+        textArea.text = (value as? String) ?: ""
```
hunterwerlla, line 190: Removed the trim? The reason the trim was there was for when you end up with messages,  like "abc\n\n" which happen, but maybe it's better this way?

abrooksv: Eh, if we want to keep the trim, we should trim before the table model, not every paint call.

If you think we should keep it, I can go hunt that down.

hunterwerlla: (with wrapping turned off)
Screen Shot 2020-08-19 at 7 53 10 AM
Can we re-add trim (and add a comment that this is why)

abrooksv:
Screen Shot 2020-08-19 at 8 20 41 AM

PR #2000, opened 2020-08-20
Stop using project / application as parent disposable

abrooksv:
Uses our services where applicable, or a more more accurate child

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSdkClient.kt
```diff
@@ -21,10 +20,6 @@ import software.aws.toolkits.core.utils.getLogger
 import software.aws.toolkits.core.utils.info
 
 class AwsSdkClient : Disposable {
-    init {
```
kiiadi, line 24: Is the intent that the client manager disposes these?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ExplorerToolWindow.kt
```diff
@@ -88,7 +89,7 @@ class ExplorerToolWindow(project: Project) : SimpleToolWindowPanel(true, true),
         background = UIUtil.getTreeBackground()
         setContent(treePanelWrapper)
 
-        project.messageBus.connect().subscribe(AwsConnectionManager.CONNECTION_SETTINGS_STATE_CHANGED, this)
```
kiiadi, line 91: We should probably have a ktlint rule around this.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsSdkClient.kt
abrooksv: AwsSdkClient is an applicationService, so this is implicitly disposed when either IDE goes away, or plugin is uninstalled (if we ever get dynamic working...)

https://jetbrains.org/intellij/sdk/docs/basics/disposers.html#automatically-disposed-objects

PR #2011, opened 2020-08-24
Save Lambda update settings between updates

hunterwerlla:
# Screenshots
Kapture 2020-08-24 at 08 48 58

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
```diff
@@ -308,12 +313,25 @@ class EditFunctionDialog(
...
     }
 
+    private fun loadSettings() {
+        val updateSettings = UpdateLambdaSettings.getInstance(project)
+        view.sourceBucket.selectedItem = updateSettings.bucketName(name)
+        view.buildInContainer.isSelected = updateSettings.useContainer(name) ?: false
+    }
+
+    private fun saveSettings() {
+        val updateSettings = UpdateLambdaSettings.getInstance(project)
```
kiiadi, line 330: we probably want it bound by lambda rather than by project no?

kiiadi: how will this work if the customer has switched accounts?

I think there's a bit of a hierarchy here - for a given account, pick the bucket based on :
1) lambda if the setting exists
2) fall back to project-wide setting if it exists
3) fall back to app-wide setting if it exists

jetbrains-core/resources/META-INF/plugin.xml
```diff
@@ -155,6 +155,7 @@
         <projectService serviceImplementation="software.aws.toolkits.jetbrains.settings.LambdaSettings"/>
 
         <moduleService serviceImplementation="software.aws.toolkits.jetbrains.settings.DeploySettings"/>
+        <projectService serviceImplementation="software.aws.toolkits.jetbrains.settings.UpdateLambdaSettings"/>
```
kiiadi, line 158: `DeploySettings` is a module service - should this be also?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/EditFunctionDialog.kt
hunterwerlla: We could make it a `applicationService`, but it is currently bound by Lambda name and project, which matches how we do other settings

jetbrains-core/resources/META-INF/plugin.xml
hunterwerlla: According to Austin `ModuleService is an antipattern now`, and it would not make any sense anyway since that maps it backwards.

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/UpdateLambdaSettings.kt
```diff
@@ -0,0 +1,43 @@
...
+@State(name = "updateLambdaSettings", storages = [Storage("aws.xml")])
+class UpdateLambdaSettings : PersistentStateComponent<UpdateSettings> {
+    private var state = UpdateSettings()
+
+    override fun getState(): UpdateSettings? = state
+    override fun loadState(state: UpdateSettings) {
+        this.state = state
+    }
+
+    fun useContainer(stackArn: String): Boolean? = state.updateConfigs[stackArn]?.useContainer
```
kiiadi, line 20: `stackArn` is this copy-pasta?

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/UpdateLambdaSettings.kt
```diff
@@ -0,0 +1,43 @@
...
+    fun setUseContainer(stackArn: String, value: Boolean) {
+        state.updateConfigs.computeIfAbsent(stackArn) { UpdateConfig() }.useContainer = value
+    }
+
+    fun bucketName(stackArn: String): String? = state.updateConfigs[stackArn]?.bucketName
```
kiiadi, line 25: It feel strange to me that you have to pass the `arn` to each of these. To me it makes sense to decouple the persistence from this a bit - so that `getInstance` takes an `arn` and returns a `UpdateConfig` that has getters/setters that talk to the underlying storage. 

So at the call-site you can do something like: 

```kt
val config = UpdateLambdaSettings.getInstance(lambdaArn)
// read values
view.bucket = config.bucketName
// write values
config.bucketName = view.bucket
```

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/UpdateLambdaSettings.kt
hunterwerlla: Sadly no, but will fix the name with the below suggestion as well

jetbrains-core/src/software/aws/toolkits/jetbrains/settings/UpdateLambdaSettings.kt
hunterwerlla: I like this, we should use this style in the other places we have maps like this, changed.

Issue #1951, opened 2020-07-24
Uncaught SDK error in Log Stream actor leads to IDE error


abrooksv:
The following prompted an IDE error:

```
Exception thrown in the LogStreamActor not handled:

software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: Connect to logs.us-west-2.amazonaws.com:443 [logs.us-west-2.amazonaws.com/52.46.216.75] failed: connect timed out
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:98)
	at software.amazon.awssdk.core.exception.SdkClientException.create(SdkClientException.java:43)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper.setLastException(RetryableStageHelper.java:205)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:66)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:34)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:56)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:80)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:60)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:42)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:48)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:31)
[...]

PR #2017, opened 2020-08-25
Fix the CloudWatch Logs table breaking when the service returns an ex…

hunterwerlla:
…ception during loading more (#1951)

## Related Issue(s)
#1951 

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/LogActor.kt
```diff
@@ -46,7 +46,7 @@ sealed class LogActor<T>(
         LOG.error(e) { "Exception thrown in the LogStreamActor not handled:" }
         notifyError(title = message("general.unknown_error"), project = project)
         table.setPaintBusy(false)
-        Disposer.dispose(this)
```
abrooksv, line 49: Is this class registered to a parent Disposable?

hunterwerlla: Yes, but, one of them was not registered, fixed the memory leak :D

Issue #2007, opened 2020-08-21
Quote parameters strings so SAM CLI doesn't truncate them when containing spaces


roneystein:
**Is your feature request related to a problem? Please describe.**
When using parameters values with spaces the SAM Cli truncates then. This is related to #1502 that I thought was fixed with newer SAM CLI version. There's an issue (https://github.com/awslabs/aws-sam-cli/issues/2018) and a PR (https://github.com/awslabs/aws-sam-cli/pull/2139) to address that, however, the value strings will need to be quoted using single quotes.

**Describe the solution you'd like**
Quote parameter values as required by SAM CLI. It still depends on a new SAM CLI release, as of today that is not supported.

**Describe alternatives you've considered**
Running the sam cli command by hand.

**Additional context**
https://github.com/awslabs/aws-sam-cli/pull/2139

abrooksv:
Released in 1.18

PR #2027, opened 2020-08-27
Quote template parameters in cloudformation deploy

abrooksv:
## Types of changes

## Related Issue(s)
#2007 

## Testing
```
Deploying with following values
	===============================
	Stack name                 : HHHH
	Region                     : us-west-2
	Confirm changeset          : False
	Deployment s3 bucket       : fooooooo-test-us-west-2
	Capabilities               : ["CAPABILITY_IAM", "CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND"]
	Parameter overrides        : {'pCron': 'cron(30 0 * * ? *)'}
```

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployDialog.kt
```diff
@@ -96,19 +99,12 @@ class SamDeployDialog(
...
-                .withParameters("--build-dir")
-                .withParameters(buildDir.toString())
-                .apply {
-                    if (useContainer) {
-                        withParameters("--use-container")
-                    }
-                }
-
-            it
+            it.samBuildCommand(
```
kiiadi, line 102: I like this much better

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployDialog.kt
```diff
@@ -209,20 +194,10 @@ class SamDeployDialog(
...
-                is ExecutableInstance.Executable -> it
-                else -> throw RuntimeException((it as? ExecutableInstance.BadExecutable)?.validationError)
-            }
-            return@thenApply samExecutable
-                .getCommandLine()
-                .withWorkDirectory(template.parent.path)
-                .withEnvironment(envVars)
+    private fun createBaseCommand(): CompletionStage<GeneralCommandLine> = ExecutableManager.getInstance().getExecutable<SamExecutable>().thenApply {
+        when (it) {
+            is ExecutableInstance.Executable -> return@thenApply it.getCommandLine()
```
kiiadi, line 199: do you even need the `return` here? I'm assuming it should use the last evaluated expression.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamExecutable.kt
```diff
@@ -56,3 +59,74 @@ class SamExecutable : ExecutableType<SemVer>, AutoResolvable, Validatable {
...
+        addParameter("--capabilities")
+        addParameters(capabilities.map { it.capability })
+    }
+
+    addParameter("--no-execute-changeset")
+
+    if (parameters.isNotEmpty()) {
+        addParameter("--parameter-overrides")
+        parameters.forEach { (key, value) ->
+            addParameter("\"$key\"=\"$value\"")
```
hunterwerlla, line 129: Some additional things I'm concerned about:
What if the value already has a `"`, what if it already has an escaped quote `\"`, what if the value is already quoted, what if the value is already single quoted, what if the value has a single quote in it

justinmk3: Is there validation before this point? I would suggest throwing an error if the key or value contains any of `\`, `"`, or `=` (and probably more).

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployDialog.kt
abrooksv: Cleaned it up, its gone now

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamExecutable.kt
abrooksv: I am not sure about '=', but \" is supported by SAM

https://github.com/wchengru/aws-sam-cli/blob/develop/tests/unit/cli/test_types.py#L65

abrooksv: Okay...I have come to the conclusion that getting this right is impossible due to bugs in SAM:

Expected: `Hello\ World`
| Tried                              | Result        |
|------------------------------------|---------------|
| `"Hello\ World"`    | `Hello World`   |
| `"Hello\\ World"`  | `Hello World`   |
| `"Hello\\\ World"` | `Hello\\ World` |

And `Hello\World` will always become `Hello\\World`

I have also have had to switch to single quotes if already quoted due a bug where they will strip off all surrounding quotes, not just the outside ones. (This requires SAM CLI 1.1.0...)

SAM is happy with `=` it seems:
`ParameterValue=\"Hello=World\" -> `'Hello=World'`

hunterwerlla: Can you make the box validate that? Quoting is way more complicated than it has any right to be

abrooksv: I am not sure we should, as far as a I can tell it is perfectly valid in CFN to do it.

I got the quotes issues all fixed but backslashes are broken now. I plan on cutting an issue to them and seeing what they have to say.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/sam/SamExecutableTest.kt
```diff
@@ -0,0 +1,170 @@
...
+                "--template",
+                "$templatePath",
+                "--build-dir",
+                "$buildDir"
+            ).joinToString(separator = " ")
+        )
+
+        assertThat(cmd.workDirectory).isEqualTo(tempFolder.root)
+
+        assertThat(cmd.environment).containsEntry("Foo", "Bar")
```
hunterwerlla, line 42: Should also check the common environment variables for the credentials

hunterwerlla (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/sam/SamExecutableTest.kt
abrooksv: We dont pass creds at this layer, its already in the GeneralCommandLine before this is invoked.

PR #2031, opened 2020-08-31
Use rds url and port instead of instance id

hunterwerlla:
- Use RDS URL and port instead of instance ID, which maps closer to how the CLI is used to get temporary credentials today
## Screenshots
Kapture 2020-08-31 at 10 38 20

## License


abrooksv:
@sirocode Does the video make sense to you?

RDS Host and Port are what we use to pre-sign the request.

jetbrains-ultimate/src-201+/software/aws/toolkits/jetbrains/services/rds/auth/IamAuthWidget.kt
```diff
@@ -16,48 +15,84 @@ import software.aws.toolkits.jetbrains.ui.AwsAuthWidget
...
 class IamAuthWidget : AwsAuthWidget() {
-    private val instanceIdTextField = JBTextField()
+    private val databaseHostTextField = JBTextField()
+    private val databasePortTextField = JBTextField()
+
+    // Cache the most recently set URL and port, so if we save, and have
+    // emtpy text, we set HOST and PORT properties to the current values instead of
+    // having to split the datasource url string
+    private var cachedUrl: String = ""
+    private var cachedPort: String = ""
```
abrooksv, line 29: I am confused on this, if it was set and cleared out we keep using it old value?

hunterwerlla: Yes, but this is updated on every keystroke, we don't actually need it. We can pull the port and url off of the localdatasource, but I found doing that seemed weirder to me. I think I'll go back to doing that

abrooksv (approved):

PR #2038, opened 2020-09-02
Add purge queue action

hunterwerlla:
- Add purge queue action
- Like the S3 object delete, it's very difficult to test with unit tests, so we will test it with UI tests

## Screenshots
purgeQuery

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
```diff
@@ -0,0 +1,59 @@
...
+class PurgeQueueAction :
+    SingleResourceNodeAction<SqsQueueNode>(message("sqs.purge_queue")),
+    CoroutineScope by ApplicationThreadPoolScope("PurgeQueueAction "),
+    DumbAware {
+    override fun actionPerformed(selected: SqsQueueNode, e: AnActionEvent) {
+        val project = selected.nodeProject
+        val client: SqsClient = project.awsClient()
+        launch {
+            try {
+                client.purgeQueue { it.queueUrl(selected.queueUrl) }
```
abrooksv, line 33: Should we put the table into busy mode and say purge is in progress? That way they can't alter anything while it is purging

abrooksv: Like what happens when you poll while it is purging?

hunterwerlla: The SQS API is weird, it's valid to poll during a purge and you might still get results.
It's also valid to send during a purge, which might or might not actually put it in the queue. 
Delete is also weird in that a delete might not actually delete.

So, I don't know if we actually want to do anything else.

abrooksv: welp

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
```diff
@@ -0,0 +1,59 @@
...
+import software.aws.toolkits.core.utils.warn
+import software.aws.toolkits.jetbrains.core.awsClient
+import software.aws.toolkits.jetbrains.core.explorer.actions.SingleResourceNodeAction
+import software.aws.toolkits.jetbrains.services.sqs.SqsQueueNode
+import software.aws.toolkits.jetbrains.utils.ApplicationThreadPoolScope
+import software.aws.toolkits.jetbrains.utils.notifyError
+import software.aws.toolkits.jetbrains.utils.notifyInfo
+import software.aws.toolkits.resources.message
+
+class PurgeQueueAction :
```
abrooksv, line 24: Add test please

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
hunterwerlla: see description `Like the S3 object delete, it's very difficult to test with unit tests, so we will test it with UI tests` all we can only really test that the API is called, and cannot test that it fails. If we wrap it, it's prettymuch just testing the API only, so it only makes sense as a UI test.

abrooksv: gotcha

PR #2045, opened 2020-09-03
Reduce number of API calls and stop Cloudwatch Logs query if tab is c…

rli:
…losed

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/QueryActor.kt
```diff
@@ -135,12 +158,19 @@ class QueryResultsActor(
...
-        while (response.results().size == 0) {
-            request = GetQueryResultsRequest.builder().queryId(queryId).build()
-            response = client.getQueryResults(request)
+        var response: GetQueryResultsResponse
+        while (true) {
+            response = client.getQueryResults {
+                it.queryId(queryId)
+            }
+
+            if (response.results().isNotEmpty() || response.status() == QueryStatus.COMPLETE) {
```
abrooksv, line 167: Can you clarify the API? Does empty really mean we are done? Or does it just mean that the chunk we just analyzed did not match anything and we need to keep going until QueryStatus is COMPLETE?

rli: the latter; the first GetQueryResults call is not guaranteed to return anything, so loop until we get a something we can show to the user, or the query is completed.

abrooksv: Bah, no caffeine this morning. I read it as `response.results().isEmpty()`

abrooksv (approved):

hunterwerlla (approved):

PR #2051, opened 2020-09-08
Add UI tests for SQS

hunterwerlla:
## License


ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/SQSTest.kt
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+            closeToolWindowTab()
+            step("edit queue settings") {
+                awsExplorer {
+                    openExplorerActionMenu(sqsNodeLabel, queueName)
+                }
+                findAndClick("//div[@text='$editQueueAttributesAction']")
+                // Make sure the dialog opens, then close it
+                find<NewProjectWizardDialog>(DialogFixture.byTitle(editQueueAttributesTitle)).close()
```
kiiadi, line 137: So we don't actually mutate any of the attributes?

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/SQSTest.kt
```diff
@@ -0,0 +1,261 @@
...
+                pressYes()
+                val toast = findToast()
+                assertThat(toast.hasText { it.text.contains("Started purging queue") })
+                awsExplorer {
+                    openExplorerActionMenu(sqsNodeLabel, queueName)
+                }
+                findAndClick("//div[@text='$purgeQueueText']")
+                pressYes()
+                val errorToast = findToast()
+                assertThat(errorToast.hasText { it.text.contains("Purge queue request already in progress for queue") })
```
kiiadi, line 153: This feels like it might be a bit fragile and lead to a timing issue ...

kiiadi (approved):

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/SQSTest.kt
hunterwerlla: I can add that, the reason I did not is because it's a big pain since they are all the same type of element (so they aren't easy to distinguish). I might go with a fill all method, let me see what works.

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/SQSTest.kt
hunterwerlla: It's a 60 second period, so I don't think it will be that fragile. I like having it, because this is the only way to test it in a reasonable way.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SqsWindow.kt
```diff
@@ -39,8 +39,18 @@ class SqsWindow(private val project: Project) : CoroutineScope by ApplicationThr
     private fun showQueue(queue: Queue, component: SqsWindowUi) = launch {
         SqsTelemetry.openQueue(project, queue.telemetryType())
         try {
+            val existingWindow = toolWindow.find(queue.queueUrl)
+            if (existingWindow != null) {
+                withContext(edtContext) {
```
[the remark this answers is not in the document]
hunterwerlla: No this is not exactly the same. This would be like if we had a button to open `output` directly, which we do not.

PR #2052, opened 2020-09-09
SQS Copy action and small cleanup

hunterwerlla:
- Add copy action
- Truncate in the renderer instead of in the column model, so we can copy the whole message if it is too long
- Small code cleanup - rename `toggleableWrap` to `wrapOnToggle`, replace a `runInEdt`

## License


rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/CopyMessageAction.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.ide.CopyPasteManager
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.ui.table.TableView
+import software.amazon.awssdk.services.sqs.model.Message
+import software.aws.toolkits.resources.message
+import java.awt.datatransfer.StringSelection
+
+class CopyMessageAction(private val table: TableView<Message>) : DumbAwareAction(message("sqs.copy.message", 1), null, AllIcons.Actions.Copy) {
+    override fun update(e: AnActionEvent) {
```
kiiadi, line 16: should we tie these actions to the standard shortcuts? So if they hit `cmd/ctrl + c` it will copy?

hunterwerlla: Yes, let me find the other place we do that

kiiadi (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SendMessagePane.kt
```diff
@@ -68,28 +68,32 @@ class SendMessagePane(
...
+                    it.queueUrl(queue.queueUrl)
+                    it.messageBody(inputText.text)
+                    if (queue.isFifo) {
+                        it.messageDeduplicationId(fifoFields.deduplicationId.text)
+                        it.messageGroupId(fifoFields.groupId.text)
+                    }
+                }.messageId()
+                messageSentLabel.text = message("sqs.send.message.success", messageId)
+            }
+            clear(isSend = true)
```
kiiadi, line 86: When i was playing with it this afternoon - this caused the UI to freeze. I had to forcibly put clear on the UI thread.

It's concerning that this kind of issue can creep in in coroutine land; I don't remember seeing a full UI lock-up in bugs we've had before.

kiiadi (changes requested):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/toolwindow/SendMessagePane.kt
hunterwerlla: That is very strange, I'll put that into the base PR of this PR

PR #2053, opened 2020-09-10
Move 'sent message' to the same line as buttons rather than mutating …

kiiadi:
Screen Shot 2020-09-09 at 11 06 53 PM


kiiadi:
> I was going to do this but the group and deduplication IDs spring to the left when this happens which looks a bit weird. Does it still do that?

Yes - it still does that - I can take a look see if we can "pre-compress" it so it doesn't look so odd.

kiiadi:
I think the only way to do it would be to set absolute sizes - and we don't really want to do that - I'd prefer to have things resize as they need to.

Which means we're left with two choices:
1) Always spring vertically with a bunch of wasted space (esp. in non-fifo queues) eg:

2) Spring horizontally in narrow windows, only on fifo queues eg:


Personally I think option 2 is better, the weird jumping occurs in fewer scenarios (e.g. only fifo queues) and overall wastes less UI space.

hunterwerlla:
> I think the only way to do it would be to set absolute sizes - and we don't really want to do that - I'd prefer to have things resize as they need to.
> 
> Which means we're left with two choices:
> 
>     1. Always spring vertically with a bunch of wasted space (esp. in non-fifo queues) eg:
>        
> 
>     2. Spring horizontally in narrow windows, only on fifo queues eg:
>        
> 
> 
> Personally I think option 2 is better, the weird jumping occurs in fewer scenarios (e.g. only fifo queues) and overall wastes less UI space.

Agree, thanks for looking into this

hunterwerlla (approved):

PR #2057, opened 2020-09-11
SQS UI Cleanup 3

hunterwerlla:
(Hopefully the final SQS UI/UX cleanup)

- Remove delete message. Delete is confusing since it does not always work. It also allows us to
- Set poll visibility timeout to 0. This means that the poll is actually a peek, which is nicer as a user.
- Add a confirmation to `Purge Queue` along with an approximate number of messages
- Make sure messages that open dialogs have `...` after them
- Add the purge queue action into the table
- Update "Poll Queue" to "View Messages" along with the help text

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
```diff
@@ -24,42 +25,53 @@ import software.aws.toolkits.resources.message
...
-        val client: SqsClient = project.awsClient()
+class PurgeQueueAction(
+    private val project: Project,
+    private val client: SqsClient,
+    private val queue: Queue
+) : DumbAwareAction(message("sqs.purge_queue.action")), CoroutineScope by ApplicationThreadPoolScope("PurgeQueueAction") {
+    override fun actionPerformed(e: AnActionEvent) {
+        val response = Messages.showOkCancelDialog(
+            project,
+            message("sqs.purge_queue.confirm", queue.queueName),
```
kiiadi, line 36: I was actually going to suggest using [`com.intellij.openapi.ui.Messages#showConfirmationDialog`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-api/src/com/intellij/openapi/ui/Messages.java#L387) - but it looks like it's deprecated in favour of `com.intellij.openapi.ui.MessageDialogBuilder`

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -757,11 +757,14 @@ sqs.message.validation.empty.deduplication_id=Enter a deduplication ID.
 sqs.message.validation.empty.group_id=Enter a group ID.
 sqs.message.validation.empty.message.body=Enter a message body
 sqs.message.validation.long.id=ID exceeds the maximum length of 128 characters.
-sqs.message_table_initial_text=Press "Poll for Messages" to load messages
+sqs.message_table_initial_text=Press "Peek Messages" to load messages
```
kiiadi, line 760: Could just call it 'View Messages'

Probably want this to actually refer to the button text too rather than repeating 'Peak/Poll/View' messages in this string.

resources/resources/software/aws/toolkits/resources/localized_messages.properties
```diff
@@ -757,11 +757,14 @@ sqs.message.validation.empty.deduplication_id=Enter a deduplication ID.
...
 sqs.messages.available.text={0, choice, 0#No messages|1#1 message|2#{0,number} messages} available
-sqs.poll.message=Poll for Messages
-sqs.poll.warning.text=<html><p>Once polled, the sample of messages remain in the queue and are not returned to subsequent receive requests for the duration of the visibility timeout.</p></html>
-sqs.purge_queue=Purge Queue
+sqs.poll.message=Peek Messages
+sqs.poll.warning.text=<html><p>Messages are returned to the queue immediately upon peeking.</p></html>
+sqs.purge_queue=Purge Queue...
+sqs.purge_queue.action=Purge Queue
+sqs.purge_queue.confirm=Are you sure you want to purge queue {0}?
```
kiiadi, line 766: Should we add an approximate message count?

"Are you sure you want to purge queue 'myQueue' (approx. 23,234,234 messages)?"

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
```diff
@@ -4,62 +4,90 @@
...
+                        client.purgeQueue { it.queueUrl(queue.queueUrl) }
+                        LOG.info { "Started purging ${queue.queueUrl}" }
+                        notifyInfo(
+                            project = project,
+                            title = message("aws.notification.title"),
+                            content = message("sqs.purge_queue.succeeded", queue.queueUrl)
+                        )
+                        SqsTelemetry.purgeQueue(project, Result.Succeeded, queue.telemetryType())
+                    } catch (e: PurgeQueueInProgressException) {
+                        LOG.warn { "${queue.queueUrl} already has a query purge in progress" }
```
kiiadi, line 71: Nice - good to be intelligent on exceptions that are caught so we can give the customer accurate messaging.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
```diff
@@ -4,62 +4,90 @@
...
+                        notifyError(
+                            project = project,
+                            content = message("sqs.purge_queue.failed.60_seconds", queue.queueUrl)
+                        )
+                        SqsTelemetry.purgeQueue(project, Result.Succeeded, queue.telemetryType())
```
kiiadi, line 76: Is this really succeeded? I'd say it was actually closer to "canceled" but then that's typically a user-action ... not sure. Maybe its fine.

kiiadi (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/PurgeQueueAction.kt
hunterwerlla: If we had the concept of "user caused it to fail" I would mark it as that, but we don't. From our perspective the call succeeded, and we are presenting what the service said back to the user, so I think that it's most useful as succeed.

PR #2077, opened 2020-09-21
Wire up the refresh explorer metric

abrooksv:
## License


hunterwerlla (approved):
Oh, we never did this? good catch

Issue #2073, opened 2020-09-18
Deploying a new CFN stack does not clear the resource cache


abrooksv:
**Describe the bug**
We do not clear the CFN stack cache when doing a deployment of a new stack. This means when we do Deploy again and go to use Update, the stack is not in the combo box.

PR #2078, opened 2020-09-21
Fixup some places where we didn't refresh the explorer on create

hunterwerlla:
- Refresh on Lambda create
- Refresh on CFN deploy (we don't know if it is a create or not, so refresh every time, this is probably better since while actively working with CFN the most updated view would be more useful)
- Remove duplicate `SamTelemetry.deploy`. We need failure in `SamDeployDialog` but not success as it goes on to `DeployServerlessApplicationAction` to finish deploying
## Related Issue(s)
#2073

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/DeployServerlessApplicationAction.kt
```diff
@@ -142,6 +144,8 @@ class DeployServerlessApplicationAction : AnAction(
                     project
                 )
                 SamTelemetry.deploy(project, Result.Succeeded)
+                // We do not know if it is a create or not, so refresh every time
+                project.refreshAwsTree(CloudFormationResources.LIST_STACKS)
```
abrooksv, line 148: Since CFN can touch anything and everything, should this be a full refresh?

hunterwerlla: Yep, this seems like a reasonable place to do that, changed

abrooksv (approved):

PR #2093, opened 2020-09-23
Fix some insights query editor bugs and add more tests

rli:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/QueryEditor.kt
```diff
@@ -143,6 +143,14 @@ class QueryEditor internal constructor(
         return unit.second
     }
 
+    fun setSelectedTimeUnit(unit: ChronoUnit) {
+        for (knownUnits in timeUnits) {
```
hunterwerlla, line 147: nit: seeing a `for` loops is weird, any reason to do this over something more functional like `timeunits.filter { unit == it.second }.forEach{relativeTimeUnit.selectedItem = knownUnits}`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/QueryEditor.kt
```diff
@@ -143,6 +143,14 @@ class QueryEditor internal constructor(
         return unit.second
     }
 
+    fun setSelectedTimeUnit(unit: ChronoUnit) {
```
hunterwerlla, line 146: Also uh what does this do?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/QueryEditor.kt
abrooksv: Can we just use an enum instead? And an EnumComboBoxModel?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/QueryEditor.kt
rli: the item being set needs to be equal to something else that's currently in the model, which is impossible if the model holds a bunch of pairs and we're trying to select a ChronoUnit

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/TimeUnit.kt
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cloudwatch.logs.insights
+
+import software.aws.toolkits.resources.message
+import java.time.temporal.ChronoUnit
+
+enum class TimeUnit(val unit: ChronoUnit, val text: String) {
+    MINUTES(ChronoUnit.MINUTES, message("cloudwatch.logs.time_minutes")),
+
```
hunterwerlla, line 11: nit: the spacing is a bit weird, why are there line returns between each entry?

hunterwerlla (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/insights/TimeUnit.kt
rli: fixed

PR #2099, opened 2020-09-24
Fix Rider project wizard todo when min is 193

abrooksv:
## Testing
Typing the solution name does not modify the project name

## License


jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
```diff
@@ -246,6 +233,7 @@ class DotNetSamProjectGenerator(
     private fun initProjectTextField() {
         projectNameField.text = SAM_HELLO_WORLD_PROJECT_NAME
         projectNameField.isEnabled = false
+        projectNameSetByUser = true
```
rli, line 236: assuming `changeProjectName` actually became `projectNameSetByUser`?

rli (approved):

jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
abrooksv: Seems so, I could not find that flag anywhere in the parent class.

PR #2112, opened 2020-09-30
Cleanup project wizard pt1

abrooksv:
* Remove unneeded custom module
* Remove hard coded sam init name
* Remove unneeded parent classes making a confusing type hierarchy

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2122, opened 2020-10-01
Remove project from ResourceSelector

abrooksv:
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.java
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.java
@@ -8,6 +8,7 @@ import com.intellij.openapi.ui.ValidationInfo;
 import com.intellij.ui.ColoredListCellRenderer;
 import com.intellij.ui.components.JBLabel;
 import com.intellij.uiDesigner.core.GridConstraints;
+import com.intellij.util.text.SemVer;
 import java.awt.Dimension;
 import java.awt.event.ItemEvent;
 import java.util.List;
@@ -20,8 +21,6 @@ import javax.swing.JLabel;
 import javax.swing.JList;
 import javax.swing.JPanel;
 import javax.swing.JTextField;
-import com.intellij.util.text.SemVer;
-import kotlin.Pair;
 import kotlin.Unit;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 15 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -253,23 +255,21 @@ class ResourceSelector<T> private constructor(
 
         fun disableAutomaticSorting(): ResourceBuilderOptions<T> = also { it.sortOnLoad = false }
 
-        fun awsConnection(awsConnection: AwsConnection): ResourceBuilderOptions<T> = awsConnection { awsConnection }
+        fun activeAwsConnection(project: Project): ResourceBuilderOptions<T> = awsConnection { AwsConnectionManager.getInstance(project).connectionSettings() }
```
hunterwerlla, line 258: Would prefer this also be named `awsConnection`. We grab the connection at that point in time, so it doesn't change like might be implied by `active`. Also makes the API make more sense

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.java
```diff
@@ -39,7 +39,7 @@
 import software.aws.toolkits.jetbrains.services.lambda.RuntimeGroupUtil;
 import software.aws.toolkits.jetbrains.services.lambda.sam.SamExecutable;
 
-public class SamInitSelectionPanel implements ValidatablePanel {
```
hunterwerlla, line 42: What did `ValidatablePanel` do to begin with?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -77,11 +74,16 @@ class ResourceSelector<T> private constructor(
             super.setSelectedItem(message("loading_resource.loading"))
 
             val resource = resourceType.invoke()
-            if (resource == null) {
+            val connectionSettings = connectionSettingsSupplier()
```
hunterwerlla, line 77: If we don't have this maybe we should throw? like it will never succeed but it also never reports a failure to the user/implementer

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.java
abrooksv: Nothing, theres no reference to it anywhere but here....

This should have gone in #2112 but got into wrong change set

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
abrooksv: renamed

hunterwerlla (approved):

PR #2128, opened 2020-10-06
Add the type of credentials to validate and setCredentials

abrooksv:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -337,6 +339,13 @@ class ProfileCredentialProviderFactory : CredentialProviderFactory {
         .any { it.propertyExists(SSO_URL) }
 }
 
+private fun Profile.toCredentialType(): CredentialType? = when {
+    this.propertyExists(ProfileProperty.AWS_SESSION_TOKEN) -> CredentialType.StaticProfile
```
rli, line 343: while we're doing this, do we want to differentiate between session-keys (which should be a low number) and static-keys?

abrooksv: Ya, probably. Will need to change it in the upstream

abrooksv: https://github.com/aws/aws-toolkit-common/pull/82

hunterwerlla (approved):

PR #2143, opened 2020-10-15
Fixing Rider clean to delete bin/ obj/ dirs

kiiadi:


abrooksv (approved):

hunterwerlla (approved):
There was a todo because it couldn't resolve them properly, but woo

kiiadi:
> There was a todo because it couldn't resolve them properly, but woo

Seems to be resolving them now ...

PR #2145, opened 2020-10-15
Fix configuring sqs queue as a Lambda trigger

hunterwerlla:
CoroutineUiContext was throwing, revert back to a `runInEdt` with a message not to mess with it
## License


abrooksv:
I don't follow this, was there an exception?

hunterwerlla:
"Wrong modality state"

abrooksv (approved):

PR #2147, opened 2020-10-16
Fix flaky CredentialsRegionHandlerTest tests

abrooksv:
* Make MockRegionProvider#createRegion generate a unique region ID
* Increase the number of IDs we can generate to reduce collision

## License


hunterwerlla (approved):

kiiadi (approved):
NICE

PR #2151, opened 2020-10-18
Adding action for opening a terminal with the current AWS connection

kiiadi:
## Types of changes

## Description

Adding the ability to spawn a new local terminal that has the current AWS connection details injected as environment variables.

## Motivation and Context

Sometimes it's necessary to run CLI commands for AWS tools independent of the toolkit - but it's useful to maintain the toolkits connection context.

## Testing

**TODO: need to write tests**

terminal

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
```diff
@@ -0,0 +1,71 @@
...
+import software.aws.toolkits.jetbrains.core.plugins.pluginIsInstalledAndEnabled
+import software.aws.toolkits.resources.message
+
+class StartAwsTerminal : AnAction(
+    message("aws.terminal.action"),
+    message("aws.terminal.action.tooltip"),
+    TerminalIcons.OpenTerminal_13x13
+) {
+    private val enabled by lazy {
+        pluginIsInstalledAndEnabled("org.jetbrains.plugins.terminal")
```
hunterwerlla, line 29: Is terminal a dynamic plugin yet? If so, or if it ever will be, this needs to not be lazy. On the other hand, update is called a lot, and I'm not sure how cheap the plugin enabled check is.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
```diff
@@ -0,0 +1,71 @@
...
+                val runner = AwsTerminalRunner(project, connection)
+                TerminalView.getInstance(project).createNewSession(runner, TerminalTabState().apply { this.myTabName = connection.shortName })
+            }
+            null -> LOG.error("Unable to start AWS Terminal - connection settings invalid")
+        }
+    }
+
+    private companion object {
+        private val LOG = getLogger<StartAwsTerminal>()
+        private fun validConnection(e: AnActionEvent): ConnectionSettings? {
```
hunterwerlla, line 54: nit: `activeConnection`?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
```diff
@@ -0,0 +1,71 @@
...
+            val connectionManager = AwsConnectionManager.getInstance(project)
+            if (connectionManager.isValidConnectionSettings()) {
+                return connectionManager.connectionSettings()
+            }
+            return null
+        }
+    }
+}
+
+@Suppress("UnstableApiUsage") // TODO move to [LocalTerminalDirectRunner.getInitialCommand] after 2020.3
```
hunterwerlla, line 65: nit: add `FIX_WHEN_MIN_IS_203` for easy searching

hunterwerlla:
I like this iteration of the feature 👍

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
kiiadi: it's actually both - the currently active connection, if it's valid - will rename to `activeValidConnection`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
kiiadi: that's not a nit - that's a solid call out.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsTerminal.kt
abrooksv: It is already dynamic.

Theres a message bus you can listen to to invalid the cache

abrooksv: https://upsource.jetbrains.com/idea-ce/file/idea-ce-4682003011bb42ffdb872d081e79d300bb393d17/platform/platform-impl/src/com/intellij/ide/plugins/DynamicPlugins.kt

ApplicationLevel

abrooksv: You actually dont need these checks at all I think...

It is already gated by the conditional plugin through the plgin.xml

abrooksv: Ah, I see why, its not inside an ActionGroup but instead directly added to the title actions array, interesting...

Wonder if a action group works in the title actions.

kiiadi: I think @rli found a problem where the plugin was installed but wasn't enabled - I wonder what happens in that case -- guess I can test it

core/src/software/aws/toolkits/core/region/AwsRegion.kt
```diff
@@ -35,3 +35,10 @@ data class AwsRegion(val id: String, val name: String, val partitionId: String)
         private fun String.trimPrefixAndRemoveBrackets(prefix: String) = this.removePrefix(prefix).replace("(", "").replace(")", "").trim()
     }
 }
+
+fun AwsRegion.toEnvironmentVariables(existing: MutableMap<String, String>, replace: Boolean = false) {
```
abrooksv, line 39: name this `merge[with|Into]EnvironmentVariables`?

jetbrains-core/src-202/AwsLocalTerminalRunner.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.intellij.openapi.project.Project
+import org.jetbrains.plugins.terminal.LocalTerminalDirectRunner
+import software.amazon.awssdk.auth.credentials.AwsCredentials
+import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.core.region.toEnvironmentVariables
+import software.aws.toolkits.core.utils.getLogger
+import software.aws.toolkits.jetbrains.core.credentials.toEnvironmentVariables
+
+@Suppress("UnstableApiUsage") // TODO Remove after 2020.3 FIX_WHEN_MIN_IS_203
+class AwsLocalTerminalRunner(project: Project, private val region: AwsRegion, private val credentials: AwsCredentials) : LocalTerminalDirectRunner(project) {
```
abrooksv, line 15: Is this the same as 201? Should it go in a 201-202 source set?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -146,14 +149,28 @@ class DefaultCredentialManager : CredentialManager() {
...
 
+private val CREDENTIAL_ENVIRONMENT_VARIABLES = setOf(
+    AWS_ACCESS_KEY_ID.environmentVariable(),
+    AWS_SECRET_ACCESS_KEY.environmentVariable(),
+    AWS_SESSION_TOKEN.environmentVariable()
+)
+
 fun AwsCredentials.toEnvironmentVariables(): Map<String, String> {
     val map = mutableMapOf<String, String>()
-    map["AWS_ACCESS_KEY_ID"] = this.accessKeyId()
```
abrooksv, line 151: much nicer than us hard coding

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsExplorerFactory.kt
```diff
@@ -30,6 +30,8 @@ class AwsExplorerFactory : ToolWindowFactory, DumbAware {
             toolWindow.setTitleActions(
                 actionManager.getAction("aws.settings.refresh"),
                 Separator.create(),
+                actionManager.getAction("aws.toolkit.startTerminal"),
```
abrooksv, line 33: what happens if this is null (due to plugin not enabled)?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
```diff
@@ -0,0 +1,64 @@
...
+import software.aws.toolkits.jetbrains.core.credentials.shortName
+import software.aws.toolkits.jetbrains.core.plugins.pluginIsInstalledAndEnabled
+import software.aws.toolkits.resources.message
+import software.aws.toolkits.telemetry.AwsTelemetry
+import software.aws.toolkits.telemetry.Result
+
+class OpenAwsLocalTerminal : AnAction(message("aws.terminal.action"), message("aws.terminal.action.tooltip"), TerminalIcons.OpenTerminal_13x13), DumbAware {
+
+    override fun update(e: AnActionEvent) {
+        if (!pluginIsInstalledAndEnabled("org.jetbrains.plugins.terminal")) {
```
abrooksv, line 28: Not needed anymore since you gated this action in the plugin.xml

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
```diff
@@ -0,0 +1,64 @@
...
+        e.presentation.isEnabled = e.project?.let { AwsConnectionManager.getInstance(it) }?.isValidConnectionSettings() == true
+    }
+
+    override fun actionPerformed(e: AnActionEvent) {
+        val project = e.getRequiredData(LangDataKeys.PROJECT)
+        when (val state = AwsConnectionManager.getInstance(project).connectionState) {
+            is ConnectionState.ValidConnection -> {
+                val connection = state.connection
+                val credentials = try {
+                    connection.credentials.resolveCredentials()
```
abrooksv, line 42: action performed is EDT by default, so this needs to be moved to background

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
```diff
@@ -146,14 +149,28 @@ class DefaultCredentialManager : CredentialManager() {
...
-        map["AWS_SESSION_TOKEN"] = this.sessionToken()
+        map[AWS_SESSION_TOKEN.environmentVariable()] = this.sessionToken()
     }
 
     return map
 }
+
+fun AwsCredentials.toEnvironmentVariables(existing: MutableMap<String, String>, replace: Boolean = false) {
+    val envVars = toEnvironmentVariables()
+    if (replace || existing.keys.none { it in CREDENTIAL_ENVIRONMENT_VARIABLES }) {
```
abrooksv, line 172: This will fail if you replace session with static access/secret

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/OpenAwsLocalTerminalTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    @CoreTest
+    fun `can open a terminal from explorer`() = uiTest {
+        try {
+            welcomeFrame {
+                openFolder(tempDir)
+            }
+        } catch (e: Exception) {
+            println(e)
```
abrooksv, line 36: Why is this not a failure?

ui-tests/build.gradle.kts
```diff
@@ -21,6 +21,7 @@ plugins {
 dependencies {
     testImplementation(gradleApi())
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
+    testImplementation(project(":resources"))
```
abrooksv, line 24: Dont see why this was added

core/src/software/aws/toolkits/core/region/AwsRegion.kt
kiiadi: I wonder if I should flip it around and have :

`MutableMap<String, String>.put(region: AwsRegion, replace: Boolean = false)`

Might be a little less discoverable is all - but that's effectively what we're doing - we're putting the region into the existing variables.

jetbrains-core/src-202/AwsLocalTerminalRunner.kt
kiiadi: it isn't 201 returns an `Array` and the method is called `getCommand`, 202 returns a `List` and the method is called `getCommands`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
kiiadi: You're worried about the `AWS_SESSION_TOKEN` being left dangling? 
That's why we do `CREDENTIAL_ENVIRONMENT_VARIABLES.forEach { existing.remove(it) }` to make sure that all existing "credential environment variables" get removed from the map.
There's also a test around it see: https://github.com/aws/aws-toolkit-jetbrains/pull/2151/files#diff-d0ef7b3e36ea585764aa7da10d6501228077894277b9a200d9e7529d29afbcbdR53

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsExplorerFactory.kt
kiiadi: that ... is a great question - will test.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
kiiadi: thought @rli came across some weirdness where if it was installed but not enabled it would screw up - but again - I can test.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
kiiadi: good call - what assumptions can we make about : https://github.com/aws/aws-toolkit-jetbrains/blob/master/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CorrectThreadCredentialsProvider.kt ?

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/OpenAwsLocalTerminalTest.kt
kiiadi: ah - lol - it should be.

core/src/software/aws/toolkits/core/region/AwsRegion.kt
abrooksv: /shrug the readability / discoverability may be better but im not sure. Map<String, String> is pretty generic

jetbrains-core/src-202/AwsLocalTerminalRunner.kt
abrooksv: ah thx, missed that

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
abrooksv: conditional include through the `depends` system is guaranteed to only load if the plugin is installed and enabled.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
abrooksv: That is defensive.... if the call is on UI, and leads to an MFA promot it will deadlock the UI >.>

abrooksv: The above should probably be switched to log.error only if wrong to prevent the dead log

core/src/software/aws/toolkits/core/region/AwsRegion.kt
kiiadi: k - will try to come up with a better name

jetbrains-core/src-202/AwsLocalTerminalRunner.kt
kiiadi: trust me - so did I the first time ;)

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/CredentialManager.kt
abrooksv: Welp I'm blind

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
kiiadi: #win

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
kiiadi: > The above should probably be switched to log.error only if wrong to prevent the dead lock

I don't understand what you mean by this.

Moved the other call off onto a pooled thread.

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/OpenAwsLocalTerminalTest.kt
kiiadi: fixed

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
kiiadi: Updated as discussed -it's all now driven out of the plugin.xml

jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
abrooksv: CorrectThreadCredProvider should let the call go through on the UI thread I THINK...it may still deadlock with MFA (due to going into the SDK and back out).

abrooksv: The above is not related to this PR, just an observation because you brought it up made me think about it.

core/tst/software/aws/toolkits/core/region/AwsRegionTest.kt
```diff
@@ -5,40 +5,99 @@ package software.aws.toolkits.core.region
...
+
+            region.mergeWithExistingEnvironmentVariables(env)
+
+            assertThat(env).hasSize(2)
+                .containsEntry("AWS_REGION", region.id)
+                .containsEntry("AWS_DEFAULT_REGION", region.id)
+        }
+
+        @Test
+        fun `toEnvironmentVariables does not replace existing AWS_REGION`() {
```
abrooksv, line 64: the names of these tests dont match the rename to mergeWithExistingEnvironmentVariables

jetbrains-core/resources/META-INF/ext-terminal.xml
```diff
@@ -0,0 +1,15 @@
+<!-- Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+<idea-plugin>
+    <actions>
+        <action id="aws.toolkit.startTerminal" class="software.aws.toolkits.jetbrains.core.terminal.OpenAwsLocalTerminal"/>
```
abrooksv, line 6: FYI, you can inline this into the group since you dont reference it anywhere else

abrooksv (approved):

jetbrains-core/resources/META-INF/ext-terminal.xml
kiiadi: I had it broken out cos I wanted it to show up in the all actions menu - but even nested it shows up - so I've fixed per your suggestion.

core/tst/software/aws/toolkits/core/region/AwsRegionTest.kt
kiiadi: fixed

PR #2156, opened 2020-10-20
Aurora MySQL fixes and improvments

hunterwerlla:
- Fix connecting to Aurora MySQL with IAM credentails
- Give an error message when connecting to Aurora MySQL without SSL
- Stop using `dbName` in error messages as it can be null, use `dbInstanceIdentifier` instead
- Enable SSL by default on the action (Require should work out of the box)
- code cleanup
  - Move the db classes out of the sealed `RdsEngine` becuase it was getting busy
  - Check error messages in IAM auth tests

## screenshots
Screen Shot 2020-10-20 at 10 54 19 AM

## License


jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/rds/resources/RdsResources.kt
```diff
@@ -0,0 +1,32 @@
...
+val LIST_SUPPORTED_INSTANCES: Resource.Cached<List<DBInstance>> = ClientBackedCachedResource(RdsClient::class, "rds.list_supported_instances") {
+    describeDBInstancesPaginator {
+        it.filters(
+            Filter.builder()
+                .name(ENGINE_FILTER)
+                .values(
+                    RdsEngine
+                        .values()
+                        // Filter out AuroraMySql because it is only supported on 202+
+                        .filter { e -> e != AuroraMySql }
```
abrooksv, line 27: Filter not reads better IMO

abrooksv (approved):

jetbrains-ultimate/src-201/software/aws/toolkits/jetbrains/services/rds/resources/RdsResources.kt
hunterwerlla: Good point, changed

PR #2163, opened 2020-10-21
Fixup SQS SNS Topic subscription

hunterwerlla:
- SNS subscription doesn't work out of the box because it doesn't mutate the policy properly. The console does this without notifying the user.
- Check if we have enough permissions, if not, give the user the option to add the policy statement or pick a different queue. Permissions match the console, including the SID
- Due to SQS API limitations, we have to mutate this manually using the JSON document

## Screenshots
Subscribe then subscribe again to show it doesn't add the policy when it already works
Kapture 2020-10-21 at 15 11 10

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsUtils.kt
```diff
@@ -50,3 +53,40 @@ fun SqsClient.approximateNumberOfMessages(queueUrl: String): Int? = try {
...
+    {
+        "Sid": "topic-subscription-$snsArn",
+        "Effect": "Allow",
+        "Principal": {
+            "AWS": "*"
+        },
+        "Action": "SQS:SendMessage",
+        "Resource": "$sqsArn",
+        "Condition": {
+            "ArnLike": {
```
kiiadi, line 74: I assume this just matches the console - but can't this be `ArnEquals` ?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SubscribeSnsDialog.kt
```diff
@@ -87,6 +110,32 @@ class SubscribeSnsDialog(
         }
     }
 
+    private fun topicSelected(): String = view.topicSelector.selected()?.topicArn() ?: ""
+
+    private fun needToEditPolicy(existingPolicy: String?): Boolean {
+        existingPolicy ?: return true
+
+        // This does not actually make API calls
```
kiiadi, line 118: What do you mean by this does not actually make API calls? The call to `simulateCustomPolicy` is an API call to IAM isn't it?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SubscribeSnsDialog.kt
```diff
@@ -87,6 +110,32 @@ class SubscribeSnsDialog(
...
+                    .contextKeyName("aws:SourceArn")
+                    .contextKeyValues(topicSelected())
+                    .build()
+            )
+            it.actionNames("sqs:SendMessage")
+            it.resourceArns(queue.arn)
+            it.policyInputList(existingPolicy)
+        }.evaluationResults().first()
+
+        if (allowed.evalDecision() != PolicyEvaluationDecisionType.ALLOWED) {
```
kiiadi, line 132: could be simplified to `return allowed.evalDecision() != PolicyEvaluationDecisionType.ALLOWED`

kiiadi (commented):
What testing are you planning for this?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SubscribeSnsDialog.kt
hunterwerlla: nope:
```
 Simulate how a set of IAM policies and optionally a resource-based policy works with a list of API operations and AWS resources to determine the policies' effective permissions. The policies are provided as strings.

The simulation does not perform the API operations; it only checks the authorization to determine if the simulated policies allow or deny the operations. 
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SqsUtils.kt
hunterwerlla: This matches the console, it could be `ArnEquals`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SubscribeSnsDialog.kt
hunterwerlla: True, fixed

hunterwerlla:
Added a UI test (Although they don't run in CI for now since the feature is disabled)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/SubscribeSnsDialog.kt
kiiadi: yeah but you're still calling out over the wire to AWS - it's still a network call. When they say does not perform API calls I'm assuming it means that they don't call other AWS APIs to run the simulation.

hunterwerlla: Ah yeah, looking at the implementation that is the case. You can't be sure without looking at the implementation since everything is modeled the same way, even things that actually don't make any api calls

kiiadi (approved):

ui-tests/tst/software/aws/toolkits/jetbrains/uitests/tests/SQSTest.kt
```diff
@@ -50,16 +50,21 @@ class SQSTest {
...
 
     private val queueName = "uitest-${UUID.randomUUID()}"
+    private val snsTopicName = "uitest-${UUID.randomUUID()}"
+    private var snsTopicArn = ""
     private val fifoQueueName = "fifouitest-${UUID.randomUUID()}.fifo"
 
     @Test
     @CoreTest
     fun testSqs() = uiTest {
+        val snsClient = SnsClient.create()
```
kiiadi, line 65: rather than having this whole thing commented out - can we wrap it with an "assume" or something? I'm worried we're going to forget to uncomment it when we got to turn this on.

hunterwerlla: Talked offline, we can't really do this well today, we will revisit a fix for this issue.

PR #2186, opened 2020-11-06
Significantly speed up loading the list of S3 buckets

hunterwerlla:
- By using coroutines to parallelize headbucket
- It is safe to have 100 calls of headbucket in parallel (we do it in VSCode already)

## Related issues
#2174

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/resources/S3Resources.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/resources/S3Resources.kt
@@ -3,6 +3,10 @@
 package software.aws.toolkits.jetbrains.services.s3.resources
 
 import com.intellij.openapi.project.Project
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/resources/S3Resources.kt
```diff
@@ -26,21 +31,26 @@ object S3Resources {
...
+                        }
+                    }
+                }.awaitAll().filterNotNull()
             }
         }
     }
 
     val LIST_BUCKETS: Resource<List<Bucket>> = LIST_REGIONALIZED_BUCKETS.map { it.bucket }
 
-    fun bucketRegion(bucketName: String): Resource<AwsRegion?> = Resource.View(LIST_REGIONALIZED_BUCKETS) {
```
kiiadi, line 40: what this not used?

hunterwerlla: Yep it was unused

kiiadi (approved):

PR #2228, opened 2020-11-24
Remove "open stream" button and instead show breadcrumb on insights record

kiiadi:
Screen Shot 2020-11-23 at 4 24 32 PM


hunterwerlla (approved):
Nice UX improvment

PR #2237, opened 2020-11-30
Add more diagnostics for flaky telemetry service test

hunterwerlla:
- metricEventMetadataIsSet seems to be flaky sometimes, but after 2000 runs on my local machine it did not fail once.
- Looking into how it failed it was unable to find the metric type which we set, so it seems like it was because credentials were not set properly
  - Added an assert and a verbose throw that will tell us if it is an issue with it being not stabilized when it should be or if there is an issue with it being set at all
  - The verbose throw will also allow us to see if there is something else wrong instead

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/telemetry/TelemetryServiceTest.kt
```diff
@@ -190,10 +194,13 @@ class TelemetryServiceTest {
     }
 
     private fun assertMetricEventsContains(events: Collection<MetricEvent>, event: String, awsAccount: String, awsRegion: String) {
-        val metricEvent = events.find { e ->
+        events.find { e ->
```
abrooksv, line 197: Doing 
```
    assertThat(events).anySatisfy {
            assertThat(it.data)...
        }
```

should print the whole list now

abrooksv:
```
        assertThat(events).anySatisfy {
            assertThat(it.data).extracting<String>(MetricEvent.Datum::name).contains(event)
            assertThat(it.awsAccount).describedAs("awsAccount").isEqualTo(awsAccount)
            assertThat(it.awsRegion).isEqualTo(awsRegion)
        }
```

yields:

```
java.lang.AssertionError: 
Expecting any element of:
  <[software.aws.toolkits.core.telemetry.DefaultMetricEvent@7fa760a1]>
to satisfy the given assertions requirements but none did:

  <software.aws.toolkits.core.telemetry.DefaultMetricEvent@7fa760a1> error: 
Expecting:
 <["Foo", "Faz", "Fuz", "Barr", "F"]>
to contain:
 <["Bar"]>
but could not find:
 <["Bar"]>
```

so we would need to implement the toStrings()s for Datum and MetricEvent, (which we prob should do...)

hunterwerlla: That looks like a better way to do it, more clear what it does. Changed

abrooksv (approved):

PR #2238, opened 2020-11-30
Do not fail build on empty changelog due to it breaks pipeline promot…

abrooksv:
…ions

## License


buildSrc/src/software/aws/toolkits/gradle/changelog/ReleaseCreator.kt
```diff
@@ -11,9 +11,6 @@ class ReleaseCreator(private val unreleasedFiles: Collection<File>, private val
         if (nextReleaseFile.exists()) {
             throw RuntimeException("Release file $nextReleaseFile already exists!")
         }
-        if (unreleasedFiles.isEmpty()) {
```
rli, line 14: probably worth at least a log message?

rli (approved):

buildSrc/src/software/aws/toolkits/gradle/changelog/ReleaseCreator.kt
abrooksv: I can get onboard with that

PR #2239, opened 2020-11-30
Fix resource selector when there is an invalid item selected

hunterwerlla:
- Previously, when an invalid item was selected as the previously selected item, it would select the literal string "loading..." which breaks prettymuch everything
- Check for this and set selected to null if we encounter this scenario. Since we have different types of input and marchers, it's easier to check after the fact vs checking if the previous value is in the model without breaking existing configuration.

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
```diff
@@ -163,7 +166,16 @@ class ResourceSelector<T> private constructor(
...
                 value.isEmpty() -> super.setSelectedItem(null)
                 value.size == 1 -> super.setSelectedItem(value.first())
-                else -> super.setSelectedItem(determineSelection(selector, previouslySelected))
+                else -> {
+                    // if the select item was removed or set to an invalid value, setSelectedItem
+                    // will cause the selected item to be the string "Loading..." so we need to
+                    // make sure after setting that that is not the case. Otherwise, set it to
+                    // null to inform the user they need to set it again
+                    super.setSelectedItem(determineSelection(selector, previouslySelected))
+                    if (selectedItem == loading) {
```
abrooksv, line 175: ugh....we cant easily fix this in the correct way until the parent class of this is no longer ComboBox since we cant make Selector.Left not be Any...

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/ResourceSelector.kt
rli: maybe it's because i haven't looked at kotlin in a while, but i'm slightly annoyed by the fact that this is actually `getSelectedItem`, but we also declared a `fun selectedItem(matcher: (T) -> Boolean) {` in the same file

hunterwerlla: nah it's because this is not great, we need to fixup the implementation.

abrooksv: it is inheriting stuff from ComboBox Richard making this naming convention not make sense

PR #2240, opened 2020-11-30
Fix flaky backOffTimeIsRespected test

abrooksv:
* Compare the durations instead of just the seconds due to truncation

## License


core/tst/software/aws/toolkits/core/credentials/sso/SsoAccessTokenProviderTest.kt
```diff
@@ -215,7 +215,7 @@ class SsoAccessTokenProviderTest {
         setupCacheStub(expirationClientRegistration)
 
         ssoOidcClient.stub {
-            stubStartDeviceAuthorization(interval = 1)
```
abrooksv, line 218: This doesn't do anything since the interval time is overriden by the SlowDownException which spec says to be 5s

kiiadi (approved):

hunterwerlla (approved):

PR #2244, opened 2020-12-01
Enable image lambda tests now that it is out

hunterwerlla:
- Opening as draft, once sam cli 1.13 is out we can iterate on this to make sure everything works

## License


kiiadi (approved):

hunterwerlla:
merging this as it is only failing the integration tests that are already failing on master (with the sam cli 1.13.1 update). Will be investigating that next but would like them to be separate

PR #2245, opened 2020-12-01
Add ability to run a query from log group window

kiiadi:
Screen Shot 2020-12-01 at 2 10 18 PM

Also remove "open query editor" button and move it to a toolbar on the right so its more consistent with other toolwindows
Screen Shot 2020-12-01 at 2 10 47 PM


hunterwerlla (approved):

abrooksv (approved):

jetbrains-core/resources/telemetryOverride.json
```diff
@@ -1,5 +1,11 @@
 {
-    "types": [],
+    "types": [
+        {
```
rli, line 3: probably fix and move into common

kiiadi: https://github.com/aws/aws-toolkit-common/pull/198

kiiadi: done

PR #2251, opened 2020-12-02
Fixup Lambda configuration

hunterwerlla:
- Update sdk to prod sdk
- Adjust max Lambda memory to the new max 10240
  - Note on that for the slider. Since 10240-128 = 10112 and memory increments are 64, 10112/64 = 158, there are 158 increments. However, 158 only factors into 2 and 79, so, the slider will always look wonky.

## screenshot
Screen Shot 2020-12-02 at 9 16 34 AM

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/upload/LambdaConfigPanel.form
```diff
@@ -146,14 +146,17 @@
         <constraints>
           <grid row="0" column="1" row-span="1" col-span="2" vsize-policy="3" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
         </constraints>
-        <properties/>
+        <properties>
```
abrooksv, line 149: This is also used in CreateFunction, fyi. Make sure you didnt break that

hunterwerlla: Removing this change from this PR since reindexing is taking > 30 min, we can tackle this in a different PR

abrooksv (approved):

kiiadi:
> Note on that for the slider. Since 10240-128 = 10112 and memory increments are 64, 10112/64 = 158, there are 158 increments. However, 158 only factors into 2 and 79, so, the slider will always look wonky.

I think ultimately we probably want some sort of logarithmic scale (this would mean a custom extension to the `JSlider`) but to me a logical set would be doubling each time:

128
256
512
1024
2048
4096
8172

hunterwerlla:
> > Note on that for the slider. Since 10240-128 = 10112 and memory increments are 64, 10112/64 = 158, there are 158 increments. However, 158 only factors into 2 and 79, so, the slider will always look wonky.
> 
> I think ultimately we probably want some sort of logarithmic scale (this would mean a custom extension to the `JSlider`) but to me a logical set would be doubling each time:
> 
> 128
> 256
> 512
> 1024
> 2048
> 4096
> 8172

Yeah there are ways we can handle this better, but we need to make our own construct

PR #2260, opened 2020-12-03
Standardize deploy on the same step system as everything else

hunterwerlla:
- Standardize deploy on the same step system as update code/cloud debug
- Add image based deploy test

## Screenshots
Screen Shot 2020-12-03 at 1 10 19 PM
Screen Shot 2020-12-03 at 1 09 35 PM

## License


kiiadi:
how does this work when I want to confirm before i deploy?

hunterwerlla:
> how does this work when I want to confirm before i deploy?

See screenshot 2

jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/deploy/SamDeployTest.kt
```diff
@@ -41,6 +47,11 @@ class SamDeployTest {
         .region(Region.US_WEST_2)
         .build()
 
+    private val ecrClient = EcrClient.builder()
+        .httpClient(ApacheHttpClient.builder().build())
```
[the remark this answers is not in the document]
hunterwerlla: I don't think we do, the existing one above did this and it was a copy paste, removed from both

kiiadi:
> See screenshot 2

we didn't actually allow them to view the change-set previously right? they had to do that themselves in the console right?

as a future PR - I think we should enhance it, run a 'describe-changeset' and actually show a diff in a table of what's going to happen.

kiiadi (approved):

abrooksv:
Thank you for doing this, happy to see the Step system put to more use and standardizing our code paths.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/DeployServerlessApplicationAction.kt
```diff
@@ -110,46 +118,75 @@ class DeployServerlessApplicationAction : AnAction(
...
-        if (!deployDialog.isOK) return
+        workflow.onSuccess = {
+            runBlocking {
+                val changeSetArn = it.getRequiredAttribute(DeployLambda.CHANGE_SET_ARN)
 
-        val cfnClient = project.awsClient<CloudFormationClient>()
+                if (!stackDialog.autoExecute) {
+                    val response = withContext(edtContext) { UploadFunctionContinueDialog(project, changeSetArn).showAndGet() }
+                    if (!response) {
+                        // TODO telemetry
```
abrooksv, line 144: Did we never have telemetry here?

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/actions/DeployServerlessApplicationAction.kt
hunterwerlla: Will clarify the message and fix telemetry here

PR #2263, opened 2020-12-04
Adding basic filter to stack viewer to more easily see in-progress re…

kiiadi:
filter-stack

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Stack.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Stack.kt
@@ -10,6 +10,8 @@ import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.DefaultActionGroup
 import com.intellij.openapi.actionSystem.LangDataKeys
+import com.intellij.openapi.actionSystem.ToggleAction
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.DumbAwareAction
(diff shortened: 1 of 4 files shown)


abrooksv:
Checkbox icon seems odd, since its a filter

kiiadi:
> Checkbox icon seems odd, since its a filter

I took inspiration from test filtering to show/hide completed steps :
Screen Shot 2020-12-04 at 3 36 38 PM

hunterwerlla:
> > Checkbox icon seems odd, since its a filter
> 
> I took inspiration from test filtering to show/hide completed steps :
> Screen Shot 2020-12-04 at 3 36 38 PM

I was also going to comment this seems a bit weird, but it makes sense with the prior art

kiiadi:
I figure once we have more complex filters - we can change it - but for now it's literally just "show / hide" completed.

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Stack.kt
```diff
@@ -206,9 +209,27 @@ private class StackUI(
...
         })
 
+        actionGroup.addAction(createFilterAction())
+
         return ActionManager.getInstance().createActionToolbar("", actionGroup, false).component
     }
 
+    private fun createFilterAction() =
+        object : ToggleAction(message("cloudformation.stack.filter.show_completed"), null, AllIcons.RunConfigurations.ShowPassed), DumbAware {
+            private val state = AtomicBoolean(true)
```
hunterwerlla, line 219: Why do we need the AtomicBoolean? wouldn't it make more sense to do the update synchronously on the ui thread since it's a ui update?

kiiadi: the update actually happens in the `Updater` as part of the next 'refresh' call (or one is forced if it's not already inflight).

hunterwerlla (approved):

Issue #2208, opened 2020-11-17
Copy S3 URI


noe-mis:
Most of the time, what I want to get is the s3://-uri to an object. Right now I have to copy path and then paste it after typing s3://bucket-name/, which becomes a bit tedious after a while.

I would like there to be an additional option aside from _Copy path_ and _Copy URL_, which is _Copy S3 URI_.

PR #2265, opened 2020-12-07
Add an action to copy the s3 uri and group them into an action group

hunterwerlla:
## Screenshots
Screen Shot 2020-12-07 at 9 51 13 AM

## Related issues 
#2208
## License


abrooksv:
Does this remove Copy URL from the toolboar? (It should be probably given lack of an icon)

hunterwerlla:
> Does this remove Copy URL from the toolboar? (It should be probably given lack of an icon)

I forgot about that, it's actually very wonky, will fix

kiiadi:
can you also do this to the copy actions in the footer?  
Screen Shot 2020-12-07 at 10 56 56 AM

abrooksv (approved):

hunterwerlla:
> can you also do this to the copy actions in the footer?
> Screen Shot 2020-12-07 at 10 56 56 AM

That doesn't actually work unless we want to unwrap from the action group. Having 3 copy actions there would be confusing, so I removed it entirely.

PR #2266, opened 2020-12-07
Migrate off deprecated IconLoader#getIcon

abrooksv:
## License


hunterwerlla (approved):

hunterwerlla (approved):
ktlint failed because `private fun load(path: String): Icon {` is not in expression form

kiiadi:
what's the class reference used for out of interest?

abrooksv:
> what's the class reference used for out of interest?

It is used to get the classloader which is used for the getResource call.

PR #2273, opened 2020-12-08
Fix template selection resetting while changing runtimes in wizard

rli:
## Types of changes

Kapture 2020-12-08 at 11 15 44

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaSamProjectWizard.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaSamProjectWizard.kt
@@ -32,12 +32,12 @@ class JavaSamProjectWizard : SamProjectWizard {
     override fun createSdkSelectionPanel(projectLocation: TextFieldWithBrowseButton?): SdkSelector? = IntelliJSdkSelectionPanel(BuiltInRuntimeGroups.Java)
 
     override fun listTemplates(): Collection<SamProjectTemplate> = listOf(
-        SamHelloWorldMaven(),
         SamHelloWorldGradle(),
-        SamEventBridgeHelloWorldMaven(),
+        SamHelloWorldMaven(),
         SamEventBridgeHelloWorldGradle(),
-        SamEventBridgeStarterAppMaven(),
-        SamEventBridgeStarterAppGradle()
+        SamEventBridgeHelloWorldMaven(),
+        SamEventBridgeStarterAppGradle(),
+        SamEventBridgeStarterAppMaven()
     )
 }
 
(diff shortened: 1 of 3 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamProjectWizard.kt
```diff
@@ -88,6 +88,9 @@ abstract class SamProjectTemplate {
         }
     }
 
+    // defined so that we can restore the template selection when the runtime selection changes
+    override fun equals(other: Any?) = other?.let { this::class.java == it::class.java } ?: super.equals(other)
```
abrooksv, line 92: The equals check feels odd, is it really needed?

We only create 1 SDK template ever and store them in lists, so why is the default identity check not working?

kiiadi:
while we're at it - can we make the java template default to gradle instead of maven?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamProjectWizard.kt
rli: when setting the selected item with `templateComboBox.selectedItem = selectedTemplate`, a equality check runs against everything in the model to figure out which one

rli:
> while we're at it - can we make the java template default to gradle instead of maven?

done

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/java/JavaSamProjectWizard.kt
```diff
@@ -32,12 +32,12 @@ class JavaSamProjectWizard : SamProjectWizard {
...
-        SamHelloWorldMaven(),
         SamHelloWorldGradle(),
-        SamEventBridgeHelloWorldMaven(),
+        SamHelloWorldMaven(),
         SamEventBridgeHelloWorldGradle(),
-        SamEventBridgeStarterAppMaven(),
-        SamEventBridgeStarterAppGradle()
+        SamEventBridgeHelloWorldMaven(),
+        SamEventBridgeStarterAppGradle(),
+        SamEventBridgeStarterAppMaven()
```
kiiadi, line 40: 👍

kiiadi (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamProjectWizard.kt
rli: to be more specific, the specific instances in the model change after a runtime change:

https://github.com/aws/aws-toolkit-jetbrains/blob/a110ba448825fb7b3bff298d3e19bebad1be807d/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.kt#L117-L124

rli: swapped to be keyed off `displayName()` instead and added hashCode() override

abrooksv (approved):

PR #2289, opened 2020-12-15
Fix NodeJsLambdaHandlerResolverTest testing the wrong thing in negati…

abrooksv:
…ve cases

## License


hunterwerlla (approved):
whoops, nice catch

Issue #2224, opened 2020-11-21
UI deadlock issue during SSO


kiiadi:
**Describe the bug**

UI thread-lock completely locks IDE UI thread during SSO authentication flow.

**To reproduce**

1. Create SSO credential
2. Select SSO credential
3. When prompted hit "Open in Browser"
4. Close project (leave IDE running)
5. Re-open project
6. Hit "Start SSO Sign-In" action in AWS Explorer

UI Hangs

Thread dump shows UI hang is occurring here : 
https://github.com/aws/aws-toolkit-jetbrains/blob/master/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/RefreshConnectionAction.kt#L26

Relevant excerpt from stack-dump:

```
"AWT-EventQueue-0" prio=0 tid=0x0 nid=0x0 blocked
     java.lang.Thread.State: BLOCKED
 on java.util.concurrent.ConcurrentHashMap$ReservationNode@6c11a232 owned by "ApplicationImpl pooled thread 5" Id=24
	at java.base@11.0.7/java.util.concurrent.ConcurrentHashMap.clear(ConcurrentHashMap.java:1201)
	at software.aws.toolkits.jetbrains.core.DefaultAwsResourceCache.clear(AwsResourceCache.kt:345)
	at software.aws.toolkits.jetbrains.core.credentials.RefreshConnectionAction.actionPerformed(RefreshConnectionAction.kt:30)
	at software.aws.toolkits.jetbrains.core.explorer.ExplorerToolWindow$createActionLabel$1.hyperlinkActivated(ExplorerToolWindow.kt:141)
	at com.intellij.ui.HyperlinkAdapter.hyperlinkUpdate(HyperlinkAdapter.java:11)
	at com.intellij.ui.HyperlinkLabel.fireHyperlinkEvent(HyperlinkLabel.java:205)
	at com.intellij.ui.HyperlinkLabel.processMouseEvent(HyperlinkLabel.java:128)
	at java.desktop@11.0.7/java.awt.Component.processEvent(Component.java:6415)
	at java.desktop@11.0.7/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop@11.0.7/java.awt.Component.dispatchEventImpl(Component.java:5025)
	at java.desktop@11.0.7/java.awt.Container.dispatchEventImpl(Container.java:2321)
[...]

PR #2293, opened 2020-12-17
Move resource cache fetching onto a Job

rli:
Extension of #2229, #2224 

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -296,7 +298,9 @@ class DefaultAwsResourceCache(
                 val result = cache.compute(context.cacheKey) { _, value ->
                     fetchIfNeeded(context, value as Entry<T>?)
                 } as Entry<T>
-                context.future.complete(result.value)
+                launch {
+                    context.future.complete(result.value.await())
```
hunterwerlla, line 302: if this throws it won't be caught by the catch block anymore?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -291,12 +294,25 @@ class DefaultAwsResourceCache(
...
+            var currentValue: Entry<T>? = null
             try {
                 @Suppress("UNCHECKED_CAST")
                 val result = cache.compute(context.cacheKey) { _, value ->
-                    fetchIfNeeded(context, value as Entry<T>?)
+                    currentValue = value as Entry<T>?
+                    fetchIfNeeded(context, currentValue)
                 } as Entry<T>
-                context.future.complete(result.value)
+                runBlocking {
```
abrooksv, line 304: Why does this need to be runBlocking?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -291,12 +294,25 @@ class DefaultAwsResourceCache(
 
     private fun <T> getCachedResource(context: Context<T>) {
         ApplicationManager.getApplication().executeOnPooledThread {
+            var currentValue: Entry<T>? = null
             try {
                 @Suppress("UNCHECKED_CAST")
                 val result = cache.compute(context.cacheKey) { _, value ->
```
abrooksv, line 300: Why is this compute and not computeIfAbsent?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
rli: it doesn't; there were some issues with bubbling exceptions since it needs to be catch `Throwable` instead of `Exception`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
rli: we could rewrite as `computeIfAbsent`. the logic right now re-inserts the same value if the cached value is still good, otherwise it kicks off a refresh.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
rli: fixed

kiiadi (approved):
wow that was much simpler than I thought - nice.

PR #2307, opened 2020-12-22
Switch to fluxbox from icewm, improve video

abrooksv:
* Use fluxbox WM instead of IceWM
* Change pixel format of ffmpeg to make it work in quicktime
* Make the grouping size match the framerate so we can seek every 1s
* Drop the record UI if statements since we never set it to false

## License

### DIFF
--- a/buildspec/linuxUiTests.yml
+++ b/buildspec/linuxUiTests.yml
@@ -8,9 +8,13 @@ version: 0.2
 env:
   variables:
     CI: true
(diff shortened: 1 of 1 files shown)


abrooksv:
UI test failures are expected, this was not meant to fix those, but ease debugging

shruti0085 (approved):

rli (approved):

Issue #2282, opened 2020-12-11
Add Environment Variables with 'From template' option


AMcManigal:
**Your Environment**

- OS: macOS Catalina
- JetBrains product: PyCharm
- JetBrains product version: 2020.3
- AWS Toolkit version:1.22-203
- SAM CLI version: 1.62
- JVM/Python version: 3.8.6

**Question**

I recently upgraded to the latest version of AWS Toolkit. Now when I use the 'From template' option in the 'Run/Debug Configurations' screen I can't see an option to add environment variables. It does appear if I use 'From handler' instead but then I lose the ability to run a function with layers. 

Is there a way to add environment variables with the 'From template' option? This could be done in AWS Toolkit version 1.16-193.

Thanks.

rdt712:
You can do it in the run config by going to the SAM CLI tab and adding the `--env-vars` or `-n` flags and the absolute path to your `env.json` file in the local invoke args section.

abrooksv:
Sorry for the inconvenience, it was removed as we re-did the UI and it was deemed confusing given the current UX and the actual limitations of SAM:

In handler mode, you could add any env vars you wanted, but in template mode, only env vars defined in the template would actually be overridden and that was not explained to the user leading to confusion.

We can look into fixing that UI to reveal that limitation. @rdt712 is correct on the current work around.

kiiadi:
I also think what might be useful is an "import from remote lambda" feature for both template and handler modes - this would allow us to point at a remote Lambda function and automatically import the environment variables from it. Handy when you have environment variables that are populated by refs' in cfn/cdk.

PR #2319, opened 2021-01-05
Re-add environment variables field to template based lambda run configurations

hunterwerlla:
- Re-add the field to template based lambda run configurations
- When selecting a template-based function, it will now auto-fill the environment variables the function has
- To make this work 100% correctly as SAM expects, we will need a new control with metadata of what the environment variable belongs to since we load and select asynchronously. I started work on this but we can revisit it in the future.

## Related Issue(s)
#2282

https://user-images.githubusercontent.com/11964782/103676240-b1d02000-4f35-11eb-87a3-1aedb3168889.mp4

## License


rli (approved):

abrooksv:
Can we wrap the template and selector in another panel so it doesnt mess with the env var label?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt
```diff
@@ -36,6 +37,25 @@ object SamTemplateUtils {
     private val MAPPER = ObjectMapper(YAMLFactory())
     private const val S3_URI_PREFIX = "s3://"
 
+    fun getFunctionEnvironmentVariables(template: Path, logicalId: String): Map<String, String> = readTemplate(template) {
```
abrooksv, line 40: Add unit tests for this please

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/sam/SamTemplateUtils.kt
```diff
@@ -36,6 +37,25 @@ object SamTemplateUtils {
...
+        val globalVars = try {
+            MAPPER.convertValue<Map<String, String>>(globals)
+        } catch (e: Exception) {
+            null
+        } ?: emptyMap()
+        val vars = try {
+            MAPPER.convertValue<Map<String, String>>(variables)
+        } catch (e: Exception) {
+            null
+        } ?: emptyMap()
```
abrooksv, line 54: Can be simplified to `runCatching { MAPPER.convertValue<Map<String, String>>(variables) }.getOrDefault(emptyMap())`

hunterwerlla (commented):
> Can we wrap the template and selector in another panel so it doesnt mess with the env var label?

Sure,
Screen Shot 2021-01-06 at 9 15 22 AM

abrooksv (approved):

PR #2328, opened 2021-01-08
Make JTreeFixture do things in chunks and wait to cope with async tree modelJtree fixture loading

abrooksv:
Experimental change to see if we can cope with async tree model better

## License


abrooksv:
UI tests seem to be running mostly pretty well without this, so will drop it and revive if we need it

PR #2330, opened 2021-01-08
Use RemoteRobot instead of IdeaFrame to find project structure dialog

rli:
## License


hunterwerlla (approved):
Looks like the test passes now (Cfn failed) NICE

PR #2345, opened 2021-01-14
Split Lambda debugger extension point into image and zip

hunterwerlla:
- Split the extension point so we can have image based runtimes without having zip as well (like in the case of dotnet5). It also simplifies some logic in the debugger extension point
- Based on Austin's old branch, pulled forward to master then made to work

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/ImageDebugSupport.kt
```diff
@@ -0,0 +1,53 @@
...
+import software.aws.toolkits.jetbrains.core.utils.buildList
+import java.util.UUID
+
+interface ImageDebugSupport : SamDebugSupport<ImageTemplateRunSettings> {
+    val id: String
+
+    /**
+     * The primary language id, used to find the correct builder for the runtime
+     */
+    val languageId: String
```
hunterwerlla, line 18: To be able to build properly, it seems like we can either:
1. Store the runtime type itself
2. Store the runtime type id
3. store a language id associated with the runtime
I think storing the language id, while being a bit strange, seems the most reasonable so we don't have a dependency on runtime type. Thoughts?

jetbrains-core/resources/META-INF/ext-java.xml
```diff
@@ -9,7 +9,10 @@
     <extensions defaultExtensionNs="aws.toolkit.lambda">
         <runtimeGroup implementation="software.aws.toolkits.jetbrains.services.lambda.java.JavaRuntimeGroup"/>
         <builder id="JAVA" implementationClass="software.aws.toolkits.jetbrains.services.lambda.java.JavaLambdaBuilder"/>
-        <sam.debugSupport id="JAVA" implementationClass="software.aws.toolkits.jetbrains.services.lambda.java.JavaSamDebugSupport"/>
+        <sam.runtimeDebugSupport id="JAVA" implementationClass="software.aws.toolkits.jetbrains.services.lambda.java.JavaRuntimeDebugSupport"/>
```
rli, line 12: should we split the runtime variants as well?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamDebugSupport.kt
```diff
@@ -60,7 +38,7 @@ interface SamDebugSupport {
         val promise = AsyncPromise<XDebugProcessStarter?>()
         val bgContext = ExpirableExecutor.on(AppExecutorUtil.getAppExecutorService()).expireWith(environment).coroutineDispatchingContext()
 
-        val timerTask = Timer("Debugger Worker launch timer", true).schedule(debuggerAttachTimeoutMs) {
+        val timerTask = Timer("Debugger Worker launch timer", true).schedule(SamDebugger.debuggerConnectTimeoutMs()) {
```
rli, line 41: maybe a TODO, but i thought there was some talk about resetting the timer on process output, like VSC

hunterwerlla: That would require a lot of additional changes, not sure if we actually want to do that but definitely not part of this PR

jetbrains-core/resources/META-INF/ext-java.xml
hunterwerlla: I was thinking that, as well as maybe pushing id into the implementations themselves. Either way it's a bit weird, but explicit registering seems to make the most sense. Let me see if that would make sense to put in this PR.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/ImageDebugSupport.kt
abrooksv: In the JVM world, if we have kotlin and java,

would we use langaugeId Java?

hunterwerlla: As is yes because it's not actually target. I think we need to invert this to put the additional information in our own `Runtime` long term

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/TemplateSettings.kt
```diff
@@ -38,15 +36,14 @@ class TemplateSettings(val project: Project) {
...
-    lateinit var runtime: JComboBox<Runtime>
-        private set
     lateinit var pathMappingsTable: PathMappingsComponent
         private set
     private lateinit var functionModels: DefaultComboBoxModel<Function>
     private lateinit var imageSettingsPanel: JPanel
-    private lateinit var runtimeModel: SortedComboBoxModel<Runtime>
     lateinit var environmentVariables: EnvironmentVariablesTextField
         private set
+    private lateinit var imageDebuggerModel: SortedComboBoxModel<ImageDebugSupport>
```
rli, line 45: is this panel image-only? what happened to the zip runtimes?

hunterwerlla: This panel is image only, for zip runtimes we pull from the template

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
```diff
@@ -106,6 +107,10 @@ class LocalLambdaRunConfiguration(project: Project, factory: ConfigurationFactor
         }
     }
 
+    private fun checkImageDebugger() {
+        imageDebugger() ?: throw RuntimeConfigurationError(message("lambda.image.missing_debugger", rawImageDebugger().toString()))
```
abrooksv, line 111: `rawImageDebugger` returns String, do we need the toString()

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamInvokeRunner.kt
```diff
@@ -168,7 +173,19 @@ class SamInvokeRunner : AsyncProgramRunner<RunnerSettings>() {
...
+            runtime = TelemetryRuntime.from(
+                when (lambdaSettings) {
+                    is ZipSettings -> {
+                        lambdaSettings.runtime.toString()
+                    }
+                    is ImageSettings -> {
+                        lambdaSettings.imageDebugger.id
+                    }
+                    else -> {
+                        ""
```
abrooksv, line 185: Should we sue TelemetryRuntime.Unkown?

abrooksv (approved):
This seems like it is moving the needle in the right direction.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
hunterwerlla: `rawImageDebugger` is nullable so we need the `toString` (or we could do `?: ""` but if a user reports it says `null` we'll know what actually happened)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamInvokeRunner.kt
hunterwerlla: This is a `when` inside of the `from` so passing in the raw value would be equivalent

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
abrooksv: Does null not passed to message() print "null"?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/sam/SamInvokeRunner.kt
abrooksv: True, itll still become Unknown, and i made it not nullable. Fair enough.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunConfiguration.kt
hunterwerlla: Nope, requires a string

PR #2352, opened 2021-01-19
Add dotnet5 debugging

hunterwerlla:
- Add dotnet5 debugging, tested and worked fine in 2020.1
- Add test data for integration tests, but we will need to fix them on codebuild per the existing comment, so that should be a different PR

## License

### DIFF
--- a/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotnetImageDebugSupport.kt
+++ b/jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotnetImageDebugSupport.kt
@@ -54,3 +54,9 @@ class Dotnet31ImageDebug : DotnetImageDebugSupport() {
     override val id: String = Runtime.DOTNETCORE3_1.toString()
     override fun displayName() = Runtime.DOTNETCORE3_1.toString().capitalize()
 }
+
+class Dotnet50ImageDebug : DotnetImageDebugSupport() {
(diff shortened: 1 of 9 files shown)


jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotnetImageDebugSupport.kt
```diff
@@ -54,3 +54,9 @@ class Dotnet31ImageDebug : DotnetImageDebugSupport() {
     override val id: String = Runtime.DOTNETCORE3_1.toString()
     override fun displayName() = Runtime.DOTNETCORE3_1.toString().capitalize()
 }
+
+class Dotnet50ImageDebug : DotnetImageDebugSupport() {
+    // TODO pull the const from LambdaRuntime when the wizard PR merges
+    override val id: String = "dotnet5.0"
+    override fun displayName() = "dotnet5.0".capitalize()
```
rli, line 61: nit: why not stylize as `.NET`?

rli (approved):

jetbrains-rider/src/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotnetImageDebugSupport.kt
hunterwerlla: This matches how we do our other runtime entries (which are based off of the Lambda enum) so keeping with that style

PR #2355, opened 2021-01-20
Fix and Improve the SAM init wizard

hunterwerlla:
- The last PR broke the wizard because validation will always fail since `as? Runtime` no longer works (I missed this because Rider does no validation :( )
- add `minSamInit` and `minSamDebugging` to LambdaRuntime. They can't be `SemVer` since that comes from the intellij sdk and they are in core, so there are extension functions for that
  - With this, make `RuntimeGroup` based off of `LambdaRuntime` which allows us to remove hacks in the wizard for dotnet

## License

### DIFF
--- a/core/src/software/aws/toolkits/core/lambda/LambdaRuntime.kt
+++ b/core/src/software/aws/toolkits/core/lambda/LambdaRuntime.kt
@@ -5,11 +5,16 @@ package software.aws.toolkits.core.lambda
 
 import software.amazon.awssdk.services.lambda.model.Runtime
 
-enum class LambdaRuntime(private val runtime: Runtime?, private val runtimeOverride: String? = null) {
+enum class LambdaRuntime(
+    private val runtime: Runtime?,
+    val minSamInit: String? = null,
+    val minSamDebugging: String? = null,
+    private val runtimeOverride: String? = null
+) {
     NODEJS10_X(Runtime.NODEJS10_X),
     NODEJS12_X(Runtime.NODEJS12_X),
     JAVA8(Runtime.JAVA8),
-    JAVA8_AL2(Runtime.JAVA8_AL2),
+    JAVA8_AL2(Runtime.JAVA8_AL2, minSamDebugging = "1.2.0"),
     JAVA11(Runtime.JAVA11),
     PYTHON2_7(Runtime.PYTHON2_7),
     PYTHON3_6(Runtime.PYTHON3_6),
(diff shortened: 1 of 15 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.kt
```diff
@@ -147,7 +147,7 @@ class SamInitSelectionPanel(
      * Updates UI fragments in the wizard after a combobox update
      */
     private fun wizardUpdate() {
-        val selectedRuntime = runtimeComboBox.selectedItem as? Runtime
+        val selectedRuntime = runtimeComboBox.selectedItem as? LambdaRuntime
```
abrooksv, line 150: If you go through the model (`runtimes`), you dont need the cast

hunterwerlla: Changed

core/src/software/aws/toolkits/core/lambda/LambdaRuntime.kt
```diff
@@ -5,19 +5,24 @@ package software.aws.toolkits.core.lambda
...
+    private val runtime: Runtime?,
+    val minSamInit: String? = null,
+    val minSamDebugging: String? = null,
+    private val runtimeOverride: String? = null
+) {
     NODEJS10_X(Runtime.NODEJS10_X),
     NODEJS12_X(Runtime.NODEJS12_X),
     JAVA8(Runtime.JAVA8),
-    JAVA8_AL2(Runtime.JAVA8_AL2),
+    JAVA8_AL2(Runtime.JAVA8_AL2, minSamDebugging = "1.2.0"),
```
abrooksv, line 17: Our min version is 1.13 right now, we can remove this

core/src/software/aws/toolkits/core/lambda/LambdaRuntime.kt
```diff
@@ -5,19 +5,24 @@ package software.aws.toolkits.core.lambda
...
     JAVA11(Runtime.JAVA11),
     PYTHON2_7(Runtime.PYTHON2_7),
     PYTHON3_6(Runtime.PYTHON3_6),
     PYTHON3_7(Runtime.PYTHON3_7),
     PYTHON3_8(Runtime.PYTHON3_8),
     DOTNETCORE2_1(Runtime.DOTNETCORE2_1),
     DOTNETCORE3_1(Runtime.DOTNETCORE3_1),
-    DOTNET5_0(null, "dotnet5.0");
+    DOTNET5_0(null, minSamInit = "1.16.0", minSamDebugging = "1.13.0", runtimeOverride = "dotnet5.0");
```
rli, line 25: add quick comment? this makes sense but also funny how the versions seem backwards

rli (approved):

core/src/software/aws/toolkits/core/lambda/LambdaRuntime.kt
hunterwerlla: It's actually no longer needed, removed

Issue #2275, opened 2020-12-09
Retry does not re-read credentials file


strowk:
**Describe the bug**
When I update credentials file, plugin does not see new creds.

**To reproduce**
1. prepare profile and credentials in ~/.aws/config and ~/.aws/credentials
1. make creds in ~/.aws/credentials wrong
1. start Intellij Idea (plugin should already be installed)
1. as expected - plugin will show "Invalid AWS Connection"
1. fix creds in file ~/.aws/credentials
1. click Retry
1. still invalid connection
1. restart IDE
1. now it's connected OK

**Expected behavior**
Retry should really retry everything, including reading creds file. No IDE restart should be needed.

**Your Environment**
- OS: Ubuntu 20.04
- JetBrains product: Intellij Idea
- JetBrains product version: 2020.3
- AWS Toolkit version: 1.22-203
- SAM CLI version: --not important--
- JVM/Python version: jdk8

**Additional context**
In real use bug happens because credentials file has token from SSO and it becomes invalid after 12 hours, however toolkit keeps caching it somehow and even "Retry" does not work as expected. Firstly "Retry" must be fixed. Once it is figured out how to reload creds, it ideally should be done automatically on some sort of regular schedule.

abrooksv:
We instruct IntelliJ to watch the credentials file and notify us of changes so we re-parse it but a few people have reported issues with this but we have not been able to reproduce it.

Could you try the following:

Try to edit the credentials file through the IDE by hitting the Edit Credentials File menu item under All Credentials.
<img width="448" alt="Screen Shot 2020-12-09 at 9 46 26 AM" src="https://user-images.githubusercontent.com/8992246/101666733-6d6b6600-3a03-11eb-8c96-bd22d28a4190.png">

Make some changes (even just new lines) and hit Cntrl+S to force a save.

If that works, that indicates it is the ProfileWatcher class not getting updates.

We have added debug logs to this that can be enabled by going to: Help->Diagnostic Tools->Debug Log Settings and adding:

```
software.aws.toolkits.jetbrains.core.credentials.profiles
```

and open the logs through the Help menu as well and look for logs for ProfileWatcher

e.g.
```
ProfileWatcher - Starting profile watcher, profile locations: [...] 
ProfileWatcher - Received events: [...] 
ProfileWatcher - Profile file change detected, scheduling refresh 
```

strowk:
I tried doing following - 

1. break credentials file
[...]

PR #2356, opened 2021-01-20
Emit a credential change event when a profile's source has been modified

rli:
## Types of changes

## Description
Should probably move the diff logic out of `loadProfiles` and be a bit smarter with keeping track of dependencies between profiles, but the naive for-loop only adds on the order of couple ms for a large profile file and SDK calls to STS dominate the clock time for profiles with deep nested chains.

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
@@ -114,6 +114,21 @@ class ProfileCredentialProviderFactory(private val ssoCache: SsoCache = diskCach
             }
         }
 
+        // any profiles with a modified 'source_profile' need to be marked as well
+        newProfiles.validProfiles.forEach { (_, profile) ->
+            val profileId = profile.asId(newProfiles.validProfiles)
+            if (profileId in profilesModified) {
+                // already marked; skip
+                return@forEach
+            }
+            for (source in profile.traverseCredentialChain(newProfiles.validProfiles)) {
+                if (source != profile && source.asId(newProfiles.validProfiles) in profilesModified) {
+                    profilesModified.add(profileId)
+                    break
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


abrooksv:
Related issue #2275

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -114,6 +114,21 @@ class ProfileCredentialProviderFactory(private val ssoCache: SsoCache = diskCach
...
         }
 
+        // any profiles with a modified 'source_profile' need to be marked as well
+        newProfiles.validProfiles.forEach { (_, profile) ->
+            val profileId = profile.asId(newProfiles.validProfiles)
+            if (profileId in profilesModified) {
+                // already marked; skip
+                return@forEach
+            }
+            for (source in profile.traverseCredentialChain(newProfiles.validProfiles)) {
```
abrooksv, line 124: Do we need to reverse the chain?

If you have `profile 1 (root`) -> `profile 2` -> `profile 3 (leaf)`, and `profile 1` is modified, then when we traverse 3->2->1...2 will not be in the profilesModified list yet...leading to only profile 1 and 2 being notified.

abrooksv: Nvm. This code ends up being logically equivalent to `profile.traverseCredentialChain(newProfiles.validProfiles).any {it in profilesModified }`

abrooksv: Can we at least add the above scenario to a unit test? Or just extend `testModifyingASourceProfile` to use 3 profiles instead of 2.

rli: added additional tests

abrooksv (approved):

PR #2359, opened 2021-01-21
Fix validation in Rider wizard

hunterwerlla:
As a follow up to #2355 , fix the issue that caused the PR to be necessary in the first place

Fix Rider wizard to do validation. There were 2 issues:
  1. It did not validate because the call to validate was not wired in
  2. After that was fixed, it still did not validate because changing the runtime/template box would not re-run validation. So, plumb another callback into the wizard to accomplish this

## Screenshot
https://user-images.githubusercontent.com/11964782/105420695-ad725b00-5bf5-11eb-9fb1-777ee491f992.mp4

## License

### DIFF
--- a/jetbrains-rider/src-201-202/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
+++ b/jetbrains-rider/src-201-202/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
@@ -24,12 +24,8 @@ import com.jetbrains.rider.projectView.actions.projectTemplating.impl.ProjectTem
 import com.jetbrains.rider.projectView.actions.projectTemplating.impl.ProjectTemplateTransferableModel
 import com.jetbrains.rider.ui.themes.RiderTheme
 import software.aws.toolkits.core.lambda.LambdaRuntime
-import software.aws.toolkits.jetbrains.core.executables.ExecutableInstance
-import software.aws.toolkits.jetbrains.core.executables.ExecutableManager
-import software.aws.toolkits.jetbrains.core.executables.getExecutableIfPresent
 import software.aws.toolkits.jetbrains.services.lambda.BuiltInRuntimeGroups
 import software.aws.toolkits.jetbrains.services.lambda.RuntimeGroup
-import software.aws.toolkits.jetbrains.services.lambda.sam.SamExecutable
 import software.aws.toolkits.jetbrains.services.lambda.wizard.SamInitSelectionPanel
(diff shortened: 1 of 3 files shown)


jetbrains-rider/src-201-202/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
```diff
@@ -57,10 +53,13 @@ class DotNetSamProjectGenerator(
 
     // TODO: Decouple SamProjectGenerator from the framework wizards so we can re-use its panels
     private val generator = SamProjectGenerator()
-    private val samPanel = SamInitSelectionPanel(generator.wizardFragments) {
+    private val samPanel = SamInitSelectionPanel(
+        generator.wizardFragments,
+        // needed to rerun the validation when the wizard is changed
+        wizardUpdateCallback = { validateData() },
```
rli, line 59: nit: arguments are named but don't match order they were declared in constructor :(

jetbrains-rider/src-203+/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
```diff
@@ -56,9 +52,13 @@ class DotNetSamProjectGenerator(
 
     // TODO: Decouple SamProjectGenerator from the framework wizards so we can re-use its panels
     private val generator = SamProjectGenerator()
-    private val samPanel = SamInitSelectionPanel(generator.wizardFragments) {
-        RuntimeGroup.getById(BuiltInRuntimeGroups.Dotnet).supportedRuntimes.contains(it)
-    }
+    private val samPanel = SamInitSelectionPanel(
+        generator.wizardFragments,
+        // needed to rerun the validation when the wizard is changed
+        wizardUpdateCallback = { validateData() },
```
rli, line 58: ditto

rli (approved):

jetbrains-rider/src-201-202/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
hunterwerlla: That is funky, fixed

jetbrains-rider/src-203+/software/aws/toolkits/jetbrains/services/lambda/dotnet/DotNetSamProjectGenerator.kt
hunterwerlla: fixed as well

PR #2372, opened 2021-01-26
Remove the rest of the GlobalScope

hunterwerlla:
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/DeleteObjectAction.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/DeleteObjectAction.kt
@@ -6,11 +6,11 @@ package software.aws.toolkits.jetbrains.services.s3.objectActions
 import com.intellij.icons.AllIcons
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.Messages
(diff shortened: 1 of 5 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/DeleteObjectAction.kt
```diff
@@ -45,7 +45,7 @@ private fun deleteNodes(project: Project, treeTable: S3TreeTable, nodes: List<S3
     if (response != Messages.OK) {
         S3Telemetry.deleteObject(project, Result.Cancelled)
     } else {
-        GlobalScope.launch {
+        ApplicationThreadPoolScope("DeleteObjectAction").launch {
```
rli, line 48: feels weird that the scope isn't attached to the action

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/objectActions/DeleteObjectAction.kt
hunterwerlla: Yeah so this function is actually called when there is not a copy of the action as well, so it can't just move into the class (the first thing I tried)

abrooksv (approved):

PR #2374, opened 2021-01-27
Make the schema panel lazy in the project wizard

hunterwerlla:
- Make all of the panels lazy so they aren't initialized unless the user goes into the aws wizard
- Make schemas only load once a template that supports them is selected
Any thoughts on a less gross way to do the second part?

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaResourceSelector.kt
```diff
@@ -14,26 +14,31 @@ import software.aws.toolkits.jetbrains.ui.ResourceSelector
...
-    private var currentAwsConnection: ConnectionSettings? = null
+    var awsConnection: ConnectionSettings? = null
+        set(value) {
+            val previous = field
+            field = value
+            // if we are loading for the first time (creation), do not load the schemas selector yet
+            if (previous != null) {
+                schemasSelector.reload()
+            }
+        }
```
abrooksv, line 25: `Delegates.observable` can simplify this

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanel.kt
```diff
@@ -54,7 +64,13 @@ class SchemaSelectionPanel : WizardFragment {
         return null
     }
 
-    override fun isApplicable(template: SamProjectTemplate?): Boolean = template?.supportsDynamicSchemas() == true
+    override fun isApplicable(template: SamProjectTemplate?): Boolean {
+        val supported = template?.supportsDynamicSchemas() == true
+        if (supported && !selectorInitialized.getAndSet(true)) {
+            schemaSelector.reload()
+        }
```
abrooksv, line 71: hmmm...do we need to reload it? The schemas aren't going to change since we aren't changing the creds.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanel.kt
```diff
@@ -17,23 +17,33 @@ import software.aws.toolkits.jetbrains.services.lambda.sam.SamSchemaDownloadPost
...
-        schemaSelector::reloadSchemas
-    )
-    private val component = panel {
-        row {
-            awsConnectionSelector.selectorPanel()(grow)
+    /*
+     * We only want to initialize the selector once it is applicable, so we have to keep track of it
+     */
+    private val selectorInitialized = AtomicBoolean(false)
+    private val schemaSelector by lazy { SchemaResourceSelector() }
```
abrooksv, line 31: can we just load it here?

hunterwerlla: load what here? the selector? We don't want it to load if the template does not support schemas?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaResourceSelector.kt
abrooksv: Actually I dont think we need to make any changes to this class. Changing the components to be lazy in the other file should be enough

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanel.kt
abrooksv: I forgot my own logic -.-

The component method isnt gated by isApplicable as I thought it was.

abrooksv: Can we do the following?

Change SamInitSelectionPanel.kt:

```
        wizardFragments.forEach { (wizardFragment, jComponent) ->
            wizardFragment.updateUi(projectLocation, selectedRuntime?.runtimeGroup, selectedTemplate)
            jComponent.isVisible = wizardFragment.isApplicable(selectedTemplate)
        }
```
to

```
        wizardFragments.forEach { (wizardFragment, jComponent) ->
            val isApplicable = wizardFragment.isApplicable(selectedTemplate)
            if (isApplicable) {
                wizardFragment.updateUi(projectLocation, selectedRuntime?.runtimeGroup, selectedTemplate)
            }
            jComponent.isVisible = isApplicable
        }
```

and then do the `reload` call in the updateUi method?

abrooksv: Also making the vals lazy does not seem to do much given we eagerly load each WizardFragment into the UI and then hide it after

hunterwerlla: Making the vals lazy means they aren't initialized when the wizard is opened, otherwise even if you want to make a java project (as an example) they still get loaded

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaResourceSelector.kt
hunterwerlla: Slightly different, I guess that works. We need to edit this class or it loads the panel when the item is created as we already reload when this is changed

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanel.kt
hunterwerlla: the change in the wizard itself simplifies the logic

abrooksv: Ah, didn't realize the wizard init'ed our stuff even if it wasn't selected.

abrooksv (approved):

PR #2390, opened 2021-01-29
Speed up integration tests by grouping tests together

hunterwerlla:
- Each run of these integration tests duplicated work and took several seconds each * number of runtimes. This leads to minutes of unnecessary running SAM and more potential for stochastic failure.
- Speeds up integration tests by 2-3 minutes

## License


jetbrains-core/it/software/aws/toolkits/jetbrains/services/lambda/python/PythonLocalLambdaRunConfigurationIntegrationTest.kt
```diff
@@ -151,47 +133,12 @@ class PythonLocalLambdaRunConfigurationIntegrationTest(private val runtime: Runt
 
         assertThat(executeLambda.exitCode).isEqualTo(0)
         assertThat(jsonToMap(executeLambda.stdout))
```
abrooksv, line 135: I don't think this is that controversial, but maybe we should use multiple assertThats with descriptions set so we can see better from test reports?

hunterwerlla: Sure, I wanted to see how much this actually sped up the builds first. Will edit it if it works well

abrooksv (approved):

PR #2396, opened 2021-01-29
Index optional BuildMethod metadata for SAM templates and apply Lambd…

rli:
…a line marker

If user is using a custom makefile, there is not enough information to determine if any handler in the project is in the template

## License


hunterwerlla:
A lot of the newer code has moved away from using the index, should we still be adding to it? @abrooksv thoughts ?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/IndexedResources.kt
```diff
@@ -12,28 +12,24 @@ import java.io.DataOutput
...
-                }
-                .mapNotNull { (key, value) -> value?.let { key to it } }
-                .toMap()
+            indexProperties.mapScalars(resource, Resource::getScalarProperty),
+            indexMetadata.mapScalars(resource, Resource::getScalarMetadata)
         )
 
     fun save(dataOutput: DataOutput) {
         dataOutput.writeUTF(type)
         dataOutput.writeInt(indexedProperties.size)
```
hunterwerlla, line 26: I didn't realize we wrote our own serialization for this, rip

hunterwerlla (approved):

PR #2405, opened 2021-02-02
Workaround our test failure of opening too many project structure dialogs

hunterwerlla:
- Since we are still opening too many and randomly failing tests, this is a hack to make it work until we diagnose and fix the root cause
- See build Linux-UI-Tests-2020-2-ui-test:adabec47-a21a-4923-947a-2b1da5ad84e2 as an example of a recent failure

## License


hunterwerlla:
Closing this in favor of a different thing to try

PR #2411, opened 2021-02-03
Try to delete services created by CloudDebug integration tests when deinstrument fails

hunterwerlla:
- On deinstrument failure, the ecs services were not cleaned up properly. This has lead to multiple zombie services in our integration test accounts
- Also simplify the code using waiters

## License


jetbrains-core/it/software/aws/toolkits/jetbrains/services/clouddebug/CloudDebugTestCase.kt
```diff
@@ -87,15 +88,36 @@ abstract class CloudDebugTestCase(private val taskDefName: String) {
 
     @After
     open fun tearDown() {
+        try {
+            deinstrumentService()
+        } finally {
+            // If deinstrumenting fails, or initialization doesn't work properly, we still want to try to delete the services, so kick that off
+            try {
```
abrooksv, line 95: if you dont care about results, you can wrap in `runCatching {}`

abrooksv (approved):

jetbrains-core/it/software/aws/toolkits/jetbrains/services/clouddebug/CloudDebugTestCase.kt
hunterwerlla: I read this as `runCaching` which really confused me. Good point, changed

PR #2452, opened 2021-02-22
Fix showing stack status notification on opening existing CloudFormation stack (#2157)

### LINKED ISSUE #2157: CloudFromation tool window should not show terminal state upon opening
If the cloud formation stack is already in a terminal state, we should not show it in the notification popup upon opening the window

hunterwerlla:
## Related issues
#2157

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Updater.kt
```diff
@@ -96,13 +96,16 @@ class Updater(
         val newStackStatusType = newStackStatus.type
         val newStackStatusNotInProgress = newStackStatusType !in setOf(StatusType.UNKNOWN, StatusType.PROGRESS)
 
-        // Stack changed to some "final" status just now, notify user
-        val stackSwitchedToFinalStatus = previousStackStatusType != newStackStatusType && newStackStatusNotInProgress
+        // Stack changed to some "final" status just now, notify the user. Don't show a notification if the state
+        // happened at creation time, which is distinguished by stackStatus being null
```
rli, line 100: probably want some tests here, but doesn't look that testable :(

rli (approved):

Issue #2166, opened 2020-10-22
Context menu should activate on differently named templates


sdole:
At the moment, I must name my cloudformation file "template.yaml" for intellij context menu to activate the "Deploy Serverless Application" menu item. I should be able to deploy serverless application from any file name as per my project's and company's conventions.

---
Toolkit: AWS Toolkit For JetBrains 1.19-202
OS: Windows 10 10.0
IDE: IntelliJ IDEA Ultimate Edition 2020.2.3

hunterwerlla:
Hello, currently we only check for name because we have no concept of what is a CloudFormation template/a random yaml file. In the long term we are planning on improving that, but would you find a good short term solution to be able to deploy templates from the `CloudFormation` node?

PR #2465, opened 2021-02-26
Allow deploying SAM templates from the CloudFormation node

hunterwerlla:
- Add the `Serverless Deploy` action to the CloudFormation root node
- Opens a file browser to select a template and runs all the same checks that the normal path does. The reason this wasn't refactored is the checking logic is all over the place.

## Related Issue(s)
#2166
## Screenshot

https://user-images.githubusercontent.com/11964782/109360325-d95bae80-783b-11eb-8718-df1916c8b010.mp4

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/deploy/DeployServerlessApplicationPanel.kt
```diff
@@ -0,0 +1,111 @@
...
+import javax.swing.JButton
+import javax.swing.JCheckBox
+import javax.swing.JLabel
+import javax.swing.JPanel
+import javax.swing.JRadioButton
+import javax.swing.JTextField
+
+class DeployServerlessApplicationPanel(private val project: Project) {
+    lateinit var newStackName: JTextField
+    lateinit var createS3BucketButton: JButton
```
abrooksv, line 29: add private sets

hunterwerlla: Added (and was able to set a few more `private`)

abrooksv (approved):

PR #2479, opened 2021-03-05
Inject AWS_PROFILE if the Credential Provider is a Profile credential

kiiadi:
Screen Shot 2021-03-05 at 1 26 41 PM


jetbrains-core/src/software/aws/toolkits/jetbrains/core/terminal/OpenAwsLocalTerminal.kt
```diff
@@ -41,7 +45,13 @@ class OpenAwsLocalTerminal : DumbAwareAction(message("aws.terminal.action"), mes
                         return@executeOnPooledThread
                     }
                     runInEdt {
-                        val runner = AwsLocalTerminalRunner(project, connection.region, credentials)
+                        val runner = AwsLocalTerminalRunner(project) { envs ->
```
abrooksv, line 48: Do we not have tests for this action?

kiiadi: :( no - we probably should.

*sigh* will add.

kiiadi: (well actually that's not true - we have a UI test)

hunterwerlla (approved):

PR #2480, opened 2021-03-05
Retry only on CI

kiiadi:


rli (approved):

build.gradle.kts
```diff
@@ -183,10 +183,12 @@ subprojects {
             html.isEnabled = true
         }
 
-        retry {
-            failOnPassedAfterRetry.set(false)
-            maxFailures.set(5)
-            maxRetries.set(2)
+        if (System.getenv("CI") != null) {
```
hunterwerlla, line 186: There are 3 retry blocks (1 for each test type), please apply it to all 3/maybe make a function (probably not worth it)

kiiadi: oh yeh - ok

build.gradle.kts
```diff
@@ -303,7 +308,6 @@ intellij {
     version = ideProfile.community.sdkVersion
     pluginName = "aws-jetbrains-toolkit"
     updateSinceUntilBuild = false
-    downloadSources = System.getenv("CI") == null
```
hunterwerlla, line 306: We download source on ci now?

hunterwerlla (approved):

abrooksv (approved):

PR #2495, opened 2021-03-12
AwsSdkClient refactor

abrooksv:
* Introduce an interface for it so it can be mocked easier
* Remove the suppliers from the profile factory for the http client

## License


hunterwerlla:
What's the difference between `sharedSdkClient` and `sdkHttpClient`?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/profiles/ProfileCredentialProviderFactory.kt
```diff
@@ -190,7 +190,7 @@ class ProfileCredentialProviderFactory(private val ssoCache: SsoCache = diskCach
 
         // Some profiles failed to load
         if (newProfiles.invalidProfiles.isNotEmpty()) {
-            val message = newProfiles.invalidProfiles.values.joinToString("\n") { it.message ?: it::class.java.name }
```
hunterwerlla, line 193: What does this change?

abrooksv: Whoops, shouldn't be in this PR

abrooksv:
> What's the difference between `sharedSdkClient` and `sdkHttpClient`?

None, function vs val.

Found the shared name a bit clearer that we share it instead of making a new one

hunterwerlla (approved):

PR #2496, opened 2021-03-12
Fix sam init selection panel making 2 sdks

hunterwerlla:
- wizardUpdate was called twice. Once by setting the runtime the first time, then again when we called the wizardUpdate manually.
- This caused an ide fatal error with the Go one
- In general we were making 2 of the panel though which makes 0 sense

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamInitSelectionPanel.kt
```diff
@@ -89,13 +88,13 @@ class SamInitSelectionPanel(
             }
         )
 
+        // this will also fire wizardUpdate since templateComboBox will change
+        // othwise we make 2 of them
```
rli, line 92: nit
```suggestion
        // otherwise we make 2 of them
```

rli (approved):

abrooksv (approved):

PR #2503, opened 2021-03-16
Use CodeArtifact when on CodeBuild

abrooksv:
## License

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -61,6 +61,7 @@ repositories {
 
 allprojects {
     repositories {
+        mavenLocal()
         System.getenv("CODEARTIFACT_URL")?.let {
             println("Using CodeArtifact proxy: $it")
             maven {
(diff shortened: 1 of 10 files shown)


build.gradle.kts
```diff
@@ -66,8 +61,19 @@ repositories {
...
+                url = uri(it)
+                credentials {
+                    username = "aws"
+                    password = System.getenv("CODEARTIFACT_AUTH_TOKEN")
+                }
+            }
+        }
+        gradlePluginPortal()
         mavenCentral()
+        mavenLocal()
```
hunterwerlla, line 76: mevenLocal should go first, right? Or does the order not matter

abrooksv: Why does it need to go first?

abrooksv: Order matters for the order it is checked in, local going first only matters if you need to use a custom / patched version to override what would be in maven central.

hunterwerlla: Yes, which means we should always put it first just in case we need it? Like if we want to test a common package before?

abrooksv: Fair

hunterwerlla (dismissed):

hunterwerlla (approved):

PR #2514, opened 2021-03-19
Fix issues with Go found in the team bug bash

hunterwerlla:
- Run configurations made from gutter icons will now have the runtime. Having an sdk, or any other indication that go is actually supported, is not guaranteed. When go 2.0 is closer we will need to figure out if there's actually a reasonable way to do this
- Make debugger work again after the big refactor
  - Remove the debugger sleep, instead scan the output
- Remove 201 support
- Investigate "Debugger disconnected unexpectedly", we can't fix it on our side, issue filed https://youtrack.jetbrains.com/issue/GO-10775
- Increase the default `aws.sam.goDebuggerDelay`, this will soon be obsoleted by a PR once we move attach debugger to the step system, but in the mean time it's a higher value
## License


jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/go/GoRuntimeGroup.kt
```diff
@@ -22,20 +22,12 @@ class GoRuntimeGroup : SdkBasedRuntimeGroup() {
         LambdaRuntime.GO1_X
     )
 
-    override fun runtimeForSdk(sdk: Sdk): LambdaRuntime? {
-        if (sdk.sdkType !is GoSdkType) {
```
hunterwerlla, line 26: Maybe instead of always being GO_1X we should do this then fall back to GO_1X, thoughts?

abrooksv: I don't follow

abrooksv: Uhhh..... maybe? What are the ramfications

hunterwerlla: Well, it would do nothing different for now. This function is used in 2 paths:
1. picking up runtime when setting up from source file
2. download code from schema
I think we should probably simplify this in the future since it's used in so few places, and isn't really a problem if it's "wrong" but the correct runtime. Until then, I think we are fine just setting `GO1_X` no matter what since it's only called when we already think it's go

abrooksv: Makes sense

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/go/GoHelper.kt
```diff
@@ -45,51 +52,66 @@ fun inferSourceRoot(project: Project, virtualFile: VirtualFile): VirtualFile? {
...
-    return object : XDebugProcessStarter() {
+    // TODO see https://youtrack.jetbrains.com/issue/GO-10775 for "Debugger disconnected unexpectedly" when the lambda finishes
+    suspend fun createGoDebugProcess(
+        debugHost: String,
+        debugPorts: List<Int>,
+        context: Context
+    ): XDebugProcessStarter = object : XDebugProcessStarter() {
         override fun start(session: XDebugSession): XDebugProcess {
-            val process = createDelveDebugProcess(session, executionResult)
+            val process = DlvDebugProcess(session, DlvRemoteVmConnection(DlvDisconnectOption.KILL), null, true)
```
rli, line 67: any ramifications due to a null ExecutionResult?

rli (approved):

abrooksv (approved):

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/go/GoHelper.kt
hunterwerlla: Yep, it doesn't try to execute the state in a loop. it's marked `@nullable` so this is the correct usage

Issue #2519, opened 2021-03-22
Cloudformation stack events not updated after reaching terminal state


manodnyab:
**Describe the bug**

Cloudformation stack events not updated after a final/complete stage has been reached.

**To reproduce**

1) Open the stack window.
2) Delete the stack.
3) Observe the stack events

**Expected behavior**

The event status should be updated.

hunterwerlla:
We already have an issue for this: https://github.com/aws/aws-toolkit-jetbrains/issues/2180

PR #2522, opened 2021-03-23
Update stack events periodically even after reaching terminal/final state

manodnyab:
## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Description

This change updates stack events in the stack window periodically and stops polling once the stack status is DELETE_COMPLETE

## Related Issue(s)

Cloudformation stack events not updated after reaching terminal state #2519

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Updater.kt
```diff
@@ -134,8 +135,14 @@ class Updater(
             updating.set(!newStackStatusNotInProgress)
 
             // Reschedule next run
-            if (!alarm.isDisposed && updating.get() && alarm.isEmpty) {
-                alarm.addRequest({ fetchDataSafely() }, updateInterval.toMillis())
+            if (stackStatus == StatusType.DELETED) {
```
hunterwerlla, line 138: If I press refresh and the stack has changed state, it should start updating at the new update rate (if changed)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Updater.kt
```diff
@@ -48,7 +49,7 @@ class Updater(
     private var predicate: (StackResource) -> Boolean = { true }
     private val updating = AtomicBoolean(false)
     val running get() = updating.get()
-    private val eventsFetcher = EventsFetcher(stackName)
+    private val eventsFetcher = EventsFetcher(stackId)
```
hunterwerlla, line 52: Why did we move from stackName to stackId?

manodnyab: The DescribeStack and DescribeStackResources calls require the unique ID when called on deleted stacks. Since, ID can be used for regular calls as well, ID may be a better option than changing the parameters based on the status of the stacks

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudformation/stack/Updater.kt
manodnyab: Added changes for after refresh polling with new update rate in the next commit

hunterwerlla:
Now that we are polling deleted stacks again, I think it might have the same old behavior of a bunch of errors when it's deleted (after a minute), can you confirm it doesn't do that?

manodnyab:
> Now that we are polling deleted stacks again, I think it might have the same old behavior of a bunch of errors when it's deleted (after a minute), can you confirm it doesn't do that?

It doesn't raise any errors. If the stack is open, after deletion, the periodic polling is halted.

hunterwerlla (approved):

PR #2532, opened 2021-04-02
Fix flaky ProfileWatcherTest.kt

hunterwerlla:
- On my machine this test would fail about 1/100 runs, but it has been failing a lot more often on CI recently.

## License

### DIFF
--- a/buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
+++ b/buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
@@ -131,7 +131,7 @@ afterEvaluate {
 
         systemProperty("aws.telemetry.skip_prompt", "true")
         systemProperty("aws.suppress_deprecation_prompt", true)
-        ciOnly {
+        ciOnly() {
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
```diff
@@ -131,7 +131,7 @@ afterEvaluate {
 
         systemProperty("aws.telemetry.skip_prompt", "true")
         systemProperty("aws.suppress_deprecation_prompt", true)
-        ciOnly {
+        ciOnly() {
```
rli, line 134: why this change?

rli (approved):

buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
hunterwerlla: IDE suggested it and I had it open, and I was thinking there was an issue with it (there wasn't), also too small to warrant its own PR IMO

PR #2549, opened 2021-04-09
Fix UI test coverage

abrooksv:
Create a jacoco server for the life of our ui tests that listens to coverage from the IDE

## License

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -30,6 +30,7 @@ dependencies {
     testImplementation("com.pinterest.ktlint:ktlint-core:$ktlintVersion")
     testImplementation("com.pinterest.ktlint:ktlint-test:$ktlintVersion")
 
+    implementation("org.jacoco:org.jacoco.core:0.8.6")
     implementation("org.gradle:test-retry-gradle-plugin:$gradleRetryPluginVersion")
     implementation("com.adarshr:gradle-test-logger-plugin:$gradleTestLoggerPlugin")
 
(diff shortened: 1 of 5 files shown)


buildSrc/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
     testImplementation("com.pinterest.ktlint:ktlint-core:$ktlintVersion")
     testImplementation("com.pinterest.ktlint:ktlint-test:$ktlintVersion")
 
+    implementation("org.jacoco:org.jacoco.core:0.8.6")
```
hunterwerlla, line 33: nit: put version in gradle properties file

buildSrc/src/main/kotlin/software/aws/toolkits/gradle/jacoco/RemoteCoverage.kt
```diff
@@ -0,0 +1,104 @@
...
+            }
+        } ?: task.logger.warn("$task does not have Jacoco enabled on it")
+    }
+
+    abstract class JacocoServer : BuildService<JacocoServer.Params>, AutoCloseable {
+        interface Params : BuildServiceParameters {
+            val execFile: RegularFileProperty
+        }
+
+        private val serverSocket = ServerSocket(6300) // Default jacoco port
```
hunterwerlla, line 48: port is duplicated, probably makes sense as `DEFAULT_JACOCO_PORT`

buildSrc/build.gradle.kts
abrooksv: Pulls version from the Gradle plugin since we use the default

buildSrc/src/main/kotlin/software/aws/toolkits/gradle/jacoco/RemoteCoverage.kt
abrooksv: Fixed

hunterwerlla (approved):

PR #2552, opened 2021-04-09
Improve ValidateLocalizedMessages

hunterwerlla:
- When there is not an `=` in a line, fail (it previously passed)
- Add line numbers to errors so you can quickly jump to them when run through the ide
## Screenshots
Screen Shot 2021-04-09 at 1 41 37 PM
Screen Shot 2021-04-09 at 1 41 49 PM

## License

### DIFF
--- a/buildSrc/src/main/kotlin/software/aws/toolkits/gradle/resources/ValidateMessages.kt
+++ b/buildSrc/src/main/kotlin/software/aws/toolkits/gradle/resources/ValidateMessages.kt
@@ -14,6 +14,9 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin.VERIFICATION_GROUP
 import java.time.Instant
 
 open class ValidateMessages : DefaultTask() {
+    private companion object {
+        const val COPYRIGHT_HEADER_LINES = 2
+    }
     @InputFiles
     val paths: ConfigurableFileCollection = project.objects.fileCollection()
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/software/aws/toolkits/gradle/resources/ValidateMessages.kt
```diff
@@ -35,18 +38,20 @@ open class ValidateMessages : DefaultTask() {
                 fileLines
                     // filter out blank lines and comments
                     .filter { it.isNotBlank() && it.trim().firstOrNull() != '#' }
-                    .mapNotNull {
+                    .mapIndexed { lineNumber, it ->
                         if (it.contains("=")) {
                             it
                         } else {
-                            logger.warn(""""$filePath contains invalid message missing a '=': "$it"""")
+                            logger.error(""""$filePath:${lineNumber + COPYRIGHT_HEADER_LINES} contains invalid message missing a '=': "$it"""")
```
rli, line 45: nice the lack of line numbers was super annoying. as a future enhancement can we have a task that autosorts?

rli (approved):

buildSrc/src/main/kotlin/software/aws/toolkits/gradle/resources/ValidateMessages.kt
hunterwerlla: Yeah that would be good

PR #2554, opened 2021-04-09
Fix some deprecations

abrooksv:
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
@@ -3,6 +3,7 @@
 
 package software.aws.toolkits.jetbrains.services.clouddebug.actions
 
+import com.intellij.build.BuildContentManager
 import com.intellij.build.BuildViewManager
 import com.intellij.build.DefaultBuildDescriptor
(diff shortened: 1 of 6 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/clouddebug/actions/PseCliAction.kt
```diff
@@ -83,7 +83,8 @@ abstract class PseCliAction(val project: Project, val actionName: String, privat
 
                     runInEdt {
                         // Safe access because it is possible to close the window before this completes
-                        toolWindowManager.getToolWindow(ToolWindowId.BUILD)?.show(null)
+                        @Suppress("UsePropertyAccessSyntax")
+                        BuildContentManager.getInstance(project).getOrCreateToolWindow().show(null)
```
hunterwerlla, line 87: Add a comment why not using property access syntax, also comment is now out of date

abrooksv: Ah it just becomes the confusing name `orCreateToolWindow` which I found misleading

hunterwerlla: Oh yeah, that makes sense

hunterwerlla (approved):

PR #2572, opened 2021-04-21
Do not check for a codepipeline system tag on delete or modify of resources

abrooksv:
This system looks to never have been used and we never added it to VSC

## License


hunterwerlla (approved):
Good cleanup

Issue #2564, opened 2021-04-18
Debugger not terminating in Intellij IDEA 2021.1 with new update (1.26-211)


ashkansk:
**Describe the bug**
When debugging a long running lambda function from template, debugger continues execution even the debugger is stopped from the GUI and there's no way to stop it from continuing the execution except killing the process by Task Manager. Closing the IDE doesn't help either.

**To reproduce**
1. Create an AWS Serverless Application project
2. Write a lambda function with a long loop (like 1000 iterations with a 100 ms thread sleep in it)
3. Add the lambda function to the template.yml and use the play button in the side bar of the editor (when still in the template file) which opens the run configurations of the Intellij IDEA.
4. Debug the lambda function using the debug button
5. Stop the debugging to see the debugging still continues in the debug console

**Expected behavior**
Debugging session should be stopped properly.

**Your Environment**
- OS: Windows 10
- JetBrains product: Intellij IDEA
- JetBrains product version: 2021.1
- AWS Toolkit version: 1.26-211
- SAM CLI version: 1.19.1

Srinu3366:
I am facing a similar issue when using it for Python Lambda debugging.

ashkansk:
> 
> 
> Which language are you seeing this behavior for?

I'm using java, Corretto 11

hunterwerlla:
Just to confirm, which stop button are you pressing? There are two since there is a debug session and the process itself (It's an implementation detail due to remote debugging) or are you pressing the "stop all" button? based on your report, I found that it seems that one of them stops both but not the other, so I am wondering if that is the only issue or there are further issues.

See: 
<img width="221" alt="Screen Shot 2021-04-28 at 10 20 01 AM" src="https://user-images.githubusercontent.com/11964782/116446024-648cfc80-a80b-11eb-877e-a2ea319cb49e.png">

ashkansk:
I use stop button on the top of the toolbar, not the Stop All button

![image](https://user-images.githubusercontent.com/4866898/116511140-3c6cce80-a8db-11eb-8e3d-81d7c1aa6002.png)

Since I reverted back to the older version, I don't have it to test it again with Stop All button, but I'm pretty sure that I tested all the ways to stop it but none of them worked.

hunterwerlla:
That stop is the normal way of stopping, so it sounds like the PR fixes the issue you are encountering. Thanks for reporting this!

PR #2578, opened 2021-04-28
While debugging a Lambda function locally, make sure stopping the deb…

hunterwerlla:
…ugger will always stop the underlying SAM cli process 

See large comment on the reasoning of the change
## Related Issue(s)

#2564

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/steps/AttachDebugger.kt
```diff
@@ -73,6 +98,7 @@ class AttachDebugger(
     }
 
     private companion object {
+        const val endText = "END RequestId: "
```
rli, line 101: do we have a place noting all of the magic SAM strings we're relying on?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/steps/AttachDebugger.kt
hunterwerlla: Currently they are where they are used, but maybe we should move them into SamExecutable, thoughts?

abrooksv:
Is it possible to write a test? For example, tweak an existing test so that when we hit a breakpoint, instead of resuming, we kill off the debugger session instead.

hunterwerlla:
> Is it possible to write a test? For example, tweak an existing test so that when we hit a breakpoint, instead of resuming, we kill off the debugger session instead.

Added an integration test, still need to investigate the aparent dotnet image issue

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/steps/AttachDebugger.kt
rli: yeah i like that idea

jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/RunConfigTestUtils.kt
```diff
@@ -89,6 +89,30 @@ fun executeRunConfigurationAndWait(runConfiguration: RunConfiguration, executorI
...
+            override fun processStarted(debugProcess: XDebugProcess) {
+                println("Debugger attached: $debugProcess")
+
+                debugProcess.session.addSessionListener(
+                    object : XDebugSessionListener {
+                        override fun sessionPaused() {
+                            runInEdt {
+                                val suspendContext = debugProcess.session.suspendContext
+                                println("Stopping: $suspendContext")
+                                callback()
```
abrooksv, line 105: We dont seem to use the callback, do we need it?

abrooksv (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/RunConfigTestUtils.kt
hunterwerlla: Correct, removed

PR #2579, opened 2021-04-29
Move from ktlint to detekt

hunterwerlla:
- Add DeteKt for static code analysis 
- Port ktlint rules to work with DeteKt instead of KtLint
- Uses its own script separate from conventions. This is because I could not get it to work if it used `toolkit-kotlin-conventions` it failed silently for some reason, even with `--debug` I couldn't find any relevant logs.

TODO:
- any rules we want to adjust?
- ~Add baseline files so it doesn't fail~
- ~is there any better way to do this than subprojects? It currently doesn't work in buildSrc and including kotlin conventions breaks it entierly~

## License


detekt-rules/detekt.yml
```diff
@@ -0,0 +1,185 @@
...
+    active: true
+  SpacingBetweenDeclarationsWithComments:
+    active: true
+    excludes: [ '**/icons/**' ]
+
+naming:
+  active: true
+  ClassNaming:
+    active: true
+    excludes: [ '**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**' ]
```
rli, line 67: nit: android/ios seem irrelevant

detekt-rules/src/main/kotlin/software/aws/toolkits/gradle/detekt/rules/BannedImportsRule.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override fun visitImportList(importList: KtImportList) {
+        super.visitImportList(importList)
+        importList.imports.forEach { element ->
+            if (element.importedFqName?.asString() == "org.assertj.core.api.Assertions") {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(element),
+                        message = "Import the assertion you want to use directly instead of importing the top level Assertions"
```
rli, line 25: too bad that `ForbiddenImport` doesn't offer suggestions like our rule does

detekt-rules/detekt.yml
```diff
@@ -0,0 +1,185 @@
...
+  GlobalCoroutineUsage:
+    active: true
+  RedundantSuspendModifier:
+    active: true
+  SleepInsteadOfDelay:
+    active: true
+  SuspendFunWithFlowReturnType:
+    active: true
+
+formatting:
```
rli, line 34: what about the `style` or `empty-blocks` rule sets?

buildSrc/src/main/kotlin/toolkit-detekt.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+    buildUponDefaultConfig = false
+    parallel = true
+    allRules = false
+    config = files("$rulesProject/detekt.yml")
+    baseline = file("$rulesProject/baseline.xml")
+
+    reports {
+        html.enabled = true // observe findings in your browser with structure and code snippets
+        xml.enabled = true // checkstyle like format mainly for integrations like Jenkins
+        sarif.enabled = true // standardized SARIF format to support integrations with Github Code Scanning
```
rli, line 31: do we need to upload the SARIF file?

https://docs.github.com/en/code-security/secure-coding/uploading-a-sarif-file-to-github#uploading-a-code-scanning-analysis-with-github-actions

detekt-rules/detekt.yml
hunterwerlla: Yeah this file was generated then I messed with it, I'll remove some of the irrelevent stuff

detekt-rules/src/main/kotlin/software/aws/toolkits/gradle/detekt/rules/BannedImportsRule.kt
hunterwerlla: Yep, otherwise I would have used it

detekt-rules/detekt.yml
```diff
@@ -151,7 +149,7 @@ potential-bugs:
     active: true
   LateinitUsage:
     active: false
-    excludes: [ '**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**' ]
+    excludes:  [ '**/test/**' ]
```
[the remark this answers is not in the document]
hunterwerlla: Completely removed, we use tst instead of test and we want to lint tests so shrug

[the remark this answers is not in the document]
hunterwerlla: wait I didn't realize this was still in here, removed because we use late init everywhere (especially in tests and UI)

PR #2588, opened 2021-05-07
Enable autoformatting and re-enable missing ktlint rules

rli:
## License


hunterwerlla (approved):

detekt-rules/src/test/kotlin/ExpressionBodyRuleTest.kt
```diff
@@ -114,4 +114,3 @@ class ExpressionBodyRuleTest {
         assertThat(rule.lint(code)).isEmpty()
     }
 }
-
```
abrooksv, line 117: Any idea why the final new line bu was striped? but only in the detekt-rules project

PR #2612, opened 2021-05-24
Ecs exec terminal plugin requirement

manodnyab:
## Description

Display the "Open Shell" Action only if the terminal plugin is installed

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
```diff
@@ -159,6 +160,6 @@ class ExecuteCommandInShellAction(
 
     override fun update(e: AnActionEvent) {
         e.presentation.isVisible = container.service.enableExecuteCommand() &&
-            !EcsUtils.isInstrumented(container.service.serviceArn())
+            !EcsUtils.isInstrumented(container.service.serviceArn()) && pluginIsInstalledAndEnabled("org.jetbrains.plugins.terminal")
```
rli, line 163: we should gate this in the plugin.xml file instead (example: https://github.com/aws/aws-toolkit-jetbrains/blob/b76bb7b898418c2309c264ea789785fecab5c202/jetbrains-core/resources/META-INF/plugin.xml#L72 + https://github.com/aws/aws-toolkit-jetbrains/blob/2c09b2784f55f6a33618c872dccacde4f0537e76/jetbrains-core/resources/META-INF/ext-docker.xml#L7-L11)

rli: @abrooksv i think it should be possible to get rid of the dynamically built action group and make this work, but do you have strong opinions?

abrooksv: I mentioned it in a previous PR but I am not sure how...

We build these actions per container inside of `ContainerActions`. There is no way that I know of to inject additional data into the the action group to make it work.

Actions defined in the plugin.xml have to be no-args constructor so we need a way to get the container info into them. Usually this is done by the DataProvider, but not sure we can in this situation.

abrooksv (approved):

Issue #2605, opened 2021-05-18
Property CodeUri not found in HelloWorldFunction after SAM init


autohandle:
**Describe the bug**

**To reproduce**

o new project
- Package Type: Image
- Create
- Runtime: python3.7 (python3.8 has its own problems)

**Expected behavior**

expected a new project, got an error
examining template.html, it doesn't have a: CodeUri, and
PackageType:Image is underlined in red

**Screenshots**

**Your Environment**
- OS: macOS Big Sur, 11.3.1
- JetBrains product: pycharm 2021.1
- JetBrains product version: Build #PY-211.7142.13, built on April 21, 2021
- AWS Toolkit version: 1-26-211
- SAM CLI version: SAM CLI, version 1.23.0
- JVM/Python version:

**Additional context**

```
java.lang.IllegalStateException: Property CodeUri not found in HelloWorldFunction
	at software.aws.toolkits.jetbrains.services.cloudformation.SamFunction.getScalarProperty(Resources.kt:42)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion$getCodeUrisFromTemplate$$inlined$runReadAction$1$lambda$1.invoke(SamCommon.kt:57)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion$getCodeUrisFromTemplate$$inlined$runReadAction$1$lambda$1.invoke(SamCommon.kt:22)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:210)
	at kotlin.sequences.SequencesKt___SequencesKt.toCollection(_Sequences.kt:752)
	at kotlin.sequences.SequencesKt___SequencesKt.toMutableList(_Sequences.kt:782)
	at kotlin.sequences.SequencesKt___SequencesKt.toList(_Sequences.kt:773)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion$getCodeUrisFromTemplate$$inlined$runReadAction$1.compute(actions.kt:66)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:834)
	at software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon$Companion.getCodeUrisFromTemplate(SamCommon.kt:71)
[...]

PR #2618, opened 2021-05-28
Add source roots correctly while generating SAM projects

manodnyab:
## Description

CodeUri Property is required only if the packageType is Zip. (Source: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-codeuri), so this change ensures that the codeUri is not accessed when the packageType is Image, thus resolving the error mentioned in the related Github Issue.

## Related Issue(s)

https://github.com/aws/aws-toolkit-jetbrains/issues/2605
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/python/PythonSamProjectWizard.kt
```diff
@@ -51,7 +51,10 @@ abstract class PythonSamProjectTemplate : SamAppTemplateBased() {
         indicator: ProgressIndicator
     ) {
         super.postCreationAction(settings, contentRoot, rootModel, indicator)
-        addSourceRoots(rootModel.project, rootModel, contentRoot)
+        if(settings.packagingType != PackageType.IMAGE) {
```
rli, line 54: image templates still technically have source code (and therefore source roots). should we read out the docker context and use that instead?

abrooksv: See `software.aws.toolkits.jetbrains.services.lambda.sam.SamTemplateUtils#getCodeLocation`

The current method it is using (`software.aws.toolkits.jetbrains.services.lambda.sam.SamCommon.Companion#getCodeUrisFromTemplate`) should be migrated off of and deleted

manodnyab: Migrated off getCodeUrisFromTemplate from this function, will add  a separate PR for deleting it completely

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SamProjectWizard.kt
```diff
@@ -68,11 +71,28 @@ abstract class SamProjectTemplate {
...
+        val templatePath = Paths.get(template.parent.path)
+        runInEdt {
+            val functions = SamTemplateUtils.findFunctionsFromTemplate(project, template)
+            val functionLocations = functions.map {
+                val codeLocation = SamTemplateUtils.getCodeLocation(template.toNioPath().toAbsolutePath(), it.logicalName)
+                if(codeLocation.startsWith("./")){
+                    templatePath.resolve(codeLocation.substring(2))
+                } else {
+                    templatePath.resolve(codeLocation)
+                }
```
abrooksv, line 83: I dont think you need to check "./", the path API should always relativize it if its not "/"

I would do `templatePath.parent.resolve(codeLocation)`

manodnyab: Made the change

abrooksv:
Detekt failures, but look good once you fix those

PR #2631, opened 2021-06-07
Progress indicator for enabling and disabling commands

manodnyab:
## Description

1) Adding a progress indicator while enabling and disabling command execution
2) Error handling if the session manager plugin is not installed

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -30,36 +32,51 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
-            if (enable) {
-                notifyError(message("ecs.execute_command_enable"), message("ecs.execute_command_enable_failed", service.serviceName()))
-                EcsTelemetry.enableExecuteCommand(project, Result.Failed)
-            } else {
-                notifyError(message("ecs.execute_command_disable"), message("ecs.execute_command_disable_failed", service.serviceName()))
-                EcsTelemetry.disableExecuteCommand(project, Result.Failed)
-            }
+    private fun checkServiceState(project: Project, service: Service, enable: Boolean) {
+        val title = if (enable) {
+            "Enabling Command Execution"
```
abrooksv, line 41: these should live in the messages file

manodnyab: Yes, added, forgot about that!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
```diff
@@ -145,7 +145,7 @@ class ExecuteCommandInShellAction(
     private val container: ContainerDetails
 ) : AnAction(message("ecs.execute_command_run_command_in_shell"), null, null) {
     override fun actionPerformed(e: AnActionEvent) {
-        val sessionManagerInstalled = SessionManagerPluginInstallationVerfication.checkInstallation()
+        val sessionManagerInstalled = SessionManagerPluginInstallationVerification.checkInstallation()
```
kiiadi, line 148: Minor unrelated nit, since we duplicate the handling of the plugin logic in multiple places and the handling is the same (showing a warning) - I'd probably push them together so you have something like :

```kt
SessionManagerPluginInstallationVerification.requireSessionManager {
  // do thing that requires session manager
}
```

```kt
object SessionManagerPluginInstallationVerification {
  fun requireSessionManager(block: () -> Unit) {
    if( /* session manager plugin verified */ ) {
      block()
    } else {
      SessionManagerPluginWarning(project).show()<br class="Apple-interchange-newline">
    }
  }
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -43,33 +45,47 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+        ProgressManager.getInstance().run(
+            object : Task.Backgroundable(project, title, false) {
+                override fun run(indicator: ProgressIndicator) {
+                    /* There appears to be a lag between the time the UpdateService call is made and
+                    the time the changes are surfaced via the DescribeServices call.
+                    Hence a delay of 5000 milliseconds is added before polling for the completion state of the service
+                    */
+                    Thread.sleep(5000)
+                    val request = DescribeServicesRequest.builder().cluster(service.clusterArn()).services(service.serviceArn()).build()
+                    val client = project.awsClient<EcsClient>()
```
kiiadi, line 63: What happens if the customer changes the profile/region in the 5 seconds between when they call the action and this starts?

I think we probably actually want to pass the connection along, rather than assuming that the "current connection" is going to remain static.

manodnyab: Currently, if they change the region/credentials within the 5 secs they will get a notification saying the command could not be enabled/disabled. So this makes sense, I'll make the change.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
manodnyab: Made the change, thanks!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: Thread.sleep is not required anymore, however we need to pass the connection settings for refreshing the explorer if the region and credentials change during the time taken to enable/disable ecs exec

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/RefreshAwsTree.kt
```diff
@@ -5,15 +5,26 @@ package software.aws.toolkits.jetbrains.core.explorer
...
+        }
     } else {
-        this.clearResourceForCurrentConnection(resource)
+        if (resource == null) {
+            AwsResourceCache.getInstance().clear(connectionSettings)
+        } else {
+            AwsResourceCache.getInstance().clear(resource, connectionSettings)
+        }
     }
+
```
abrooksv, line 27: I am not sure this is is the best route for this since in this branch it will still refresh the tree even though it shouldn't be.

We should probably have a higher level util that takes care of do this if connectionSetts still match, else only invalidate cache

manodnyab: what if an else if condition is added instead of else? Also, do you mean we don't refresh the AWS explorer just invalidate the cache?

abrooksv (approved):

PR #2632, opened 2021-06-07
Clean up RegionSelector, expose it as a UI DSL component

abrooksv:
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
@@ -7,22 +7,20 @@ import com.intellij.openapi.ui.ComboBox
 import com.intellij.ui.CollectionComboBoxModel
 import com.intellij.ui.ComboboxSpeedSearch
 import com.intellij.ui.SimpleListCellRenderer
+import com.intellij.ui.layout.Cell
+import com.intellij.ui.layout.CellBuilder
+import com.intellij.ui.layout.applyToComponent
 import software.aws.toolkits.core.region.AwsRegion
+import software.aws.toolkits.jetbrains.core.region.AwsRegionProvider
 import software.aws.toolkits.jetbrains.utils.ui.selected
+import kotlin.reflect.KMutableProperty0
 
 /**
  * Combo box used to select a region
  * TODO: Determine the UX for the box, do we want to categorize?
  */
 class RegionSelector : ComboBox<AwsRegion>() {
-    private val comboBoxModel = object : CollectionComboBoxModel<AwsRegion>() {
-        fun setItems(newItems: List<AwsRegion>) {
-            internalList.apply {
-                clear()
-                addAll(newItems)
-            }
-        }
-    }
+    private val comboBoxModel = CollectionComboBoxModel<AwsRegion>()
 
     init {
         model = comboBoxModel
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
```diff
@@ -31,7 +29,7 @@ class RegionSelector : ComboBox<AwsRegion>() {
     }
 
     fun setRegions(regions: List<AwsRegion>) {
-        comboBoxModel.items = regions
+        comboBoxModel.replaceAll(regions)
```
manodnyab, line 32: Whats the behavior change after using replaceAll?

rli: behavior should be the same, but we're no longer hacking in something that the `CollectionComboBoxModel` already implements

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
```diff
@@ -44,9 +42,21 @@ class RegionSelector : ComboBox<AwsRegion>() {
...
         }
 
-    private companion object {
-        val RENDERER = SimpleListCellRenderer.create<AwsRegion>("") {
+    companion object {
+        private val RENDERER = SimpleListCellRenderer.create<AwsRegion>("") {
             it.displayName
         }
+
+        fun Cell.regionSelector(prop: KMutableProperty0<AwsRegion>, serviceId: String? = null): CellBuilder<ComboBox<AwsRegion>> {
```
rli, line 50: clarify that the service id filters the region list?

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
```diff
@@ -44,9 +42,21 @@ class RegionSelector : ComboBox<AwsRegion>() {
...
+            val regionProvider = AwsRegionProvider.getInstance()
+            val regions = when {
+                serviceId != null -> regionProvider.allRegionsForService(serviceId).values.toMutableList()
+                else -> regionProvider.allRegions().values.toMutableList()
+            }
+            val model = CollectionComboBoxModel(regions)
+            return comboBox(model, prop, RENDERER).applyToComponent {
```
rli, line 57: is this preferred over reusing our own component?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
abrooksv: Richard is correct

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
abrooksv: I don't think we need to. long term we should move off .form files and so we shouldn't need our own

jetbrains-core/src/software/aws/toolkits/jetbrains/ui/RegionSelector.kt
abrooksv: Will added

rli (approved):

PR #2642, opened 2021-06-11
Migrate off deprecated DefaultProgramRunner

abrooksv:
Migrated off deprecated method.

Also switched it to async so that invoking the state is on BG thread always

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/remote/RemoteLambdaRunner.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/remote/RemoteLambdaRunner.kt
@@ -25,14 +25,14 @@ class RemoteLambdaRunner : AsyncProgramRunner<RunnerSettings>() {
 
     override fun execute(environment: ExecutionEnvironment, state: RunProfileState): Promise<RunContentDescriptor?> {
         val runPromise = AsyncPromise<RunContentDescriptor?>()
+        val remoteState = state as RemoteLambdaState
         ApplicationManager.getApplication().executeOnPooledThread {
             try {
-                state.execute(environment.executor, this)?.let {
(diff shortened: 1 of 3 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/remote/RemoteLambdaRunner.kt
```diff
@@ -4,14 +4,39 @@
...
+        DefaultRunExecutor.EXECUTOR_ID == executorId && profile is RemoteLambdaRunConfiguration
+
+    override fun execute(environment: ExecutionEnvironment, state: RunProfileState): Promise<RunContentDescriptor?> {
+        val runPromise = AsyncPromise<RunContentDescriptor?>()
+        ApplicationManager.getApplication().executeOnPooledThread {
+            val builder = state.execute(environment.executor, this)?.let {
+                RunContentBuilder(it, environment)
+            }
+
+            builder?.let {
```
rli, line 33: does the promise needs to be failed if the builder is null?

abrooksv: Our code can't return null, we can probably downcast it to make it safe

rli (approved):

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/remote/RemoteLambdaRunner.kt
abrooksv: No longer possible if we downcast, and thats guarded by the canRun method so we good :D

rli (approved):

PR #2651, opened 2021-06-14
Fix Windows test getting stuck for ~2 hours

abrooksv:
Logic came from: 
https://github.com/JetBrains/intellij-community/blob/master/plugins/maven/src/test/java/org/jetbrains/idea/maven/MavenImportingTestCase.java#L409

https://github.com/JetBrains/intellij-community/blob/master/plugins/maven/src/test/java/org/jetbrains/idea/maven/MavenImportingTestCase.java#L78

## License


kiiadi (approved):
uuuum k

PR #2653, opened 2021-06-14
Check for permissions required for ECS Exec

manodnyab:
## Description

Checks whether the task role has the basic permissions required for ECS Exec

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -87,4 +97,53 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
             }
         )
     }
+
+    fun checkRequiredPermissions(project: Project, clusterArn: String, taskArn: String) {
+        val ecsClient = project.awsClient<EcsClient>()
+        val iamClient = project.awsClient<IamClient>()
+        val task = ecsClient.describeTasks(DescribeTasksRequest.builder().tasks(taskArn).cluster(clusterArn).build())
+        val taskRoleArn = if (task.tasks()[0].overrides().taskRoleArn() != null) {
+            task.tasks()[0].overrides().taskRoleArn()
```
abrooksv, line 106: Why do we always take the first task?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -87,4 +97,53 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+        } else {
+            val roleArn = project.getResourceNow(EcsResources.describeTaskDefinition(task.tasks()[0].taskDefinitionArn())).taskRoleArn()
```
abrooksv, line 108: What thread is this on?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -87,4 +97,53 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
             }
         )
     }
+
+    fun checkRequiredPermissions(project: Project, clusterArn: String, taskArn: String) {
```
abrooksv, line 101: this method is super dense

please break it up into smaller methods and add comments

abrooksv, line 101: There is a lot of branching in here and 0 tests, please add some

abrooksv (changes requested):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialog.kt
```diff
@@ -71,7 +79,19 @@ class RunCommandDialog(private val project: Project, private val container: Cont
     override fun doOKAction() {
         super.doOKAction()
         commandsEnteredPreviously.add(command)
-        runCommand()
+        val task = tasks.selected() ?: throw IllegalStateException("Task not Selected")
+        launch {
+            try {
+                EcsExecUtils.checkRequiredPermissions(project, container.service.clusterArn(), task)
```
abrooksv, line 85: This should be best effort, if they lack the ability to check the permissions, then we should not prevent them from running ecs exec, we just need to tell them it may fail

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: We are just checking the permissions on one task selected, so we need the details/TaskRoleArn on the first task of the DescribeTasks call

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: The thread begins here: https://github.com/aws/aws-toolkit-jetbrains/pull/2653/files#diff-dd5452d78ae3e069245b59d2a785ead9ef21710f88f584703405e6295df57bb4R87, and similarly for `RunCommandDialog` . Reason for doing that is exceptions could be caught from within the same thread instead of communicating between 2.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: okay, will do that!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
abrooksv: Can each task have a different role? If so do we need to search for the task that matches the requested one by the user?

manodnyab: Task Roles are attached to task definitions. Searching through numerous tasks and checking if all/any of them have required permissions might be more expensive (includes list-task, describe-task calls, followed by this function for each task) than checking for just the task selected by the user. One alternative that we can use is if the task doesn't have required permissions, we display the warning and redirect them to the same dialog to select another task.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: Tests added

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,6 +105,78 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+            }
+        }
+    }
+
+    fun checkRequiredPermissions(project: Project, clusterArn: String, taskArn: String): Boolean {
+        try {
+            val iamClient = project.awsClient<IamClient>()
+            val taskRoleArn = getTaskRoleArn(project, clusterArn, taskArn) ?: return false
+            val permissions = listOf(
+                message("session_manager_create_control_channel_permission"),
```
rli, line 156: this will never be localized so it can just be a const in the file IMO

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,6 +105,78 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+                if (launchType == LaunchType.EC2) {
+                    val ec2InstanceTaskRoleArn = getEc2InstanceTaskRoleArn(project, clusterArn, ecsClient, task)
+                    ec2InstanceTaskRoleArn
+                } else {
+                    return null
+                }
+            } else {
+                roleArn
+            }
+        }
```
rli, line 148: i think this would read better if we inverted the logic
```kt
val launchType = task.launchType()
when (launchType) {
    LaunchType.EC2 -> {
        // ec2 stuff
    }
    LaunchType.FARGATE -> {
        // fargate stuff
    }
    else -> {
        throw RuntimeException(message(/* launch type not supported */))
    }
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialog.kt
```diff
@@ -27,7 +33,9 @@ import javax.swing.JComponent
 import javax.swing.JTextField
 import javax.swing.plaf.basic.BasicComboBoxEditor
 
-class RunCommandDialog(private val project: Project, private val container: ContainerDetails) : DialogWrapper(project) {
+class RunCommandDialog(private val project: Project, private val container: ContainerDetails) :
+    DialogWrapper(project),
+    CoroutineScope by ApplicationThreadPoolScope("RunCommandDialog") {
```
rli, line 38: i thought we were moving away from coroutine scope by delegation?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,6 +105,78 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
         )
     }
 
+    fun getEc2InstanceTaskRoleArn(project: Project, clusterArn: String, ecsClient: EcsClient, task: software.amazon.awssdk.services.ecs.model.Task): String? {
```
rli, line 108: probably do something like:
```
import software.amazon.awssdk.services.ecs.model.Task as EcsTask
```
instead of using fully qualified class name

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,6 +105,78 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+                    .builder()
+                    .cluster(clusterArn)
+                    .containerInstances(containerInstanceArn).build()
+            )
+            val ec2InstanceId = res.containerInstances().first().ec2InstanceId()
+            val instanceProfileArn = project.awsClient<Ec2Client>().describeInstances(
+                DescribeInstancesRequest.builder().instanceIds(ec2InstanceId).build()
+            ).reservations().first().instances().first().iamInstanceProfile().arn() ?: return null
+            val instanceProfileName = instanceProfileArn.substringAfter(":instance-profile/")
+            return iamClient.getInstanceProfile(GetInstanceProfileRequest.builder().instanceProfileName(instanceProfileName).build()).instanceProfile()
```
rli, line 123: style in our codebase prefers using the consumer pattern here instead of explicit builders
```kt
iamClient.getInstanceProfile {
    it.instanceProfileName(instanceProfileName)
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
rli, line 108: also i think this can be private

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,6 +105,78 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+        val task = ecsClient.describeTasks(DescribeTasksRequest.builder().tasks(taskArn).cluster(clusterArn).build()).tasks().first()
+        return if (task.overrides().taskRoleArn() != null) {
+            task.overrides().taskRoleArn()
+        } else {
+            val roleArn = project.getResourceNow(EcsResources.describeTaskDefinition(task.taskDefinitionArn())).taskRoleArn()
+            if (roleArn == null) {
+                val launchType = task.launchType()
+                if (launchType == LaunchType.EC2) {
+                    val ec2InstanceTaskRoleArn = getEc2InstanceTaskRoleArn(project, clusterArn, ecsClient, task)
+                    ec2InstanceTaskRoleArn
```
rli, line 141:
```suggestion
                    getEc2InstanceTaskRoleArn(project, clusterArn, ecsClient, task)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/OpenShellInContainerDialog.kt
```diff
@@ -78,7 +83,17 @@ class OpenShellInContainerDialog(
 
     override fun doOKAction() {
         super.doOKAction()
-        runExecCommand()
+        val task = tasks.selected() ?: throw IllegalStateException("Task not Selected")
+        launch {
+            val taskRoleFound: Boolean = EcsExecUtils.checkRequiredPermissions(project, container.service.clusterArn(), task)
```
rli, line 88: this kicks off a lot of AWS API calls. how do we provide a better experience here in the worst-case? what about caching?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtilsTest.kt
```diff
@@ -29,13 +57,187 @@ class EcsExecUtilsTest {
...
+        }
+
+        iamClient.stub {
+            on { simulatePrincipalPolicy(any<SimulatePrincipalPolicyRequest>()) } doAnswer {
+                SimulatePrincipalPolicyResponse.builder().evaluationResults(evaluationResult).build()
+            }
+        }
+
+        val haveRequiredPermissions = EcsExecUtils.checkRequiredPermissions(projectRule.project, clusterArn, taskArn)
+        assertThat(haveRequiredPermissions).isTrue
```
rli, line 240: should we test what happens when AWS calls error out too?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: I'll change it to constant...

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: Since we only need to find out whether the launch type is EC2 to determine if it has a task role in its instance profile, and other launch types(fargate, unknown) need to be treated the same, an if else block can also work i think

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: Will change this

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/OpenShellInContainerDialog.kt
manodnyab: Most of the API calls are made in the scenario where task role is not found in the task definition and if the launch type is EC2, honestly will probably need to understand from the exec team the probability of such a scenario occurring, cause if its not too much im not sure if we should cache it?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtilsTest.kt
manodnyab: yes, will add this test!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: made the change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: made the change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
manodnyab: made the change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialog.kt
abrooksv: We should be.

`private val coroutineScope = ApplicationThreadPoolScope("RunCommandDialog", disposable)` would be the more correct route.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
rli: for the unknown case, we should probably throw instead of letting behavior fall through (for example, the [EXTERNAL launch type](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Task.html#ECS-Type-Task-launchType) was released [recently](https://aws.amazon.com/blogs/aws/getting-started-with-amazon-ecs-anywhere-now-generally-available/)). the other thing im confused about is what happens if there is both an ec2 instance profile and task role arn defined in the task definition?

manodnyab: In this instance, since we have to check the attributes and move forward only if the launch type is EC2, i was thinking of if-else . But you're right, I'll add separate cases for supported and unknown launch types.
According to the ECS Exec Checker script, (referred to by the ECS Exec team :https://github.com/aws-containers/amazon-ecs-exec-checker/blob/main/check-ecs-exec.sh#L467), this is the order for retrieving the task role:
1) Check if the task role is overriden in the task
2) if not, consider the task role in the task definition
3) If this is null and launch type is EC2, check for an instance profile and see if there's a task role there
Otherwise we display a warning saying that valid task role not found

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,12 +108,76 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
         )
     }
 
+    private fun getEc2InstanceTaskRoleArn(project: Project, clusterArn: String, ecsClient: EcsClient, task: EcsTask): String? {
+        try {
+            val iamClient = project.awsClient<IamClient>()
+            val containerInstanceArn = task.containerInstanceArn()
+            val res = ecsClient.describeContainerInstances {
+                it.cluster(clusterArn).containerInstances(containerInstanceArn)
```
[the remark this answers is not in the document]
manodnyab: Changed this

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,12 +108,76 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+            val instanceProfileName = instanceProfileArn.substringAfter(":instance-profile/")
+            return iamClient.getInstanceProfile { it.instanceProfileName(instanceProfileName) }.instanceProfile().roles().first().arn() ?: null
+        } catch (e: Exception) {
+            return null
+        }
+    }
+
+    fun getTaskRoleArn(project: Project, clusterArn: String, taskArn: String): String? {
+        val ecsClient = project.awsClient<EcsClient>()
+        val task = ecsClient.describeTasks(DescribeTasksRequest.builder().tasks(taskArn).cluster(clusterArn).build()).tasks().first()
```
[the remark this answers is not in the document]
manodnyab: Converted this

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
```diff
@@ -95,12 +103,76 @@ object EcsExecUtils : CoroutineScope by ApplicationThreadPoolScope("EcsExec") {
...
+            }
+
+            val permissionResults = response.evaluationResults().map { it.evalDecision().name }
+            for (permission in permissionResults) {
+                if (permission != PolicyEvaluationDecisionType.ALLOWED.name) {
+                    return false
+                }
+            }
+        } catch (e: Exception) {
+            notifyWarn(message("ecs.execute_command_permissions_required_title"), message("ecs.execute_command_permissions_not_verified"))
```
rli, line 166: project should be passed along
```suggestion
            notifyWarn(
                project = project,
                title = message("ecs.execute_command_permissions_required_title"),
                content = message("ecs.execute_command_permissions_not_verified")
            )
```

rli (approved):

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/EcsExecUtils.kt
abrooksv: Good catch

manodnyab: Added the parameter names

PR #2656, opened 2021-06-16
Valid version check for AWS CLI

manodnyab:
## Description

Checks if AWS CLI version is valid for running ECS Exec

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
```diff
@@ -65,6 +65,34 @@ class ExecutableCommon {
             }
         }
 
+        @JvmStatic
+        fun checkSemVerVersionForParallelValidVersions(
+            version: SemVer,
+            minVersionLower: SemVer,
+            maxVersionLower: SemVer,
+            minVersionHigher: SemVer,
+            maxVersionHigher: SemVer,
```
abrooksv, line 74: Thoughts on taking a List<Pair<SemVer, SemVer>> supportedRanges?

We can then add a new method that is just a single one and creates a list out of it
`fun checkSemVerVersionForParallelValidVersions(version:SemVer, lower:SemVer, higher:SemVer, executableName: String) = checkSemVerVersionForParallelValidVersions(version, listOf(lower to higher), executableName)`

manodnyab: Pair sounds like a good idea, it'll be much cleaner and less confusing.
Since there's just 2 pairs, a list can be avoided, would improve readability?

abrooksv: I think the list can simplify the code since we can loop and avoid so many if's and it can scale (let's hope we dont need >2...) but the point still stands

manodnyab: Yep, makes sense, made the change!

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
```diff
@@ -65,6 +65,30 @@ class ExecutableCommon {
             }
         }
 
+        @JvmStatic
+        fun checkSemVerVersionForParallelValidVersions(
+            version: SemVer,
+            versionRange: List<ExecutableVersionRange>,
+            executableName: String
+        ) {
+            // for use of AWS CLI version 1 and version 2 in ECS Exec
```
abrooksv, line 74: Don't think we need this comment, especially since it is in common code and not related to that feature specifically

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
```diff
@@ -65,6 +65,30 @@ class ExecutableCommon {
...
+
+            val versionRanges = versionRange.joinToString(separator = " / ", transform = { "${it.min}  ≤ version < ${it.max}" })
+            val versionOutOfRangeMessage = message(
+                "executableCommon.version_range_wrong",
+                executableName,
+                versionRanges,
+                version
+            )
```
abrooksv, line 82: We can move this below the loop as a small optimization since we dont need it in a happy case

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
manodnyab: done

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
manodnyab: done

PR #2659, opened 2021-06-17
Stablize SchemaSelectionPanelTest

abrooksv:
* Also fix the detekt violations

## Testing
Used `schemaTemplateParametersBuiltAfterSelectionCustomerUploadedSchema`
Before:
Screen Shot 2021-06-16 at 4 38 41 PM

After:
Screen Shot 2021-06-16 at 4 53 26 PM

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanelTest.kt
```diff
@@ -223,13 +190,53 @@ class SchemaSelectionPanelTest {
...
+
+        return selector
     }
 
-    private fun load() {
-        initMockResourceCache()
-        schemaSelectionPanel.awsConnection = connectionManager.settingsManager.connectionSettings()
-        schemaSelectionPanel.reload()
-        schemaSelectionPanel.schemasSelector.waitToLoad()
+    private companion object {
```
manodnyab, line 212: Why are these moved to a companion object?

manodnyab (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/wizard/SchemaSelectionPanelTest.kt
abrooksv: 1. Naming convnetion is all uppercase is for static values so  they should be in companion
2. Constant vals should be in static fields, not per instance

PR #2678, opened 2021-06-25
Fix flaky multipleCallsInDifferentThreadsStillOnlyCallTheUnderlyingRe…

abrooksv:
…sourceOnce

Under high CPU, test can fail due to not waiting long enough

## Testing
Increased concurrency to 2000 to simulate higher CPU

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/core/AwsResourceCacheTest.kt
```diff
@@ -322,7 +322,7 @@ class AwsResourceCacheTest {
                 future
             }.toTypedArray()
             latch.countDown()
-            CompletableFuture.allOf(*futures).value
+            CompletableFuture.allOf(*futures).get(10, TimeUnit.SECONDS)
```
rli, line 325: wish there were a better way to test async code in java

rli (approved):

PR #2688, opened 2021-06-29
Ban use of Dispatchers

abrooksv:
## License


core/tst/CleanUpS3.kt
```diff
@@ -0,0 +1,93 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
rli, line 1: accident

abrooksv: Grrr... I need to move that to an ignore file...

rli (approved):

PR #2693, opened 2021-06-30
Credentials added as environment variables in Run Command Tool

manodnyab:
## Description

Adds credentials as environment variables

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandToolCustomizer.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.intellij.execution.configurations.GeneralCommandLine
+import com.intellij.openapi.actionSystem.CommonDataKeys
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.tools.ToolsCustomizer
+import software.aws.toolkits.jetbrains.core.credentials.AwsConnectionManager
+import software.aws.toolkits.jetbrains.core.credentials.toEnvironmentVariables
+
+class RunCommandToolCustomizer : ToolsCustomizer() {
+    override fun customizeCommandLine(dataContext: DataContext, commandLine: GeneralCommandLine): GeneralCommandLine {
```
abrooksv, line 14: This will affect any and all tools, which we should avoid

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandToolCustomizer.kt
```diff
@@ -0,0 +1,23 @@
...
+        val connectionSettings = dataContext.getData(CommonDataKeys.PROJECT)?.let { AwsConnectionManager.getInstance(it).connectionSettings() }
```
abrooksv, line 15: Cam we make sure we re-use the creds from the original invoke to avoid a possible race?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandToolCustomizer.kt
manodnyab: Removed the customizer

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandToolCustomizer.kt
manodnyab: Made the change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
```diff
@@ -190,3 +185,11 @@ class ExecuteCommandInShellAction(
             pluginIsInstalledAndEnabled("org.jetbrains.plugins.terminal") && AwsToolkit.isEcsExecEnabled()
     }
 }
+
+object CurrentConnectionSettings {
```
abrooksv, line 189: We should make this an extension method on the AwsConnectionManager and probably move it to that file as well

abrooksv, line 189: This feels like an odd home for this method

We should probably make ConnectionSettings.toEnvironmentVariables() be an extension method as well

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialog.kt
```diff
@@ -32,7 +31,7 @@ import javax.swing.JComponent
 import javax.swing.JTextField
 import javax.swing.plaf.basic.BasicComboBoxEditor
 
-class RunCommandDialog(private val project: Project, private val container: ContainerDetails) :
+class RunCommandDialog(private val project: Project, private val container: ContainerDetails, private val connectionSettings: Map<String, String>) :
```
abrooksv, line 34: Callers of this should not care about us turning the creds into env vars, I would just take the ConnectionSettings here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandRunProfile.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.tools.FilterInfo
+import com.intellij.tools.Tool
+import com.intellij.tools.ToolProcessAdapter
+import com.intellij.tools.ToolRunProfile
+import com.intellij.tools.ToolsCustomizer
+import java.io.File
+import javax.swing.Icon
+
+class RunCommandRunProfile(private val tool: Tool, private val dataContext: DataContext, private val credentials: Map<String, String>) : RunProfile {
```
abrooksv, line 39: Do we still need the tool? Or should we just take in a GenralCommandLine?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandRunProfile.kt
```diff
@@ -0,0 +1,127 @@
...
+        return commandLineState
+    }
+
+    override fun getName(): String = ToolRunProfile.expandMacrosInName(tool, dataContext)
+
+    override fun getIcon(): Icon? = null
+
+    fun getCommandLine(): GeneralCommandLine? {
+        if (StringUtil.isEmpty(tool.workingDirectory)) {
+            tool.workingDirectory = ("\$ProjectFileDir$")
```
abrooksv, line 63: I dont think we need all the macro logic for running a command on a remote machine, we can remove it and simplify this a lot

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandRunProfile.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        val commandLine = if (Registry.`is`("use.tty.for.external.tools", false)) PtyCommandLine().withConsoleMode(true) else GeneralCommandLine()
```
abrooksv, line 66: PtyCommandLine looks to be an old way of doing this, so I dont think we need to keep that logic path

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -64,12 +96,89 @@ class RunCommandDialogTest {
...
+                        DefaultRunExecutor.getRunExecutorInstance(),
+                        RunCommandRunProfile(execCommand, SimpleDataContext.getProjectContext(projectRule.project), connectionSettings)
+                    )
+                    .build {
+                        it.processHandler?.addProcessListener(object : ProcessAdapter() {
+
+                            override fun onTextAvailable(event: ProcessEvent, outputType: Key<*>) {
+                                super.onTextAvailable(event, outputType)
+                                environmentVariables.add(event.text.split("\n").first())
+                                counter.countDown()
```
abrooksv, line 133: You can move this to the onProcessTerminated and not tie the countdown to the number of lines to make the test less rigid

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
manodnyab: Added an Extension function to ConnectionSettings

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandRunProfile.kt
manodnyab: made the changes to remove tool and just use GeneralCommandLine

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
manodnyab: done

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
```diff
@@ -167,7 +169,10 @@ class ExecuteCommandInShellAction(
             if (EcsExecUtils.ensureServiceIsInStableState(project, container.service)) {
                 SessionManagerPluginInstallationVerification.requiresSessionManager(project) {
                     runInEdt {
-                        OpenShellInContainerDialog(project, container, CurrentConnectionSettings.getConnectionSettings(project)).show()
+                        val connectionSettings = AwsConnectionManager.getInstance(project).connectionSettings()
```
kiiadi, line 172: this is basically the same as the above - and I suspect we use something similar to this in quite a number of places. Would it be worth introducing a new extension function on `Project` something like `fun <T> Project.withAwsConnection(block: (ConnectionSettings) -> T) : T`

which we could then use like this:

```
project.withAwsConnection {
  OpenShellInContainerDialog(project, container, it).show()
}
```

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -114,36 +93,44 @@ class RunCommandDialogTest {
...
                             override fun onTextAvailable(event: ProcessEvent, outputType: Key<*>) {
                                 super.onTextAvailable(event, outputType)
                                 environmentVariables.add(event.text.split("\n").first())
-                                counter.countDown()
+                            }
+                        })
+                        it.processHandler?.addProcessListener(object : ProcessAdapter() {
+                            override fun processTerminated(event: ProcessEvent) {
+                                super.processTerminated(event)
+                                assertThat(environmentVariables[1]).isEqualTo("AccessEcsExecDummy")
```
kiiadi, line 118: how are these created?

We have helper functions that generate region codes and credentials:

* `software.aws.toolkits.core.region.AwsRegionTestKt#anAwsRegion`
* `software.aws.toolkits.core.credentials.MocksKt#aCredentialsIdentifier`

Rather than hard-coding `us-east-1` into tests which can make certain assumptions that are later broken.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
manodnyab: The extension function would always return ConnectionSettings right?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
```diff
@@ -143,7 +142,10 @@ class ExecuteCommandAction(
             if (EcsExecUtils.ensureServiceIsInStableState(project, container.service)) {
                 SessionManagerPluginInstallationVerification.requiresSessionManager(project) {
                     runInEdt {
-                        RunCommandDialog(project, container, CurrentConnectionSettings.getConnectionSettings(project)).show()
+                        val connectionSettings = AwsConnectionManager.getInstance(project).connectionSettings()
+                        if (connectionSettings != null) {
+                            RunCommandDialog(project, container, connectionSettings).show()
+                        }
```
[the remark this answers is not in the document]
manodnyab: yep, Will add the exception handling and remove the null check!

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
manodnyab: The credentials are created here: https://github.com/aws/aws-toolkit-jetbrains/pull/2693/files/03ead727a93fd234f3179725393c7d26f8f66a19..aff9010a7e697b863fa0d80bce99e503191926fc#diff-48601f6230786dbe8deaf616942766b274bb77efcd36051f856b3d77da011a53R70
I'll add the mock regions and credentials change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
kiiadi: I'd say its return type would be `T` where `T` is whatever type is returned by the passed `block` - and it is `T` not `T?` because the `withAwsConnection` method would through an exception if the connection wasn't valid - this should only be used in cases where the connection should be valid.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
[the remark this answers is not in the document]
manodnyab: Added an extension function that retrieves the ConnectionSettings and throws an exception if connectionSettings are null

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/ContainerActions.kt
manodnyab: Added the function

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -30,19 +48,37 @@ class RunCommandDialogTest {
...
     private val clusterArn = "arn:aws:ecs:us-east-1:123456789012:cluster/cluster-name"
     private val serviceArn = "arn:aws:ecs:us-east-1:123456789012:service/service-name"
     private val taskArn = "arn:aws:ecs:us-east-1:123456789012:task/task-name"
     private val ecsService = Service.builder().serviceArn(serviceArn).clusterArn(clusterArn).build()
 
     private val containerDefinition = ContainerDefinition.builder().name("sample-container").build()
     private val container = ContainerDetails(ecsService, containerDefinition)
-
+    lateinit var connectionSettings: ConnectionSettings
     val command = "ls"
```
abrooksv, line 63: These should be private

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -30,19 +48,37 @@ class RunCommandDialogTest {
...
     private val task = Task.builder().clusterArn(clusterArn).taskArn(taskArn).build()
     private val taskList = listOf(task.taskArn())
     private val containerName = containerDefinition.name()
     private val verifyCommand = "ecs execute-command --cluster $clusterArn --task $taskArn --command $command --interactive --container $containerName"
+    private val dummyRegion = anAwsRegion()
+
+    @Before
+    fun setup() {
+        val accountSettings = MockAwsConnectionManager.getInstance(projectRule.project)
```
abrooksv, line 72: This should be replaced with the `@Rule` ProjectAccountSettingsManagerRule

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -30,19 +48,37 @@ class RunCommandDialogTest {
...
+        val dummyCredential = aCredentialsIdentifier(defaultRegionId = dummyRegion.id)
+        val credentials = credentialManager.addCredentials(
```
abrooksv, line 74: The default args for addCredentials will make you all of this and return the CredentialIdentifier

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -51,25 +87,92 @@ class RunCommandDialogTest {
...
+                                assertThat(environmentVariables[3]).isEqualTo(dummyRegion)
+                                assertThat(environmentVariables[4]).isEqualTo(dummyRegion)
+                            }
+                        })
+                    }
+            environment.runner.execute(environment)
         }
     }
+
+    private fun makeSampleCliExecutable(path: String? = null, exitCode: Int = 0): Path {
```
abrooksv, line 136: Looks like path is hard coded to always be null? So no need for the param

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
```diff
@@ -51,25 +87,92 @@ class RunCommandDialogTest {
...
+
+                            override fun processTerminated(event: ProcessEvent) {
+                                super.processTerminated(event)
+                                assertThat(environmentVariables[1]).isEqualTo("AccessKeyEcsExecDummy")
+                                assertThat(environmentVariables[2]).isEqualTo("SecretKeyEcsExecDummy")
+                                assertThat(environmentVariables[3]).isEqualTo(dummyRegion)
+                                assertThat(environmentVariables[4]).isEqualTo(dummyRegion)
+                            }
+                        })
+                    }
```
abrooksv, line 131: Do these actually fail the test? Or is this thrown on a different thread so the test ignores it?

Could instead use a CapturingProcessAdaptor and then make the asserts on the outputs

May want a CountDownLatch and an await after the execute to make sure the test does not exit before the process exists too

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandRunProfile.kt
```diff
@@ -0,0 +1,94 @@
...
+    fun getCommandLine(): GeneralCommandLine? {
+        val exePath = path
+        val commandLine = GeneralCommandLine()
+        try {
+            commandLine.parametersList.addParametersString(parameters)
+            val exeFile = File(exePath)
+            if (exeFile.isDirectory && exeFile.name.endsWith(".app")) {
+                commandLine.exePath = "open"
+                commandLine.parametersList.prependAll("-a", exePath)
+            } else {
```
abrooksv, line 58: Do we need this?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/ecs/exec/RunCommandDialogTest.kt
manodnyab: Hadn't changed the region in the credentials, so they failed and were not in the same thread so the tests were displayed as passed. A countdown latch can be used with a specific count (5) and without the CapturingProcessAdapter

abrooksv (approved):

Issue #2610, opened 2021-05-24
Not able to view s3 bucket contents when profile doesn't have full ListObjects permission 


elenasch:
---
Toolkit: AWS Toolkit For JetBrains 1.26-211
OS: Mac OS X 10.15.7
IDE: IntelliJ IDEA Ultimate Edition 2021.1.1

My profile has access only to certain bucket in my organization, so with that limited access I am unable to use the AWS Explorer for S3.
It would be great to allow manually adding the limited resources I have access to so I can explore them in the UI. Otherwise I'm back to CLI

PR #2722, opened 2021-07-19
View bucket by entering bucket name/URI

manodnyab:
## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

## Description

Bucket can be opened by typing bucket name/URI

## Related Issue(s)

https://github.com/aws/aws-toolkit-jetbrains/issues/2610

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.intellij.openapi.ui.DialogWrapper
+import com.intellij.ui.layout.panel
+import javax.swing.JComponent
+
+class ViewResourceDialog(project: Project, val resourceType: String) : DialogWrapper(project) {
+    var resourceName = ""
+    private val component by lazy {
+        panel {
+            row("$resourceType:") {
+                textField(::resourceName).constraints(grow)
```
rli, line 16: should we have validation if string was empty?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
```diff
@@ -0,0 +1,26 @@
...
+            }
+        }
+    }
+
+    init {
+        super.init()
```
rli, line 22: should we set the title?

rli, line 22:
```suggestion
        init()
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    override fun createCenterPanel(): JComponent? = component
```
rli, line 25: probably drop the null?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/bucketActions/ViewBucketAction.kt
```diff
@@ -0,0 +1,26 @@
...
+import software.aws.toolkits.jetbrains.services.s3.S3ServiceNode
+import software.aws.toolkits.jetbrains.services.s3.openEditor
+import software.aws.toolkits.jetbrains.utils.notifyError
+import software.aws.toolkits.resources.message
+
+class ViewBucketAction : ViewResourceAction<S3ServiceNode>(message("action.aws.toolkit.s3.open.bucket.viewer.text"), message("s3.bucket.label")) {
+
+    override fun viewResource(resourceToView: String, selected: S3ServiceNode) {
+        try {
+            if (resourceToView.startsWith("S3://")) {
```
rli, line 17: needs to work with lowercase as well

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
manodnyab: Added validation for when string is empty and when string has just the prefix(in this case S3://)

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
manodnyab: done

manodnyab: Made the change

jetbrains-core/src/software/aws/toolkits/jetbrains/services/s3/bucketActions/ViewBucketAction.kt
manodnyab: Added the change to work in both cases

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
```diff
@@ -0,0 +1,29 @@
...
+                textField(::resourceName).constraints(grow).withErrorOnApplyIf("$resourceType must be entered") {
+                    it.text.isNullOrBlank() || it.text.equals(prefix, ignoreCase = true)
+                }
+            }
+        }
+    }
+
+    init {
+        init()
+        title = actionTitle
```
rli, line 25: i think title needs to be set before init

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/ViewResourceAction.kt
```diff
@@ -0,0 +1,22 @@
...
+
+package software.aws.toolkits.jetbrains.core.explorer.actions
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.project.DumbAware
+import software.aws.toolkits.jetbrains.core.explorer.ViewResourceDialog
+import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerNode
+
+abstract class ViewResourceAction<in T : AwsExplorerNode<*>>(private val actionTitle: String, val resourceType: String, private val prefix: String) :
+    SingleExplorerNodeAction<T>(actionTitle), DumbAware {
```
rli, line 12: view resource implies `SingleResourceNodeAction`. is that what you intended?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/ViewResourceAction.kt
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.explorer.actions
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.project.DumbAware
+import software.aws.toolkits.jetbrains.core.explorer.ViewResourceDialog
+import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerNode
+
+abstract class ViewResourceAction<in T : AwsExplorerNode<*>>(private val actionTitle: String, val resourceType: String, private val prefix: String) :
```
rli, line 11: IMO this is a bit awkward since it seems written specifically for the s3 protocol prefix. is there a cleaner way to make this work?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/ViewResourceDialog.kt
manodnyab: Changed the order, but is there a particular reason for the order, since it works in both cases....

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/ViewResourceAction.kt
manodnyab: The action is based off the S3 service node not the buckets which is why used` SingleExplorerNodeAction` instead of `SingleResourceNodeAction`, but the action is to view a specific resource(here: bucket)

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/ViewResourceAction.kt
manodnyab: Changed the prefix type to String with default as null so it shouldn't affect any other uses of this class. Another alternative I considered was changing the resourceType sent from ViewBucketAction as S3 Bucket Name/URI and edit that, but that would not be a great choice, but since now the prefix is nullable, it should be better?

manodnyab: Added a function as a parameter instead of prefix to check different conditions for ensuring correct input text is entered

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/actions/ViewResourceAction.kt
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.explorer.actions
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.project.DumbAware
+import software.aws.toolkits.jetbrains.core.explorer.ViewResourceDialog
+import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerNode
+
+abstract class ViewResourceAction<in T : AwsExplorerNode<*>>(private val actionTitle: String, val resourceType: String, private val prefix: String? = null) :
```
rli, line 11: probably drop this val

manodnyab: oops, yes, done!

PR #2731, opened 2021-07-23
Fix ApplicationThreadPoolScope without disposable

rli:
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/ProjectLevelCoroutineScopeTracker.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/ProjectLevelCoroutineScopeTracker.kt
@@ -6,14 +6,14 @@ package software.aws.toolkits.jetbrains.core
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.project.Project
+import com.intellij.util.containers.ContainerUtil
 import software.aws.toolkits.jetbrains.utils.ApplicationThreadPoolScope
-import kotlin.reflect.KClass
 
 class ProjectLevelCoroutineScopeTracker(
-    @SuppressWarnings("UnusedProperty")
+    @Suppress("unused")
     private val project: Project
 ) : Disposable {
-    private val scopes: MutableMap<String, ApplicationThreadPoolScope> = mutableMapOf()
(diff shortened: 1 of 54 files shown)


manodnyab (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/ProjectLevelCoroutineScopeTracker.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.intellij.openapi.components.ServiceManager
+import com.intellij.openapi.project.Project
+import software.aws.toolkits.jetbrains.utils.ApplicationThreadPoolScope
+import kotlin.reflect.KClass
+
+class ProjectLevelCoroutineScopeTracker(
+    @SuppressWarnings("UnusedProperty")
+    private val project: Project
+) : Disposable {
+    private val scopes: MutableMap<String, ApplicationThreadPoolScope> = mutableMapOf()
```
abrooksv, line 16: Should this be a weak hash map?

Are the benefits to caching the scope?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/ProjectLevelCoroutineScopeTracker.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun dispose() { }
+
+    companion object {
+        fun getInstance(project: Project) = ServiceManager.getService(project, ProjectLevelCoroutineScopeTracker::class.java)
+    }
+}
+
+fun Project.applicationThreadPoolScope(coroutineName: String) = ProjectLevelCoroutineScopeTracker.getInstance(this).applicationThreadPoolScope(coroutineName)
+
+fun Project.applicationThreadPoolScope(clazz: KClass<out Any>) = ProjectLevelCoroutineScopeTracker.getInstance(this).applicationThreadPoolScope(clazz.java.name)
```
abrooksv, line 29: If we inline this we do not need to pass the class

```
inline fun <reified T : Any> T.applicationThreadPoolScope() = applicationThreadPoolScope(T::class.java.name)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/ProjectLevelCoroutineScopeTracker.kt
rli: changed to weak value map. re: caching, for our use case i think we can go either way

PR #2739, opened 2021-07-29
Create new ExecutableManager

abrooksv:
Heavy WIP, but wanted to get early feedback on API and see if anyone can poke holes in it.
Still need to migrate CLIs over...yes I will rename it to drop the "2" off everything 😄 

Concepts that differ from original:

Services
* Settings are moved to their own `applicationService` to keep logic cleaner and more clear
* Only store settings if its a custom path, auto-detected CLIs should **not** store path so we can re-detect it in future
* We no longer save the last version and when we checked it to disk

Versions
* We can check CLI version from any thread, if a UI thread a modal progress bar is shown
* `ExecutableType` had the concept of a version but supported versions was not in the base class. It is now moved up into the base class (Making `Validatable` no longer needed (we have no validation beyond versions today))
* VersionRange made a first class citizen
* ~~Support for multiple VersionRange is a first class citizen (**cough** AWS CLI **cough**)~~
* Support for more than SemVer versions (i.e. SSM is X.Y.Z.B format)

Getting an Exeutable
* No longer async, since we can do validation from any thread safely
* No longer returns a sealed class that must be `when'ed` on everywhere
* Given we can check CLIs from any thread, pre-loading the cache should no longer be needed which improves startup (`ExecutableLoader`). Needs to be validated post-SAM CLI migration

Open Questions
1. Should we have a higher level `getAndValidate`? How will this scale into `getAndValidateAndInstallIfNeeded`?
2. Should we wrap `GeneralCommandLine` and do validation for every invocation? Do we need to wrap it to do WSL correctly?
3. Auto-installing is still TBD
4. Right now we do **not** migrate users who manually configured CLIs before....should we given it is probably safe to assume most users now use brew and thus auto-discover is the intended mechanism?

TODO:
1. Create higher level method to turn `Compatibility` into error messages
2. Migrate SAM CLI to see growing pains
3. Way more tests
4. Don't leak `SemVer` into the API of `SemanticVersion` take the args manually
5. Write more docs 😄 

## License


jetbrains-core/resources/META-INF/plugin.xml
```diff
@@ -117,6 +117,7 @@ with what features/services are supported.
         </extensionPoint>
 
         <extensionPoint name="executable" interface="software.aws.toolkits.jetbrains.core.executables.ExecutableType" dynamic="true"/>
+        <extensionPoint name="executable2" interface="software.aws.toolkits.jetbrains.core.executables.ExecutableType2" dynamic="true"/>
```
abrooksv, line 120: Technically not needed anymore, ExecutableInstances2's should all be objects and stateless

TODO: Delete this...

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableSettings.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.intellij.openapi.components.RoamingType
+import com.intellij.openapi.components.SimplePersistentStateComponent
+import com.intellij.openapi.components.State
+import com.intellij.openapi.components.Storage
+import com.intellij.openapi.components.service
+import com.intellij.util.xmlb.annotations.Attribute
+import com.intellij.util.xmlb.annotations.Property
+import com.intellij.util.xmlb.annotations.Tag
+
+@State(name = "executables2", storages = [Storage("aws.xml", roamingType = RoamingType.DISABLED)])
```
abrooksv, line 16: I turned off Roaming, there is no reason we should sync executable paths across computers IMO

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableSettings.kt
```diff
@@ -0,0 +1,43 @@
...
+class ExecutableSettings : SimplePersistentStateComponent<ExecutableOptions>(ExecutableOptions()) {
+    fun getExecutablePath(executable: ExecutableType2<*>) = state.value[executable.id]?.executablePath
+
+    fun setExecutablePath(executable: ExecutableType2<*>, value: String?) {
+        if (value == null) {
+            state.value.remove(executable.id)
+        } else {
+            val original = state.value[executable.id] ?: ExecutableState2()
+            state.value[executable.id] = original.copy(executablePath = value)
```
abrooksv, line 25: We avoid mutating the state directly since BaseState is monitoring the map for puts/removals and upping the ModificationTracker for us. If we update state in place, we take that responsibility.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableManager2.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        val baseVersionCompatability = isVersionValid(version, executable.type.supportedVersions())
+        if (baseVersionCompatability != Compatability.Valid) {
+            return baseVersionCompatability
+        }
+
+        additionalStrictRange?.let {
+            return isVersionValid(version, additionalStrictRange)
+        }
```
abrooksv, line 79: Our use case for this is "this one feature needs 1.X+2, but rest of toolkit needs 1.X", so maybe this should actually be `additionalMinVersion: Version`?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableType.kt
```diff
@@ -32,25 +34,64 @@ interface ExecutableType<VersionScheme> {
...
+
+    /**
+     * Name of the executable for users, e.g. the marketing name of the executable
+     */
+    val displayName: String
+
+    /**
+     * List of supported [VersionRange]. An empty list means any version is supported
+     */
+    fun supportedVersions(): List<VersionRange<VersionScheme>> = emptyList()
```
rli, line 56: bit more work, but maybe `VersionRange` should support disjoin union ranges?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/Versions.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.intellij.util.text.SemVer
+
+interface Version : Comparable<Version> {
+    fun displayValue(): String
+}
+
+data class SemanticVersion(private val version: SemVer) : Version {
+    override fun displayValue(): String = version.toString()
+
+    override fun compareTo(other: Version): Int = version.compareTo((other as SemanticVersion).version)
```
rli, line 15: needs to handle other versions

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/Compatability.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.core.executables
+
+sealed class Compatability {
+    data class NotInstalled(val detailedMessage: String? = null) : Compatability()
+    data class VersionTooOld(val minVersion: Version) : Compatability()
+    data class VersionTooNew(val maxVersion: Version) : Compatability()
+    object Valid : Compatability()
```
rli, line 10: nit: `Validity` or something else

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableManager2.kt
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+        null
+    }
+
+    fun <T : Version> determineVersion(executable: Executable<ExecutableType2<T>>): T {
+        assertIsNonDispatchThread()
+
+        return when (val result = versionCache.getValue(executable)) {
+            is Result.Failure -> {
+                throw RuntimeException(message("failed to identify version of {0}", executable.type.displayName), result.reason)
```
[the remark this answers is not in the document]
abrooksv: ya, havent done any localization

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/Versions.kt
abrooksv: It is actually valid to throw class cast exception on compareTo. Theres no way to compare differing schemes any other way that I can think of

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/Compatability.kt
abrooksv: Instead of Valid or Compatability?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableVersionCache.kt
```diff
@@ -0,0 +1,57 @@
...
+                Result.Failure(e, lastModifiedTime).also {
+                    cache[executable] = it
+                } as Result<V>
+            }
+        }
+    )
+
+    private fun <T : Version> testExecutable(executable: Executable<ExecutableType2<T>>): Version = executable.type.determineVersion(executable.path)
+
+    private fun getLastModificationTime(path: Path): Long = Files.getLastModifiedTime(path).toMillis()
```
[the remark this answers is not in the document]
abrooksv: At least on Mac is 1 sec granularity.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableType.kt
```diff
@@ -32,25 +34,64 @@ interface ExecutableType<VersionScheme> {
...
      *
      * @return the resolved path or null if not found
-     * @throws if an exception occurred attempting to resolve the path, when success was expected
+     * @throws Exception if an exception occurred attempting to resolve the path
      */
     fun resolve(): Path?
 }
 
-interface Validatable {
+interface Manged : AutoResolvable {
```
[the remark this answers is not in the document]
abrooksv: Fixeedd

jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableManager2.kt
abrooksv: The reason I was hesitant was then it needs to worry about the max version

kiiadi:
While we're doing this can we separate the out the IDE specific logic into a generally applicable abstraction? Each IDE we build toolkits for needs to solve this problem, I think it makes sense to have abstractions that we can easily transfer between tools.

abrooksv:
> While we're doing this can we separate the out the IDE specific logic into a generally applicable abstraction? Each IDE we build toolkits for needs to solve this problem, I think it makes sense to have abstractions that we can easily transfer between tools.

I think we can make a new design doc that is IDE agnostic and covers these topics, but from a code standpint I am not very inclined to. Almost every method in ExecutableManager2 touches an API from the IDE. We can achieve the split using abstract classes and protected abstract methods, but I would argue that the additional cognitive load is not worth it. 
* ExectuableSettings is 100% tied to the IDE's State system
* ExecutableManager2 is tied to ProgressManager / Threading APIs
* ExecutableVersionCache is tied to the ProgressInidcator system.

abrooksv:
Migrated the UI for SAM Init:

* Executable browsing and test button can be reused
  * Error message is shown after check, or version if Valid
* Behavior was broken for the UI (if a runtime required a newer SAM version than min version, it was impossible to configure it) so we will always just show the selection now

Screen Shot 2021-08-13 at 2 40 52 PM

Screen Shot 2021-08-13 at 2 41 02 PM

Screen Shot 2021-08-13 at 2 41 13 PM

Screen Shot 2021-08-13 at 2 40 39 PM

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/SemanticVersion.kt
```diff
@@ -0,0 +1,31 @@
...
+            .thenBy { it.patch }
+
+        fun parse(version: String): SemanticVersion {
+            val parts = version.split(".")
+            if (parts.size != 3) {
+                throw IllegalArgumentException("$version not in the format of MAJOR.MINOR.PATCH")
+            }
+
+            try {
+                val patch = parts[2].substringBefore("-").toInt()
```
kiiadi, line 24: prly also need to substring before `+` too while we're at it, this is a valid semantic version: `1.2.2+12423`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolManager.kt
```diff
@@ -0,0 +1,113 @@
...
+        project: Project? = null,
+        tool: Tool<ToolType<T>>?,
+        stricterMinVersion: T? = null
+    ): Validity {
+        if (tool == null) {
+            return Validity.NotInstalled()
+        }
+
+//        return runUnderProgressIfNeeded(project, message("version.progress.title"), false) {
+        return runUnderProgressIfNeeded(project, "Validating CLI", false) {
```
kiiadi, line 82: localization?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolManager.kt
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Checks the [Validity] of the specified tool at the specified path. An optional stricter version can be specified to override the minimum version.
+     *
+     * If called on a UI thread, a modal dialog is shown while the validation is in progress to avoid UI lock-ups
+     */
+    fun <T : Version> validateCompatability(
+        project: Project? = null,
+        path: Path,
+        type: ToolType<T>,
+        stricterMinVersion: T? = null
```
kiiadi, line 64: I assume the `stricterMinVersion` is for the case where generally this tool is compatible, but certain features require a newer version?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/Versions.kt
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @return Human-readable representation of the version
+     */
+    fun displayValue(): String
+}
+
+/**
+ * @return true if the specified version is compatible with any of the specified version ranges. Always returns true if no ranges are specified.
+ */
+fun <T : Version> isVersionValid(version: T, ranges: List<VersionRange<T>>): Validity {
```
kiiadi, line 19: nit: could prly be an extension function, since `version: T` is the "subject"

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Name of the executable for users, e.g. the marketing name of the executable
+     */
+    val displayName: String
+
+    /**
+     * List of supported [VersionRange]. An empty list means any version is supported
+     */
+    fun supportedVersions(): List<VersionRange<VersionScheme>> = emptyList()
```
kiiadi, line 27: I saw your "ahem" in the PR description - by why do we need a list of ranges?

The CLI example - is that we have a supported range for CLI v1 and CLI v2 ?

kiiadi (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/SemanticVersion.kt
abrooksv: Fair, there is a TODO a few lines up for pre-release comparison but we should def do it for parse I think

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolManager.kt
abrooksv: Opps, I knew I forgot something...everything needs localization 😄

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
abrooksv: Yes, the AWS CLI is the only use case today, where ECS Exec has a min in the V1 branch and and a min in the v2 branch

abrooksv: We could use stricterMinVersion to gate it though. Thoughts?

```
val newMin = if(version.majorVersion == 1) SemanticVersion(1, ..) else SemanticaVersion(2...)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolPathSelector.kt
```diff
@@ -0,0 +1,78 @@
...
+            null,
+            null,
+            FileChooserDescriptorFactory.createSingleFileDescriptor(),
+        )
+
+        testButton.addActionListener {
+            val pathToCheck = getConfiguredPath() ?: autoDetectedPath
+            val label = when (val validity = checkExecutable(Paths.get(pathToCheck), type)) {
+                is Validity.Valid -> {
+                    JBLabel("<html>${type.displayName} version is ${validity.version.displayValue()}")
```
rli, line 37: unterminated `<html>`. are we only using html for the <br>?

abrooksv: Should not be html on this one, only error

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/Versions.kt
abrooksv: changed and renamed to `isValid`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/SemanticVersion.kt
abrooksv: Spec says its hyphen. (https://semver.org/#spec-item-9)

abrooksv:
Removed having multiple version ranges, if needed use stricterMinVersion.

Multiple breaks down when 2 unrelated features build upon the same tool making the ranges impracticable to reason about.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolPathSelector.kt
abrooksv: Fixed

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
abrooksv: Removed multiple ranges support

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/SemanticVersion.kt
abrooksv: Added

PR #2751, opened 2021-08-05
Always pass the IDE profile as a property, instead of having two ways

abrooksv:
## License


rli (approved):
also need to fix run configs

abrooksv:
> also need to fix run configs

Oh! Good catch.

rli (approved):

PR #2803, opened 2021-09-16
Redo our coroutines APIs to make more sense

abrooksv:
* There are now 3 scope levels, Project, Application, and Disposable.
  * Disposable is tied to Application as its parent
  * All 3 default to running jobs in background
  * All 3 will have an implicit parent disposable that will trigger when plugin is unloaded
* The old ApplicationThreadPoolScope has been made private to avoid confusion

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/coroutines/scopes.kt
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Coroutine scope that is tied to a disposable or a plugin unloading. Default to dispatching to background thread pool.
+ *
+ * Use this if the coroutine is tied to a UI such as a tool window, or a run configuration
+ *
+ * **Note: If a call lives past the closing of a UI such as kicking off a resource creation, use [projectCoroutineScope].
+ * Otherwise, the coroutine will be canceled when the UI is closed!**
+ */
+fun disposableCoroutineScope(disposable: Disposable, coroutineName: String): CoroutineScope =
```
rli, line 40: should we ban attaching to a project/application with this one?

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/coroutines/CoroutineUtilsTest.kt
```diff
@@ -0,0 +1,42 @@
...
+            assertThat(ApplicationManager.getApplication().isDispatchThread).isFalse
+            withContext(getCoroutineUiContext()) {
+                assertThat(ApplicationManager.getApplication().isDispatchThread).isTrue
+            }
+        }
+    }
+
+    @Test
+    fun `getCoroutineBgContext context runs not on UI thread`() {
+        runBlocking {
```
rli, line 36: what about running on edt instead?

rli (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/coroutines/CoroutineUtilsTest.kt
abrooksv: as in start it on EDT? ya lets do that

jetbrains-core/src/software/aws/toolkits/jetbrains/core/coroutines/scopes.kt
abrooksv: I debated that, maybe we should throw illegal argument so prevent that.

PR #2816, opened 2021-09-23
Fix weird issue with regions-only combo label

rli:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ConnectionSettingsMenus.kt
```diff
@@ -174,15 +174,17 @@ class SettingsSelectorComboLabel(private val selectorLogic: SettingsSelectorLogi
     private val label = JBLabel()
 
     init {
+        val arrowLabel = JBLabel(AllIcons.General.ArrowDown)
         add(label, BorderLayout.CENTER)
```
abrooksv, line 178: ah this should have been addToCenter, doesnt change anything but reads nicer IMO

abrooksv (approved):

PR #2823, opened 2021-09-24
Migrate off some deprecated IJ APIs

abrooksv:
* Remove old APIs that were added into the SDK
* Migrate off deprecated methods

## License


rli (commented):
do we still need `CellBuilder<DialogPanel>.installOnParent` now that we have `validatingPanel` and `nestedPanel`? should probably mark it as deprecated

abrooksv:
> do we still need `CellBuilder<DialogPanel>.installOnParent` now that we have `validatingPanel` and `nestedPanel`? should probably mark it as deprecated

ya, it was also rather hacky :D

Issue #1883, opened 2020-06-28
Compatibility with localstack


maranqz:
**Is your feature request related to a problem? Please describe.**
If we use [localstack](https://github.com/localstack/localstack) for correct using AWS Explorer  should change default endpoint.

**Describe the solution you'd like**
Adding options to set endpoints for localstack aws services.

<details>
<summary><code>Example log</code></summary>

```
WARN - tProjectAccountSettingsManager - Failed to switch to profile 'Profile:localstack' 
software.amazon.awssdk.services.sts.model.StsException: The security token included in the request is invalid. (Service: Sts, Status Code: 403, Request ID: 5cb0bde5-daf8-4dbe-bb77-f887f4d18c25)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handleErrorResponse(CombinedResponseHandler.java:123)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handleResponse(CombinedResponseHandler.java:79)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handle(CombinedResponseHandler.java:59)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handle(CombinedResponseHandler.java:40)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:40)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:30)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:42)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:77)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:39)
[...]

PR #2830, opened 2021-09-28
Adding hook to customize SDK client creation

kiiadi:
Will allow customizing of AWS service client creation via an extension point. A typical use-case might be overriding the endpoint for a service during testing; could also be used to enable things like LocalStack (requested here : https://github.com/aws/aws-toolkit-jetbrains/issues/1883)

A typical usage might look as follows:

in `plugin.xml`

```xml
<extensions defaultExtensionNs="aws.toolkit">
  <sdk.clientCustomizer implementation="MyDevEndpointCustomizer"/>
</extensions>
```

The implementation of `MyDevEndpointCustomizer` could be:

```kt
class MyDevEndpointCustomizer : AwsClientCustomizer {
  override fun customize(connection: ConnectionSettings, builder: AwsClientBuilder<*, *>) {
    if (builder is LambdaClientBuilder) {
      builder.endpointOverride(URI.create("http://localhost:8888"))
    }
  }
}
```

### DIFF
--- a/core/src/software/aws/toolkits/core/ToolkitClientManager.kt
+++ b/core/src/software/aws/toolkits/core/ToolkitClientManager.kt
@@ -5,6 +5,7 @@ package software.aws.toolkits.core
 
 import org.jetbrains.annotations.TestOnly
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
+import software.amazon.awssdk.awscore.client.builder.AwsClientBuilder
 import software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder
 import software.amazon.awssdk.core.SdkClient
 import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption
@@ -64,6 +65,11 @@ abstract class ToolkitClientManager {
     protected abstract fun getRegionProvider(): ToolkitRegionProvider
 
     /**
+     * Allow implementations to apply customizations to clients before they are built
+     */
+    protected open fun clientCustomizer(builder: AwsClientBuilder<*, *>) {}
+
+    /**
      * Calls [AutoCloseable.close] on all managed clients and clears the cache
      */
     protected fun shutdown() {
@@ -95,7 +101,8 @@ abstract class ToolkitClientManager {
         region = Region.of(region.id),
         credProvider = credProvider,
         userAgent = userAgent,
-        endpointOverride = endpointOverride
+        endpointOverride = endpointOverride,
+        clientCustomizer = ::clientCustomizer
     )
 
     companion object {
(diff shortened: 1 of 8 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt
```diff
@@ -63,3 +71,22 @@ inline fun <reified T : SdkClient> Project.awsClient(): T {
...
+ *   override fun customize(builder: AwsClientBuilder<*, *>) {
+ *     if (builder is LambdaClientBuilder) {
+ *       builder.endpointOverride(URI.create("http://localhost:8888"))
+ *     }
+ *   }
+ * }
+ * ```
+ */
+interface AwsClientCustomizer {
+    fun customize(builder: AwsClientBuilder<*, *>)
```
abrooksv, line 91: If we did want to enable localstack support, we probably need to pass in the connection settings (to get the region)

kiiadi: k - good call - like it

core/src/software/aws/toolkits/core/ToolkitClientManager.kt
```diff
@@ -132,7 +136,14 @@ abstract class ToolkitClientManager {
                         builder.endpointOverride(URI.create(it))
                     }
                 }
+                .apply(clientCustomizer)
                 .build() as T
         }
     }
 }
+
+fun <T : SdkClient> ToolkitClientManager.getClient(
```
abrooksv, line 145: We are probably getting to the point where we need to decide if we want to go all in on ConnectionSettings and kill off the older APIs.

abrooksv (approved):

core/src/software/aws/toolkits/core/ToolkitClientManager.kt
kiiadi: I believe we should - anywhere we pass the `region` / `credential` pair we should just replace it with `ConnectionSettings' 

This was the first step in that direction - the above will be a much bigger refactor (and can probably be done piecemeal)

abrooksv: Agree. It is larger than this PR

JorisVanEijden:
> How can we set custom endpoints now? i don't see any documentation for this change?

There is this: https://github.com/aws/aws-toolkit-jetbrains/blob/55c2adf0956384711bad848d2a1f19fb99756e97/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt#L82

 ```java
  class MyDevEndpointCustomizer : AwsClientCustomizer {
    override fun customize(connection: ConnectionSettings, builder: AwsClientBuilder<*, *>) {
      if (builder is LambdaClientBuilder && connection.region.id == "us-west-2") {
        builder.endpointOverride(URI.create("http://localhost:8888"))
      }
    }
  }
 ```

KBSkims:
Any updates?

PR #2834, opened 2021-09-28
Moving ConnectionSettings to core

kiiadi:
The `ConnectionSettings` class doesn't depend on anything IDE specific and thus should reside in `core` - this also enables us to hang the following #2830 off a `connectionSetting` too.

This depends on an update to the telemetry generator (coupling is awesome).


core/src/software/aws/toolkits/core/credentials/AwsCredentials.kt
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.core.credentials
```
abrooksv, line 4: Why move this, you also didn't move the test that went with them

kiiadi: There's nothing that requires the IDE; and the `ConnectionSettings.toEnvironment` depends on it - made sense to pull it back. This is the same pattern as with region - the `toEnvironment` methods are in the `..core.region` package.

Will move the tests.

abrooksv (approved):

PR #2838, opened 2021-09-29
Default to non-global service if service not found in metadata

kiiadi:


rli (approved):

core/tst/software/aws/toolkits/core/region/ToolkitRegionProviderTest.kt
```diff
@@ -0,0 +1,32 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.core.region
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import software.aws.toolkits.core.utils.test.aString
+
+class ToolkitRegionProviderTest {
```
abrooksv, line 10: A lot of tests for this are down in AwsRegionProviderTest FYI

abrooksv (approved):

core/tst/software/aws/toolkits/core/region/ToolkitRegionProviderTest.kt
kiiadi: moved

PR #2842, opened 2021-10-01
Check for request token validity

manodnyab:
## Description

If request token is invalid, we stop polling for the resource

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
@@ -13,6 +13,7 @@ import software.amazon.awssdk.services.cloudcontrol.CloudControlClient
 import software.amazon.awssdk.services.cloudcontrol.model.Operation
 import software.amazon.awssdk.services.cloudcontrol.model.OperationStatus
 import software.amazon.awssdk.services.cloudcontrol.model.ProgressEvent
+import software.amazon.awssdk.services.cloudcontrol.model.RequestTokenNotFoundException
 import software.aws.toolkits.core.ConnectionSettings
 import software.aws.toolkits.jetbrains.core.awsClient
 import software.aws.toolkits.jetbrains.core.coroutines.projectCoroutineScope
@@ -128,7 +129,7 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
 
     private fun getProgress() {
         var size = pendingMutations.size
-
+        var isValidToken = true
         while (size > 0) {
             val mutation = pendingMutations.remove()
             if (!mutation.status.isTerminal()) {
@@ -136,21 +137,11 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
                 val progress = try {
                     client.getResourceRequestStatus { it.requestToken(mutation.token) }
                 } catch (e: Exception) {
-                    e.notifyError(
-                        message(
-                            "dynamic_resources.operation_status_notification_title",
-                            mutation.resourceIdentifier ?: mutation.resourceType,
-                            mutation.operation.name.toLowerCase()
-                        ),
-                        project
-                    )
-                    DynamicresourceTelemetry.mutateResource(
-                        project,
-                        Result.Failed,
-                        mutation.resourceType,
-                        addOperationToTelemetry(mutation.operation),
-                        ChronoUnit.MILLIS.between(mutation.startTime, DynamicResourceTelemetryResources.getCurrentTime()).toDouble()
-                    )
+                    unableToGetProgress(e, mutation)
+                    null
+                } catch (e: RequestTokenNotFoundException) {
+                    isValidToken = false
+                    unableToGetProgress(e, mutation)
                     null
                 }
                 val updatedMutation = when (val event = progress?.progressEvent()) {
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
```diff
@@ -128,29 +129,19 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
 
     private fun getProgress() {
         var size = pendingMutations.size
-
+        var isValidToken = true
```
kiiadi, line 132: This is set outside the loop; but only once on the inside. Which means if there are a batch of updates then the first one that fails for an invalid token, will effectively cause the rest to also be dropped.

As an aside- I really, *really* don't like `var` - even locally - I will usually look for anyway to avoid this mutability because it can lead to issues that make the code hard to reason about / bugs that are difficult to find. (This is a classic example).

What if we have the `try` block return a `Pair<ProgressEvent?, Boolean>` and then use the `Boolean` variable can drive whether we put it back on the queue again - we can also use this boolean to not-requeue items that are in a "terminal" state. So the "success" block becomes.

```kt
val (progressEvent, shouldRequeue) = try {
  val progress = client.getResourceRequestStatus { .. }
  progress.progressEvent() to !progress.progressEvent().operationStatus().isTerminal()
} catch (e: RequestTokenNotFoundException) {
..
null to false
} catch (e: Exception) {
..
null to true
}
```

kiiadi (changes requested):
I found two bugs that could probably have been caught with a test - we can fix them and add the test cases?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
```diff
@@ -128,29 +129,19 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
...
-                        project
-                    )
-                    DynamicresourceTelemetry.mutateResource(
-                        project,
-                        Result.Failed,
-                        mutation.resourceType,
-                        addOperationToTelemetry(mutation.operation),
-                        ChronoUnit.MILLIS.between(mutation.startTime, DynamicResourceTelemetryResources.getCurrentTime()).toDouble()
-                    )
+                    unableToGetProgress(e, mutation)
```
kiiadi, line 140: We definitely do not want to mark all exceptions as a failure; it could be a temporary network blip - which we will retry next time around. They need to be handled differently.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
```diff
@@ -128,29 +129,19 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
...
         var size = pendingMutations.size
-
+        var isValidToken = true
         while (size > 0) {
             val mutation = pendingMutations.remove()
             if (!mutation.status.isTerminal()) {
                 val client = mutation.connectionSettings.awsClient<CloudControlClient>()
                 val progress = try {
                     client.getResourceRequestStatus { it.requestToken(mutation.token) }
                 } catch (e: Exception) {
```
kiiadi, line 139: Instead of multiple catch blocks you can have a single-top-level catch block and then use `when` to differentiate them:

```kt
} catch (e: Exception) {
  val shouldRequeue = when(e) {
    is RequestTokenNotFoundException -> {
      unableToGetProgress(e, mutation)
      false
    }
    else -> true
}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
```diff
@@ -128,29 +129,19 @@ internal class DynamicResourceUpdateManager(private val project: Project) {
...
-                    DynamicresourceTelemetry.mutateResource(
-                        project,
-                        Result.Failed,
-                        mutation.resourceType,
-                        addOperationToTelemetry(mutation.operation),
-                        ChronoUnit.MILLIS.between(mutation.startTime, DynamicResourceTelemetryResources.getCurrentTime()).toDouble()
-                    )
+                    unableToGetProgress(e, mutation)
+                    null
+                } catch (e: RequestTokenNotFoundException) {
```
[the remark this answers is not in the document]
manodnyab: Changed to catching all exception and checking separately for this one

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
manodnyab: Changed to Pair

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
manodnyab: The exception specific to GetResourceRequestStatus is handled in this case, in most other cases, the operation status will change to failed, which will remove them from the queue with the updated implementation

jetbrains-core/src/software/aws/toolkits/jetbrains/services/dynamic/DynamicResourcesUpdateManager.kt
manodnyab: yep, done

PR #2906, opened 2021-10-26
Removing dependency on AWS CLI 

kiiadi:
Moving ECS exec functionality to auto-resolved SSM Session Manager Plugin


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/exec/OpenShellInContainerDialog.kt
```diff
@@ -87,7 +90,13 @@ class OpenShellInContainerDialog(
     private fun runExecCommand(task: String) {
         try {
             val commandLine = EcsExecUtils.createCommand(project, connectionSettings, container, task, shell)
-            val ptyProcess = PtyProcess.exec(commandLine.getCommandLineList(null).toTypedArray(), commandLine.effectiveEnvironment, null)
+
+            val cmd = if (SystemInfo.isWindows) {
+                arrayOf("cmd", "/C", commandLine.commandLineString)
```
rli, line 95: `cmd.exe` might actually be preferred. i'm surprised that `PtyCommandLine` doesn't handle this for us though

rli (approved):

rli (approved):

PR #2918, opened 2021-11-03
Adding telemetry for installing tools

kiiadi:


jetbrains-core/resources/telemetryOverride.json
```diff
@@ -10,6 +10,12 @@
             "name": "experimentId",
             "type": "string",
             "description": "The id of the experiment being activated or deactivated"
+        },
+        {
+            "name": "toolId",
+            "type": "string",
+            "description": "The tool being installed",
+            "allowedValues": ["ssm-plugin", "sam-cli"]
```
rli, line 18: nit: `session-manager-plugin` and `aws-sam-cli`?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/DefaultToolManager.kt
```diff
@@ -181,13 +183,13 @@ class DefaultToolManager @NonInjectable internal constructor(private val clock:
...
                 override fun run(indicator: ProgressIndicator) {
-                    installTool(type, indicator)
+                    installTool(type, indicator, project)
                 }
             }
         )
     }
 
-    private fun <V : Version> installTool(type: ManagedToolType<V>, indicator: ProgressIndicator?): Tool<ToolType<V>> {
+    private fun <V : Version> installTool(type: ManagedToolType<V>, indicator: ProgressIndicator?, project: Project?): Tool<ToolType<V>> {
```
rli, line 192: nit: i feel like typically we try to put project first

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
```diff
@@ -47,6 +48,13 @@ interface AutoDetectableToolType<VersionScheme : Version> : ToolType<VersionSche
 }
 
 interface ManagedToolType<VersionScheme : Version> : ToolType<VersionScheme> {
+    override val id: String
+        get() = telemetryId.toString()
+
+    /**
+     * Used to identify this tool with the telemetry stack
+     */
+    val telemetryId: ToolId
```
rli, line 57: is it even worth differentiating between the two?

jetbrains-core/resources/telemetryOverride.json
```diff
@@ -33,6 +39,18 @@
                     "type": "experimentState"
                 }
             ]
+        },
+        {
+            "name": "aws_toolInstallation",
```
rli, line 44: future-proof by adding an automatic vs manual invocation?

rli (approved):

jetbrains-core/resources/telemetryOverride.json
kiiadi: +1

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
kiiadi: possibly not but I didn't want to leak the `ToolId` enum all the way through given most of the time all we need is a String ...

jetbrains-core/resources/telemetryOverride.json
kiiadi: what does that mean?

rli: we don't expose it now, but i assumed eventually we'd want to allow users to manually trigger install/update. can definitely tackle the telemetry definition for that when it's actually implemented

JadenSimon: I guess add two metadata flags: `automatic` and `updated`.
Though `passive` covers the manual vs. automatic right?

kiiadi: the primary question I'm trying to answer here - is how many automated installations fail - this is something we can (potentially fix)

rli: okay, we can keep as-is for now then

PR #2931, opened 2021-11-04
Fix Kotlin warning about not implementing getChildren

abrooksv:
## Description
Fixes the warnings that look like

```
Deprecated: Class 'SqsServiceNode' is not abstract and does not implement abstract base class member public abstract fun getChildren(): (MutableCollection<out AbstractTreeNode<*>!>..Collection<AbstractTreeNode<*>!>) defined in software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerNode
```
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/nodes/AwsExplorerServiceRootNode.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/nodes/AwsExplorerServiceRootNode.kt
@@ -18,6 +18,8 @@ abstract class AwsExplorerServiceRootNode(project: Project, service: AwsExplorer
     private val serviceId = service.serviceId
 
     abstract override fun displayName(): String
+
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/nodes/AwsExplorerServiceRootNode.kt
```diff
@@ -18,6 +18,8 @@ abstract class AwsExplorerServiceRootNode(project: Project, service: AwsExplorer
     private val serviceId = service.serviceId
 
     abstract override fun displayName(): String
+
+    override fun getChildren(): List<AwsExplorerNode<*>> = super.getChildren()
```
abrooksv, line 22: Super call is defined on the interface, so delegate it there

rli (approved):

PR #2952, opened 2021-11-09
Adding new configurable panel for external tools

kiiadi:
Screen Shot 2021-11-09 at 10 22 55 AM
Screen Shot 2021-11-09 at 10 23 26 AM


abrooksv:
Random comment looking at the screen shot, should we add single quotes the stdout of the process in the "not in format" error

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolPathSelector.kt
```diff
@@ -1,78 +0,0 @@
...
-import com.intellij.ui.components.panels.Wrapper
-import com.intellij.util.text.nullize
-import com.intellij.util.ui.components.BorderLayoutPanel
-import software.aws.toolkits.resources.message
-import java.awt.event.ActionListener
-import java.nio.file.Path
-import java.nio.file.Paths
-import javax.swing.JButton
-
-open class ToolPathSelector<T : ToolType<*>>(private val type: T, private val validationResultHolder: Wrapper) : BorderLayoutPanel() {
```
abrooksv, line 20: We probably want to keep this since we want to use it in the SAM project wizard, unless we find a better way to configure SAM on that screen

kiiadi: yeah i'm looking at SAM now - so might end up pulling this back in

abrooksv: In this PR, or another?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
```diff
@@ -0,0 +1,64 @@
...
+import software.aws.toolkits.jetbrains.ui.ValidatingPanel
+import software.aws.toolkits.resources.message
+import java.nio.file.Path
+
+class ToolConfigurable : BoundConfigurable(message("executableCommon.configurable.title")), SearchableConfigurable {
+    private val settings = ToolSettings.getInstance()
+    private val manager = ToolManager.getInstance()
+    private val panel by lazy {
+        ValidatingPanel(
+            disposable!!,
```
rli, line 21: add comment on why this is safe?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
```diff
@@ -0,0 +1,64 @@
...
+            panel {
+                ToolType.EP_NAME.extensionList.forEach { toolType ->
```
rli, line 23: this EP is dynamic, so does this get reconstructed every time the configurable is shown to the user?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
```diff
@@ -0,0 +1,64 @@
...
+                    row(toolType.displayName) {
+                        textFieldWithBrowseButton(
+                            getter = { settings.getExecutablePath(toolType) ?: "" },
```
rli, line 26: why don't we just null out the value instead? is blank string a special case that means something?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolPathSelector.kt
kiiadi: another

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
kiiadi: the getter has to return a `String` it's just for display, the `textFieldWithBrowseButton` is not nullable.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
kiiadi: great question - I should check the lifecycle of the Configurable

kiiadi:
Updated error handling
Screen Shot 2021-11-10 at 1 09 37 PM
Screen Shot 2021-11-10 at 1 19 13 PM
Screen Shot 2021-11-10 at 1 18 46 PM

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
abrooksv: Theres a way to make a configurable thats dynamic powered

abrooksv: https://github.com/JetBrains/intellij-community/blob/master/docs/configurable.md#dynamic

kiiadi: I've made sure it gets cleared and re-created

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolConfigurable.kt
kiiadi: did you one better and made it into an exception

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
```diff
@@ -32,6 +38,17 @@ interface ToolType<VersionScheme : Version> {
      * Returns the [Version] for the executable of this type located at the specified location
      */
     fun determineVersion(path: Path): VersionScheme
+
+    companion object {
+        internal val EP_NAME = ExtensionPointName.create<DocumentedToolType<Version>>("aws.toolkit.tool")
```
abrooksv, line 43: Seems a little odd to me that the EP is tied to the DocumentedToolType instead of ToolType, and a  `is` check is done in the configurable.

But I can see the appeal and easy to change if the need ever arises

abrooksv (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/ToolType.kt
kiiadi: Yeah I think there may be cases where we don't want to allow customers to override and then it doesn't need to be registered in the EP (and thus won't be configurable), however I do think we want docs in the event it is surfaced

jetbrains-core/src/software/aws/toolkits/jetbrains/core/tools/DefaultToolManager.kt
```diff
@@ -138,13 +130,21 @@ class DefaultToolManager @NonInjectable internal constructor(private val clock:
...
-                return Validity.VersionTooOld(stricterMinVersion)
+                return Validity.VersionTooOld(version, stricterMinVersion)
             }
         }
 
         return Validity.Valid(version)
     }
 
+    private fun detailedMessage(exception: Exception, tool: Tool<ToolType<*>>): String {
+        println("EXCEPTION: $exception (${exception.message})")
```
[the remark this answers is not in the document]
kiiadi: Urgh, good catch.

PR #2957, opened 2021-11-10
Tell IntelliJ to auto-gen code after Gradle sync

abrooksv:
## License

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,12 +1,15 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import org.jetbrains.gradle.ext.ProjectSettings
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -47,3 +50,20 @@ tasks.register("runIde") {
         throw GradleException("Use project specific runIde command, i.e. :jetbrains-core:runIde, :intellij:runIde")
     }
 }
+
+if (idea.project != null) { // may be null during script compilation
+    idea {
+        project {
+            settings {
+                taskTriggers {
+                    afterSync(":sdk-codegen:generateSdks")
```
rli, line 59: would be nice to mark the source sets as generated roots as well

rli (approved):

build.gradle.kts
abrooksv: Ya, still cant get that to work for some reason...

PR #2959, opened 2021-11-10
Make unmanaged clients a first class citizen

abrooksv:
## License

### DIFF
--- a/core/src/software/aws/toolkits/core/ToolkitClientManager.kt
+++ b/core/src/software/aws/toolkits/core/ToolkitClientManager.kt
@@ -12,6 +12,7 @@ import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption
 import software.amazon.awssdk.core.retry.RetryMode
 import software.amazon.awssdk.http.SdkHttpClient
 import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.utils.SdkAutoCloseable
 import software.aws.toolkits.core.credentials.ToolkitCredentialsProvider
 import software.aws.toolkits.core.region.AwsRegion
 import software.aws.toolkits.core.region.ToolkitRegionProvider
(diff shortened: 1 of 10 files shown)


jetbrains-core/tst/software/aws/toolkits/jetbrains/core/MockClientManager.kt
```diff
@@ -24,17 +25,18 @@ import kotlin.reflect.KClass
 class MockClientManager : AwsClientManager() {
     private data class Key(
         val clazz: KClass<out SdkClient>,
-        val region: AwsRegion? = null,
-        val credProviderId: String? = null
+        val region: String? = null,
```
rli, line 28: rename to `regionId`?

core/src/software/aws/toolkits/core/ToolkitClientManager.kt
```diff
@@ -55,89 +52,98 @@ abstract class ToolkitClientManager {
...
     }
 
+    private fun <T : SdkClient> createNewClient(sdkClass: KClass<T>, connection: ConnectionSettings): T = constructAwsClient(
+        sdkClass = sdkClass,
+        credProvider = connection.credentials,
+        region = Region.of(connection.region.id),
+    )
+
+    /**
+     * Constructs a new low-level AWS client whose lifecycle is **NOT** managed centrally. Caller is responsible for shutting down the client
```
rli, line 70: are you envisioning this only being used for short-lived clients? what would be the best way to deal with long-lived clients

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/MockClientManager.kt
abrooksv: Good call

core/src/software/aws/toolkits/core/ToolkitClientManager.kt
abrooksv: Long lived too, like Telemetry uses it and thats the life of the toolkit. It just means that the callee is responsible for calling close on the client, not the toolkit on dispose

rli (approved):

PR #2966, opened 2021-11-17
Enable ecs exec by default

manodnyab:
## Description

Enable ecs exec flag and disable cloud debug flag by default
 
## License


kiiadi (approved):
Looks like you need to fix some unit-tests to gate on the feature, checkout `software.aws.toolkits.jetbrains.utils.rules.ExperimentRule`

PR #2982, opened 2021-11-25
Add Cloud API actions to other explorer nodes

kiiadi:
Screen Shot 2021-11-24 at 8 50 26 PM

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/CloudWatchLogsServiceNode.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/CloudWatchLogsServiceNode.kt
@@ -12,6 +12,7 @@ import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerResourceNo
 import software.aws.toolkits.jetbrains.core.explorer.nodes.AwsExplorerServiceNode
 import software.aws.toolkits.jetbrains.core.explorer.nodes.CacheBackedAwsExplorerServiceRootNode
 import software.aws.toolkits.jetbrains.services.cloudwatch.logs.resources.CloudWatchResources
+import software.aws.toolkits.jetbrains.services.dynamic.explorer.actions.CloudApiResource
 import software.aws.toolkits.resources.message
 
 class CloudWatchLogsServiceNode(project: Project, service: AwsExplorerServiceNode) : CacheBackedAwsExplorerServiceRootNode<LogGroup>(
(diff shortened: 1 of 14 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/EcsExplorerNodes.kt
```diff
@@ -53,16 +62,38 @@ class EcsClusterNode(project: Project, private val clusterArn: String) :
...
+}
+
+class ViewTaskDefinitionAction(private val node: EcsServiceNode) : AnAction(message("dynamic_resources.view_configuration_action_title", "Task Definition")) {
+    override fun actionPerformed(e: AnActionEvent) {
+        val identifier = DynamicResourceIdentifier(node.nodeProject.getConnectionSettingsOrThrow(), "AWS::ECS::TaskDefinition", node.service.taskDefinition())
+        DynamicResourceFileManager.getInstance(node.nodeProject).openEditor(identifier)
+    }
+}
+
+class EcsTaskDefinitionActionContributor : AwsExplorerActionContributor {
```
manodnyab, line 93: Could this action be added to the action group directly through `plugin.xml`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cloudwatch/logs/CloudWatchLogsServiceNode.kt
```diff
@@ -42,4 +44,8 @@ class CloudWatchLogsNode(
     override fun onDoubleClick() {
         CloudWatchLogWindow.getInstance(nodeProject).showLogGroup(logGroupName)
     }
+
+    override val typeDisplayName = "Log Group"
+    override val cloudApiResourceType = "AWS::Logs::LogGroup"
```
manodnyab, line 49: Should we maintain a separate list of all the resources we support that coincide with the ones in the manifest so it becomes easier to track if this resource type has been removed/added when we update the manifest?

kiiadi: wouldn't that just be implementors of the `CloudApiResource` interface?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/ecs/EcsExplorerNodes.kt
kiiadi: Could be - but I wanted `View Service Configuration` and `View Task Definition Configuration` to appear near each other (next to each other).

PR #2988, opened 2021-12-01
Attempt to fix dotnet test flakes

rli:
## License

### DIFF
--- a/jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/RunConfigTestUtils.kt
+++ b/jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/RunConfigTestUtils.kt
@@ -74,6 +74,16 @@ fun executeRunConfiguration(runConfiguration: RunConfiguration, executorId: Stri
             runConfiguration.project.messageBus.connect(executionEnvironment).subscribe(
                 ExecutionManager.EXECUTION_TOPIC,
                 object : ExecutionListener {
+                    override fun processTerminating(executorId: String, env: ExecutionEnvironment, handler: ProcessHandler) {
+                        println("Execution topic listener called processTerminating for $handler")
+                        super.processTerminating(executorId, env, handler)
+                    }
+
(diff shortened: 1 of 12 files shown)


jetbrains-core/tst/software/aws/toolkits/jetbrains/utils/RunConfigTestUtils.kt
```diff
@@ -59,7 +59,13 @@ fun executeRunConfiguration(runConfiguration: RunConfiguration, executorId: Stri
             val executionEnvironment = executionEnvironmentBuilder.build()
 
             val listener = object : OutputListener() {
+                override fun processWillTerminate(event: ProcessEvent, willBeDestroyed: Boolean) {
+                    println("Received processWillTerminate for ${event.processHandler}")
```
manodnyab, line 63: Do we need to log both processWillTerminate and processTerminated?

rli: trying to understand the lifecycle here and if it's possible for one to be called without the other

jetbrains-rider/tst/base/AwsReuseSolutionTestBase.kt
```diff
@@ -45,6 +49,15 @@ abstract class AwsReuseSolutionTestBase : BaseTestWithSolutionBase() {
         allowCustomDotnetRoots()
     }
 
+    @BeforeClass(alwaysRun = true)
```
manodnyab, line 52: Why do we provide this flag explicitly(in the previous After tag as well)? Is the default false?(that's surprising)

manodnyab (approved):

jetbrains-rider/tst/base/AwsReuseSolutionTestBase.kt
rli: testng default behavior is to skip the after* methods if the test fails. for the before* methods this forces the method to be run regardless of what test group it's in

PR #3012, opened 2021-12-14
Migrate to JUnit5 test runner with vintage test engine

rli:
* Switch to JUnit Platform runner with JUnit Vintage engine
* Migrate single test to JUnit5 to test SxS execution

## License


core/build.gradle.kts
```diff
@@ -26,4 +26,10 @@ dependencies {
     api(libs.bundles.jackson)
 
     testImplementation(libs.junit4)
```
breedloj, line 28: Is it easy to mix tests written against both the 4 and 5 APIs within the same module as we migrate? We should set up a policy, automated if possible, that new tests be written against the Jupiter API.

rli: yup, the Redshift test I migrated over to 5 is an example of that working (everything else in that package is junit4). it should be relatively straightforward to use detekt enforce usage of 5 in new test _classes_, but we should spend some time writing compatibility shims for our junit rules before enforcing that policy

breedloj (approved):

PR #3014, opened 2021-12-17
Display Enable/ Disable Execute Command option even if Cloud Debug is enabled

manodnyab:
## Description

Currently, if Cloud debugging is enabled for a service, the option to Enable/Disable ECS Exec/ Run Commands in Container/Open Shell is not visible. This change ensures that these actions are visible in the explorer irrespective of whether cloud-debug is enabled or not for the service. However, a warning is displayed to disable cloud debugging when any of these actions are selected and cloud-debugging for that particular service is enabled...

## Motivation and Context

Improving the visibility of Exec

## License


kiiadi:
screen shot?

is this required?

manodnyab:
> screen shot?
> 
> what's the driver that this is required?

Currently if cloud debug is already enabled for a service the option for enabling command execution is not available in the explorer... so this change might increase the visibility for ECS Exec...

PR #3022, opened 2022-01-05
Enforce LazyLog rule on Logger type instead of variable name

rli:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/docker/ToolkitDockerAdapter.kt
```diff
@@ -81,7 +81,7 @@ abstract class AbstractToolkitDockerAdapter(protected val project: Project, prot
         // unfortunately no callbacks available to grab the Deployment instance
         val deploymentPromise = AsyncPromise<Deployment>()
         serverConnection.deploy(task) { deploymentName ->
-            EcrUtils.LOG.debug("Retrieving Deployment associated with '$deploymentName'")
```
breedloj, line 84: Whoops

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ mockito = "4.0.0"
 telemetryGenerator = "1.0.25"
 testLogger = "3.1.0"
 testRetry = "1.2.1"
+slf4j = "1.7.32"
```
breedloj, line 20: They seem to be up to 1.7.36 in the interim - might as well rev to the latest while we're at it.
```suggestion
slf4j = "1.7.36"
```

breedloj (approved):

Issue #3041, opened 2022-01-26
Deploy Function Fails: Failed to locate requirements.txt


bergethan:
Since updating to 1.37-212, I am unable to deploy lambda functions on macOS. An IDE Fatal Error is thrown:

`java.lang.IllegalStateException: Failed to locate requirements.txt
	at software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaHandlerResolver.validateHandlerPath(PythonLambdaHandlerResolver.kt:79)
	at software.aws.toolkits.jetbrains.services.lambda.python.PythonLambdaHandlerResolver.findPsiElements(PythonLambdaHandlerResolver.kt:44)
	at software.aws.toolkits.jetbrains.services.lambda.Lambda.findPsiElementsForHandler(Lambda.kt:49)
	at software.aws.toolkits.jetbrains.ui.HandlerPanel.validateHandler(HandlerPanel.kt:107)
	at software.aws.toolkits.jetbrains.services.lambda.upload.UpdateFunctionCodePanel.validatePanel(UpdateFunctionCodePanel.kt:65)
	at software.aws.toolkits.jetbrains.services.lambda.upload.UpdateFunctionCodeDialog.doValidate(UpdateFunctionCodeDialog.kt:55)
	at com.intellij.openapi.ui.DialogWrapper.doValidateAll(DialogWrapper.java:376)
	at com.intellij.openapi.ui.DialogWrapper$OkAction.doAction(DialogWrapper.java:1901)
	at com.intellij.openapi.ui.DialogWrapper$DialogWrapperAction.actionPerformed(DialogWrapper.java:1868)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at java.desktop/javax.swing.plaf.basic.BasicRootPaneUI$Actions.actionPerformed(BasicRootPaneUI.java:254)
	at java.desktop/javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1810)
[...]

PR #3052, opened 2022-01-31
Fix requirements.txt detection

rli:
#3041
 
## License


vchikoti1998 (commented):
What we provide in the requirements.txt file. How it will be use for deploying lambda functions. Can you provide any resources for this task will be appreciated.

manodnyab (approved):

Issue #3056, opened 2022-02-02
exception while debugging lambda - WebStrom


rb2010:
**ToolKit version - 1.37-213
IDE- Webstrom version - WebStorm 2021.3.1
Build #WS-213.6461.79, built on December 28, 2021**

===========
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class software.aws.toolkits.core.lambda.LambdaArchitecture (java.lang.String is in module java.base of loader 'bootstrap'; software.aws.toolkits.core.lambda.LambdaArchitecture is in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @2daff561)
	at software.aws.toolkits.jetbrains.services.lambda.execution.local.LocalLambdaRunSettingsEditor.applyEditorTo(LocalLambdaRunSettingsEditor.kt:67)
	at software.aws.toolkits.jetbrains.services.lambda.execution.local.LocalLambdaRunSettingsEditor.applyEditorTo(LocalLambdaRunSettingsEditor.kt:15)
	at com.intellij.openapi.options.SettingsEditor.applyTo(SettingsEditor.java:101)
	at com.intellij.execution.impl.ConfigurationSettingsEditor$ConfigToSettingsWrapper.applyEditorTo(ConfigurationSettingsEditor.java:316)
	at com.intellij.execution.impl.ConfigurationSettingsEditor$ConfigToSettingsWrapper.applyEditorTo(ConfigurationSettingsEditor.java:294)
	at com.intellij.openapi.options.SettingsEditor.applyTo(SettingsEditor.java:101)
	at com.intellij.openapi.options.CompositeSettingsEditor.applyEditorTo(CompositeSettingsEditor.java:56)
	at com.intellij.openapi.options.SettingsEditor.applyTo(SettingsEditor.java:101)
	at com.intellij.execution.impl.ConfigurationSettingsEditor.getSnapshot(ConfigurationSettingsEditor.java:221)
	at com.intellij.execution.impl.ConfigurationSettingsEditorWrapper.getSnapshot(ConfigurationSettingsEditorWrapper.java:148)
	at com.intellij.execution.impl.ConfigurationSettingsEditorWrapper.getSnapshot(ConfigurationSettingsEditorWrapper.java:31)
	at com.intellij.execution.impl.SingleConfigurationConfigurable.createSnapshot(SingleConfigurationConfigurable.java:396)
[...]

PR #3057, opened 2022-02-02
Fix Lambda run configuration exception while setting handler architec…

rli:
…ture (#3056)
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunSettingsEditor.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/execution/local/LocalLambdaRunSettingsEditor.kt
@@ -8,6 +8,7 @@ import com.intellij.openapi.options.SettingsEditor
 import com.intellij.openapi.project.Project
 import com.intellij.util.PathMappingSettings
 import software.aws.toolkits.jetbrains.services.lambda.execution.registerConfigValidationListeners
+import software.aws.toolkits.jetbrains.services.lambda.execution.validateSupportedArchitecture
 import software.aws.toolkits.jetbrains.services.lambda.validation.LambdaHandlerValidator
 import software.aws.toolkits.jetbrains.utils.ui.selected
 import javax.swing.JComponent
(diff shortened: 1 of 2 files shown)


manodnyab (approved):

pradt2:
Hi, I tested this change, and unfortunately it does not solve https://github.com/aws/aws-toolkit-jetbrains/issues/3056 in my case.

To get the plugin to work, I had to change `LocalLambdaRunSettingsEditor.kt:67` to

```kotlin
configuration.architecture(LambdaArchitecture.fromValue(view.rawSettings.architecture.selected().toString()))
```

I suspect this change isn't a proper fix and the real issue lies elsewhere, but I hope this will point you in the right direction. I'm happy to do more testing for you if needed.

rli:
@pradt2 I'm unable to reproduce the error with my current proposed change at `LocalLambdaRunSettingsEditor.kt:42`

The `validateSupportedArchitecture()` call should ensure that `view.rawSettings.architecture.selected()` later returns a valid `LambdaArchitecture` and therefore not require your proposal for L67. Though, I suppose there's no harm in adding a defensive check here. We can also afford to be a bit hacky here since the UI is due to be rewritten.

> To get the plugin to work

Are you referring to the ClassCastException, or debugging in general? Can you share the XML serialization of the run configuration that's giving you issues? It should be in `<workspace>/.idea/workspace.xml` and should look like the following:

```
    <configuration name="[Local] app.lambdaHandler" type="aws.lambda" factoryName="Local" nameIsGenerated="true">
      <option name="credentialProviderId" value="profile:default" />
      <option name="regionId" value="us-west-2" />
      <option name="debugHost" value="localhost" />
      <option name="architecture" value="x86_64" />
      <option name="environmentVariables">
        <map />
      </option>
      <option name="handler" value="app.lambdaHandler" />
      <option name="image" value="false" />
      <option name="logicalFunctionName" />
      <option name="memorySize" value="128" />
      <option name="pathMappings">
        <list />
      </option>
      <option name="platform" />
      <option name="runtime" value="python3.9" />
      <option name="templateFile" />
      <option name="timeout" value="300" />
      <option name="useTemplate" value="false" />
      <option name="input" value="&quot;&quot;" />
      <option name="inputIsFile" value="false" />
      <sam>
        <option name="additionalBuildArgs" value="" />
        <option name="additionalLocalArgs" value="" />
        <option name="buildInContainer" value="false" />
        <option name="dockerNetwork" value="" />
        <option name="skipImagePull" value="false" />
      </sam>
      <method v="2" />
    </configuration>
```

pradt2:
Hey, I checked everything again, and it turns out I was building from the main branch of your fork... When I switched to the proper branch, the issue went away.

rli:
@pradt2 Awesome, thanks for taking the time to check!

PR #3071, opened 2022-02-21
Fix ARN reference contributor returning duplicate entries

rli:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/federation/psireferences/ArnPsiReferenceContributor.kt
```diff
@@ -17,9 +18,15 @@ class ArnPsiReferenceContributor : PsiReferenceContributor() {
             ) {
                 override fun accepts(o: Any?, context: ProcessingContext): Boolean {
                     if (o == null || o !is PsiElement) return false
+                    val manipulator = ElementManipulators.getManipulator(o)
+                    if (manipulator == null) return false
 
-                    if (o.text.contains("arn:")) {
-                        return true
+                    try {
+                        if (manipulator.getRangeInElement(o).substring(o.text).contains("arn:")) {
```
breedloj, line 25: Can we not just retrieve the `TextRange` directly off the `PsiElement`?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/federation/psireferences/ArnPsiReferenceContributor.kt
```diff
@@ -17,9 +18,15 @@ class ArnPsiReferenceContributor : PsiReferenceContributor() {
...
+                            return true
+                        }
+                    } catch (e: Exception) {
+                        return false
                     }
```
breedloj, line 30: Avoidable by checking the element's text length up front?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/federation/psireferences/ArnPsiReferenceContributor.kt
rli: the manipulators are language-aware and will strip out text that is not relevant to us (for example, stripping quotes in literal declarations, or only no-oping if we're not looking at a leaf element in the YAML case)

jetbrains-core/src/software/aws/toolkits/jetbrains/services/federation/psireferences/ArnPsiReferenceContributor.kt
rli: will keep the try/catch, but add explicit element length check

breedloj (approved):

PR #3080, opened 2022-03-03
Updating Cloud Control API resource manifest

breedloj:


rli (approved):
we should figure out automation to make changelog entries for this

PR #3141, opened 2022-04-28
Add warning to indicate time delay in SQS queue deletion

manodnyab:
## Types of changes

## Description

Add warning to indicate time delay in SQS queue deletion

Screen Shot 2022-04-28 at 11 17 10 AM

 
## License


resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
```diff
@@ -1101,6 +1101,7 @@ sqs.delete.message.failed=Failed to delete {0, choice, 1#message|2#{0,number} me
 sqs.delete.message.partial_successs=Failed to delete some messages! {0, choice, 1#message|2#{0,number} messages} were deleted.
 sqs.delete.message.succeeded=Successfully deleted {0, choice, 1#1 message|2#{0,number} messages}.
 sqs.delete.queue.action=Delete Queue...
+sqs.delete.warning_text=Deletion of the queue may take more than a minute to be reflected
```
rli, line 1104: how about we make this more generic? i'm sure there's other resource deletions that we have where delete may take a while

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/DeleteResourceDialog.kt
```diff
@@ -33,6 +34,14 @@ class DeleteResourceDialog(
             row {
                 deleteResourceConfirmation(grow)
             }
+            createNoteOrCommentRow(
+                ComponentPanelBuilder.createCommentComponent(
+                    message("sqs.delete.warning_text"),
```
kiiadi, line 39: This is a generic delete component right? We should probably not put SQS specific logic in here, we should find a way to extend it so that other resource types can (optionally) provide information if required.

resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
manodnyab: Changed the message to be more generic and accept the resourceType as parameter

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/DeleteResourceDialog.kt
manodnyab: Changed it to a more generic format which could be used by multiple resource types and the information in the comment can be custom to the resourceType if required

jetbrains-core/src/software/aws/toolkits/jetbrains/services/sqs/actions/DeleteQueueAction.kt
```diff
@@ -23,4 +23,7 @@ class DeleteQueueAction : DeleteResourceAction<SqsQueueNode>(message("sqs.delete
         client.deleteQueue { it.queueUrl(selected.queueUrl) }
         project.refreshAwsTree(SqsResources.LIST_QUEUE_URLS)
     }
+
+    override val displayComment: Boolean = true
+    override val comment: String = message("resource.delete.warning_text", "queue")
```
rli, line 28: we don't need the boolean since the presence of a comment implies that it should be displayed

rli (approved):

Issue #3098, opened 2022-03-11
Resource cache fails to complete future exceptionally & throws unhandled coroutine exception


rli:
https://github.com/aws/aws-toolkit-jetbrains/blob/5c87fd9fe01511e995ed46c58925ba889bec6ffe/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt#L313-L334

L320 in DefaultAwsResourceCache can return a cached entry where the deferred value has completed exceptionally. As a result, `result.value.await()` (L325) throws, is caught by the surrounding block, and is unhandled due to the `deferred.value.getCompleted()` call on (L329).

Reproducable by doing a `cache.getResource()` call that throws the first time(successfully bubbles out to callee code), but the second call will trigger the bug

rli:
Released in 1.44

PR #3142, opened 2022-04-28
Fixed a few issues with exception handling in the resource cache

kiiadi:
## Types of changes

## Description

Previously exceptions were getting cached, and bubbled as unhandled exceptions on subsequent calls. This fixes the following:
- exceptions are no longer considered valid "current values" when evaluating whether to return a cached value
- exceptions are pruned from the cache on maintenance runs

## Related Issue(s)

#3098 

## Testing

Added new tests to cover missing scenarios.

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
+import org.jetbrains.annotations.VisibleForTesting
 import software.amazon.awssdk.core.SdkClient
 import software.aws.toolkits.core.ConnectionSettings
 import software.aws.toolkits.core.credentials.CredentialIdentifier
@@ -324,9 +325,9 @@ class DefaultAwsResourceCache(
                     try {
                         context.future.complete(result.value.await())
                     } catch (e: Throwable) {
-                        val deferred = currentValue
-                        if (context.useStale && deferred != null && deferred.value.isCompleted) {
-                            context.future.complete(deferred.value.getCompleted())
+                        val previousValue = currentValue
+                        if (context.useStale && previousValue != null && previousValue.value.isCompleted && !previousValue.value.isCompletedExceptionally) {
+                            context.future.complete(previousValue.value.getCompleted())
                         } else {
                             context.future.completeExceptionally(e)
                         }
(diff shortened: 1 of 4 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -340,21 +341,27 @@ class DefaultAwsResourceCache(
...
-                    exceededWeight -= value.weight
-                }
-            }
+            doRunCacheMaintenance()
         } finally {
             scheduleCacheMaintenance()
         }
     }
 
+    @VisibleForTesting
```
kiiadi, line 350: I do hate this - but for internal state like this I just can't justify injection ... I'm very conflicted about it though

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/AwsResourceCacheTest.kt
```diff
@@ -475,6 +480,50 @@ class AwsResourceCacheTest {
...
+        val latch = CountDownLatch(1)
+        whenever(mockResource.fetch(any(), any())).then {
+            latch.await(500, TimeUnit.MILLISECONDS)
+            throw RuntimeException("Boom")
+        }
+
+        val first = sut.getResource(mockResource, connectionSettings)
+        val second = sut.getResource(mockResource, connectionSettings)
+        latch.countDown()
+        assertThat(first).wait().hasException.withFailMessage { "" }
```
kiiadi, line 494: oops - gotta fix these messges

rli, line 494: assert that it's the same exception and not 2 different calls?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
rli: i think we can lint for this if that makes you feel better

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
kiiadi: lint for what? that it's only used in tests?

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/AwsResourceCacheTest.kt
kiiadi: added a verify that the mock only gets hit once

jetbrains-core/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
rli: yup, we have a number of `@VisibleForTesting` peppered around the code base but i'm not actually sure those are being respected

Issue #3105, opened 2022-03-17
sam local invoke working directory is home directory, not project directory


mattzuba:
**Describe the bug**
Attempting to locally invoke functions and the `build` step correctly runs within the project directory and utilizes the samconfig.toml file, but the `local invoke` command runs in my home directory.

**To reproduce**
1. Create a samconfig.toml file with default parameters for local_invoke that includes items such as debug=true or environment variables
2. Run a function by clicking the Run option next to the function in the template.yaml file
3. The options from the samconfig.toml file are not used.

** Alternative reproduction **
1. Create a simple bash script in place of the `sam` command that simply echo's the present working directory (this seemed to only work after an initial run to prevent some errors about a bad sam executable):
```shell
#!/bin/bash

echo $(pwd)
```
2. Run a function by clicking the Run option next to the function in the template.yaml file
3. The output of the run command shows that the build step ran in the project directory, but the invoke command runs in the home directory.

**Expected behavior**
The invoke command should be run from the project directory so that it can use the samconfig.toml file

**Your Environment**
- OS: Pop! OS 21.04
- JetBrains product: PHPStorm
- JetBrains product version: 2021.2.3
- AWS Toolkit version: 1.40-212
- SAM CLI version:1.40.1
- JVM/Python version: Java 8, Python 3.9

PR #3144, opened 2022-04-29
Add working directory to `sam invoke local`

JadenSimon:
## Types of changes

## Description

This should've fixed an issue with SAM CLI not detecting `samconfig.toml` for `sam local invoke`
But apparently SAM CLI treats the template path as the project root.
I think we can just get rid of using `--template` when provided by a user. Which would fix the issue (plus less opportunity for unexpected behavior).

## Related Issue(s)

#3105

## Checklist

## License


JadenSimon:
SAM CLI recommends using `--config-file` (https://github.com/aws/aws-sam-cli/issues/3431) but if we're going to use that parameter then we should expose it in the UI.

PR #3149, opened 2022-04-29
Suggest enabling edit resources

kiiadi:
## Types of changes

## Description

This change is two fold, one is a general framework feature, the other is that feature in use

* Framework feature - introduce a common way to suggest that experiments be enabled - this includes notification and also persisting the choice. Suggestions are suppressed for a period of a week (configurable), and can be suppressed forever ("Never show me again")
* Use the above feature to suggest the "edit resources" feature for the dynamic coverage functionality - when a new resource document is opened.

## Motivation and Context

Will aid discovery of experiments.

## Testing

New tests added.

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/experiments/ToolkitExperiment.kt
```diff
@@ -89,14 +137,40 @@ internal class ToolkitExperimentManager : PersistentStateComponent<ExperimentSta
...
 
 internal class ExperimentState : BaseState() {
+    // This represents whether an experiment is enabled or not, don't want to rename it as that will
+    // cause problems with any experiments already out there in the wild who've been persisted
+    // as 'value'
     @get:Property
     val value by map<String, Boolean>()
+
+    @get:Property
+    val nextSuggestion by map<String, Long>()
```
rli, line 171: can we keep this as Instant instead of converting to epoch millis?

jetbrains-core/tst/software/aws/toolkits/jetbrains/core/experiments/ToolkitExperimentManagerTest.kt
```diff
@@ -157,17 +164,141 @@ class ToolkitExperimentManagerTest {
...
         ExtensionTestUtil.maskExtensions(ToolkitExperimentManager.EP_NAME, listOf(experiment), disposableRule.disposable)
+        val now = Instant.now()
 
         val sut = ToolkitExperimentManager.getInstance()
 
+        sut.shouldPrompt(experiment, now)
+        sut.neverPrompt(permanentlySuppressed)
         experiment.setState(true)
         val serialized = serializeState("experiments", sut)
+        println(serialized)
```
rli, line 177: delete?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/experiments/ToolkitExperiment.kt
kiiadi: the `BaseState` serialization doesn't really support it out of the box - and the nice thing about this is we can represent "never" as `-1`

I can convert if you think necessary tho, the `-1` thing is just a convenience.

PR #3151, opened 2022-05-03
Add deprecation notice for 211

rli:
## License


manodnyab (approved):

kiiadi:
I feel like there was a ton of "remove with min 212" stuff - can we kill that now?

rli:
> I feel like there was a ton of "remove with min 212" stuff - can we kill that now?

yeah but we usually do that after the deprecation notice is released

kiiadi:
> yeah but we usually do that after the deprecation notice is released

ah yes - this is just the notice - my bad.

PR #3153, opened 2022-05-05
Fix win dotnet tooling too new

rli:
## License


kiiadi (approved):
Don't really know what this does but ... sure.

rli:
> Don't really know what this does but ... sure.

CodeBuild updated their Windows base images (finally), but as a consequence 2021.1 is complaining because it doesn't understand that dotnet6 exists

PR #3213, opened 2022-07-13
CodeWhisperer percentage code completion implementation

Will-ShaoHua:
## Motivation and Context
We added functionality percentage of code modification for every Consolas Accepted recommendation. This metric give us insights on Codewhisperer accuracy. The Codewhisperer team is interested in the metric on percentage of code written by Codewhisperer after the user has enabled for improving our generated code.

## Solution
* After enabling the TOS and user start typing code then the timer start checking current time and iterates for every 5 minutes
* Calculates the user and Codewhisperer accepted metrics based on each character(token)
* On each time interval of 5 minutes it check and calculates the percentage of Codewhisperer code coverage
* Send it as new telemetry event with metrics
* Clear the tokens count for each interval with reset of start time

## Types of changes

## Description

This PR aims to record a new telemetry event: `codewhisperer_codePercentage`  based on a 5 min time window,
where metric `codewhispererPercentage = CodeWhisperer tokens / Total tokens`.

## Testing
Unit test: `CodeWhispererCodeCoverageTrackerTest`

Funtional test: `CodeWhispererTelemetryTest ` (test cases added)
## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -18,6 +21,18 @@ class CodeWhispererEditorListener : EditorFactoryListener {
             object : DocumentListener {
                 override fun documentChanged(event: DocumentEvent) {
                     CodeWhispererInvocationStatus.getInstance().documentChanged()
+                    // When open a file for the first time, IDE will also emit DocumentEvent for loading with `isWholeTextReplaced = true`
+                    // Added this condition to filter out those events
```
abhikuhikar, line 25: I tried debugging but this wasn't the case for me. Can you make sure if you are filtering this correctly?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,150 @@
...
+                    if (states.requestContext.fileContextInfo.programmingLanguage.toCodeWhispererLanguage() != language) return
+                    pushAcceptedTokens(remainingRecomm)
+                }
+            }
+        )
+        scheduleCodeWhispererTracker()
+    }
+
+    fun documentChanged(event: DocumentEvent) {
+        pushTotalTokens(event.newFragment)
```
abhikuhikar, line 67: This only covers the case when the new text is *added* to the document. 
You may also want to consider the other `DocumentEvent` cases such as:
1. Removed the existing text (You can use `event.oldLength`)
2. Replaced the existing text (You need to use old length and new length.)

Or to generalize you can just add (`event.newLength - event.oldLength`) number of tokens to the `totalTokens` to cover all the cases. Although this won't work as the user may replace the text recommended by CodeWhisperer. 
I think this logic is much more involved and needs to be taken into consideration many other cases.

abhikuhikar (changes requested):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,150 @@
...
+                return
+            }
+            emitCodeWhispererCodeContribution()
+        } finally {
+            init()
+            scheduleCodeWhispererTracker()
+        }
+    }
+
+    private fun scheduleCodeWhispererTracker() {
```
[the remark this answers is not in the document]
Will-ShaoHua: ack

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererCodeCoverageTrackerTest.kt
```diff
@@ -0,0 +1,265 @@
...
+        assertThat(pythonInstance == pythonInstance2).isTrue
+
+        CodeWhispererCodeCoverageTracker.getInstance(CodewhispererLanguage.Javascript)
+        assertThat(CodeWhispererCodeCoverageTracker.instances).hasSize(3)
+    }
+
+    @Test
+    fun `test tracker is listening to document changes and increment totalTokens - add new code`() {
+        val pythonTracker = spy(TestCodePercentageTracker(TWO_SECONDS, CodewhispererLanguage.Python, AtomicInteger(0), AtomicInteger(0)))
+        CodeWhispererCodeCoverageTracker.instances[CodewhispererLanguage.Python] = pythonTracker
```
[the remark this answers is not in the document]
Will-ShaoHua: I think we can't mock a companion object (correct me if I am wrong), but I can make it private and make another exposed companion method and add `@TestOnly` and `assert(ApplicationManager.getApplication().isUnitTestMode)` to hide this.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,150 @@
...
+    override fun dispose() {
+        if (isShuttingDown.getAndSet(true)) {
+            return
+        }
+        flush()
+    }
+
+    companion object {
+        const val FIVE_MINS_IN_SECS = 300L
+        internal val instances: MutableMap<CodewhispererLanguage, CodeWhispererCodeCoverageTracker> = mutableMapOf()
```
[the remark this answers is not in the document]
Will-ShaoHua: ack

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    override fun dispose() {
+        if (isShuttingDown.getAndSet(true)) {
+            return
+        }
+        flush()
+    }
+
+    companion object {
+        const val FIVE_MINS_IN_SECS = 300L
```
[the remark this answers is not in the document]
Will-ShaoHua: ack

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
Will-ShaoHua: image (1)

Will-ShaoHua: ^^ is what I get when I just open the IDE sandbox on loading the codewhisperer welcome.md file.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,143 @@
...
+                newTracker
+            }
+            else -> instance
+        }
+
+        @TestOnly
+        fun getInstancesMap(): MutableMap<CodewhispererLanguage, CodeWhispererCodeCoverageTracker> {
+            assert(ApplicationManager.getApplication().isUnitTestMode)
+            return instances
+        }
```
abhikuhikar, line 134: nit: Or you can just do `private set` on `val instances`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,143 @@
...
+    fun documentChanged(event: DocumentEvent) {
+        addAndGetTotalTokens(event.newLength - event.oldLength)
+    }
+
+    private fun flush() {
+        try {
+            if (isTelemetryEnabled()) emitCodeWhispererCodeContribution()
+        } finally {
+            scheduleCodeWhispererCodeCoverageTracker()
+            init()
```
abhikuhikar, line 64: You may want to call `init()` before `scheduleCodeWhispererCodeCoverageTracker ` as `scheduleCodeWhispererCodeCoverageTracker ` can technically throw an exception and `init()` may never get called.

abhikuhikar (commented):
Just couple comments. Rest LGTM.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: oh yea good point, on it.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: but Abhilash, with `val instances: MutableMap<T,E>` , we already have no setter on this static field. And for testing, we only need its getter, so the easiest way I think it to provide an api to get this map but restricted to unittest mode.

abhikuhikar (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererConstants.kt
```diff
@@ -37,6 +37,7 @@ object CodeWhispererConstants {
     const val TOTAL_BYTES_IN_KB = 1024
     const val TOTAL_BYTES_IN_MB = 1024 * 1024
     const val TOTAL_MILLIS_IN_SECOND = 1000
+    const val TOTAL_SECONDS_IN_MINUTE: Long = 60
```
rli, line 40:
```suggestion
    const val TOTAL_SECONDS_IN_MINUTE = 60
```
be consistent with providing types

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -594,5 +594,5 @@ interface CodeWhispererUserActionListener {
     fun navigatePrevious(states: InvocationContext) {}
     fun navigateNext(states: InvocationContext) {}
     fun beforeAccept(states: InvocationContext, sessionContext: SessionContext) {}
-    fun afterAccept(states: InvocationContext, sessionContext: SessionContext) {}
+    fun afterAccept(states: InvocationContext, sessionContext: SessionContext, remainingRecomm: String) {}
```
rli, line 597: why not `remainingRecommendation`?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
abhikuhikar: That was my miss. nvm.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
Will-ShaoHua: ack

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererConstants.kt
Will-ShaoHua: ack

Will-ShaoHua:
'/retryBuilds'

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -18,6 +21,18 @@ class CodeWhispererEditorListener : EditorFactoryListener {
...
+                    if (!event.isWholeTextReplaced) {
```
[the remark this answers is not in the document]
andrewyuq: question on this condition:
but what if user just replaced the whole text for example using their clipboard context? Do we still ignore them?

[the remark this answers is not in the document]
andrewyuq: If this case is property handled(i.e. do nothing) in `CodeWhispererCodeCoverageTracker` then we don't need this condition.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private fun flush() {
+        try {
+            if (isTelemetryEnabled()) emitCodeWhispererCodeContribution()
+        } finally {
+            init()
+            scheduleCodeWhispererCodeCoverageTracker()
+        }
+    }
```
andrewyuq, line 66: I am confused by this function -- why do we call init() here?

andrewyuq: I see it refers to the init() function down below - so do we reset these values in every 5-min window? Is this aligned with VSC's implemetation?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlin.math.roundToInt
+
+abstract class CodeWhispererCodeCoverageTracker(
+    private val timeWindowInSec: Long,
+    private val language: CodewhispererLanguage,
+    acceptedTokensSize: AtomicInteger,
+    totalTokensSize: AtomicInteger
+) : Disposable {
+    val percentage: Int
+        get() = if (totalTokensSize.get() != 0) (acceptedTokensSize.get().toDouble() / totalTokensSize.get() * 100).roundToInt() else 0
```
andrewyuq, line 32: Percentage calculation is wrong if we don't use rangeMarker to calculate the correct length of CodeWhisperer suggestion after 5 mins since, CodeWhisperer's suggestion when accepted can very posibilitty be different than the one after 5 mins.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,143 @@
...
+                    addAndGetAcceptedTokens(remainingRecomm.length)
+                }
+            }
+        )
+        scheduleCodeWhispererCodeCoverageTracker()
+    }
+
+    fun documentChanged(event: DocumentEvent) {
+        addAndGetTotalTokens(event.newLength - event.oldLength)
+    }
```
andrewyuq, line 57: If we just need the document length every 5 mins, can we just get the length when we need it instead of keeping this val? Do we need intermediate total length value for other purposes?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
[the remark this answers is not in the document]
Will-ShaoHua: > question on this condition: but what if user just replaced the whole text for example using their clipboard context? Do we still ignore them?

good point, let me experiment againt this scenario a bit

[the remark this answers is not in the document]
Will-ShaoHua: > If this case is property handled(i.e. do nothing) in `CodeWhispererCodeCoverageTracker` then we don't need this condition.

sounds good, we can delegate the logic to codeCoverageTracker instead of doing these here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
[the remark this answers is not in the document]
Will-ShaoHua: yes, actually for existing VSC impl, deletion of tokens is not tracked, whereas we do take deletion into account in JB (we increment/decrement totalTokens by checking the cursor offset changes), so we do have to add rangeMarker to track acceptedToken as well, ow it is likely to get a number > 100

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: yes I should have name it as `reset`, which should make more sense on its purpose. Yes it aligns with the design doc and VSC impl.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: yea make sense, added the patch PR

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: I dont get this one @andrewyuq , could you elaborate on it a bit thanks!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -594,5 +594,5 @@ interface CodeWhispererUserActionListener {
     fun navigatePrevious(states: InvocationContext) {}
     fun navigateNext(states: InvocationContext) {}
     fun beforeAccept(states: InvocationContext, sessionContext: SessionContext) {}
-    fun afterAccept(states: InvocationContext, sessionContext: SessionContext) {}
+    fun afterAccept(states: InvocationContext, sessionContext: SessionContext, remainingRecommendation: String) {}
```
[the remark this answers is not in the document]
Will-ShaoHua: yea make sense, I just thought by passing them directly from EditorManager will be cleaner. OW I will have the same logic as EditorManager has to determine the remaining recomm.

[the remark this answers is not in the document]
Will-ShaoHua: or just expose an API from EditorManager maybe something like getRemainingRecommendation(state, sessionContext)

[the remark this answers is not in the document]
Will-ShaoHua: sounds good

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
andrewyuq: Nvm I understanded this wrong, it's aggregated per language.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -18,6 +21,18 @@ class CodeWhispererEditorListener : EditorFactoryListener {
...
+                        val file = FileDocumentManager.getInstance().getFile(event.document)
+                        val lang = when (file?.extension) {
+                            "py" -> CodewhispererLanguage.Python
+                            "java" -> CodewhispererLanguage.Java
+                            "js" -> CodewhispererLanguage.Javascript
+                            else -> CodewhispererLanguage.Unknown
+                        }
+                        CodeWhispererCodeCoverageTracker.getInstance(lang).documentChanged(event)
```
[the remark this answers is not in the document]
Will-ShaoHua: @andrewyuq  could u point me how to get a PSI file here cuz based on the doc https://plugins.jetbrains.com/docs/intellij/psi-files.html kinda `project` is required but we don't have one here.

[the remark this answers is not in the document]
andrewyuq: `editor.project`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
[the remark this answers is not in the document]
Will-ShaoHua: @andrewyuq, I found that actually we only need the rangeMarker for the tracker to calculate the accepted token size, so now I only passed rangeMarker here. Remaining recommendation is not needed here anymore. We can either pass the rangeMarker in the messagebus or put it into our invocationContext or so, how do you think? thanks!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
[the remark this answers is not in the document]
Will-ShaoHua: updated

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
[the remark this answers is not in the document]
Will-ShaoHua: updated

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: updated

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -18,6 +23,12 @@ class CodeWhispererEditorListener : EditorFactoryListener {
             object : DocumentListener {
                 override fun documentChanged(event: DocumentEvent) {
                     CodeWhispererInvocationStatus.getInstance().documentChanged()
+                    editor.project?.let { project ->
+                        PsiDocumentManager.getInstance(project).getPsiFile(editor.document)?.toProgrammingLanguage() ?. let { languageName ->
+                            val language = ProgrammingLanguage(languageName)
```
andrewyuq, line 28: nit: just `language = ProgrammingLanguage(languageName).toCodeWhispererLanguage()` to be cleaner?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -0,0 +1,158 @@
...
+        totalTokens.set(0)
+        acceptedTokens.set(0)
+        rangeMarkers.clear()
+    }
+
+    private fun emitCodeWhispererCodeContribution() {
+        rangeMarkers.forEach {
+            if (!it.isValid) return@forEach
+            addAndGetAcceptedTokens(it.endOffset - it.startOffset)
+        }
```
andrewyuq, line 104: This is good, would be better if you can remove invalid ones in the list every 5 mins

Will-ShaoHua: but we already clear all rangemarkers every 5 mins @andrewyuq , refer to line 97, is that what you mean

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
Will-ShaoHua: ack

andrewyuq (approved):

PR #3221, opened 2022-07-25
Refactor CodeWhispererExplorerActionManager and add test test coverage

Will-ShaoHua:
## Types of changes

- [x] Refactor

## Description

Refactor CodeWhispererExploreActionManager class and add test coverage.

## Testing

New test cases added.

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
@@ -39,6 +39,7 @@ import software.aws.toolkits.jetbrains.services.codewhisperer.credentials.CodeWh
 import software.aws.toolkits.jetbrains.services.codewhisperer.editor.CodeWhispererEditorManager
 import software.aws.toolkits.jetbrains.services.codewhisperer.editor.CodeWhispererEditorUtil.getCaretPosition
 import software.aws.toolkits.jetbrains.services.codewhisperer.editor.CodeWhispererEditorUtil.getFileContextInfo
+import software.aws.toolkits.jetbrains.services.codewhisperer.explorer.CodeWhispererExploreStateType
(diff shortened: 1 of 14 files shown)


andrewyuq (commented):
What's the context/purpose of this refactor?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -559,11 +560,16 @@ class CodeWhispererService {
     }
 
     fun canDoInvocation(editor: Editor, type: CodewhispererTriggerType): Boolean {
-        if (type == CodewhispererTriggerType.AutoTrigger && !CodeWhispererExplorerActionManager.getInstance().isAutoEnabled()) {
+        if (type == CodewhispererTriggerType.AutoTrigger && !CodeWhispererExplorerActionManager.getInstance().getCodeWhispererExplorerState(
+                CodeWhispererExploreStateType.IsAutoEnabled
```
andrewyuq, line 564: taking this as an example, I am not sure what benefit we are getting from using the `getCodeWhispererExplorerState` as the "entry point" and pass different parameter, to me this requires more knowledge for the caller, . Here we changed logic to " get the `isAutoEnabled` state". And now it's fine since `isAutoEnabled` state is equivalent to "whether cwspr is enabled-automatically" but when we want to change it's implementation in the future we have to change this everywhere instead of just in `isAutoEnabled()`. To me this is a lack of extendibility. So again I am wonder what's the purpose of this factor and what does it achieve.

Will-ShaoHua: umm no specific purpose/context of the refactor, just thought if it will be cleaner but if it's adding the churm we can close this.

andrewyuq: I would hold this refactor for now for the extendibility reason above.

PR #3226, opened 2022-07-27
[CodeWhisperer] Update %Code Tracker  documentChanged() logic for edge case

Will-ShaoHua:
The current `if` condition aims to filter out documentChanged events which didn't result from users' typing / file changes. These special events are generated from IDE's reloading behavior. 
For example, there is a file with code `def add(`, then user paste `x, y)`, shut down the IDE very quickly and restart the IDE. The reloading happens when user restart the IDE because the IDE is trying to load the file from the disk and will emit documentEvent if there is difference between VF and file stored in the disk. The current filter logic will miss some reloading cases and fail to block them from entering metric calculation. 

This one line change PR is trying to fix the miss and handle all the reloading cases.

## Types of changes

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -60,13 +59,10 @@ abstract class CodeWhispererCodeCoverageTracker(
...
     fun documentChanged(event: DocumentEvent) {
-        // When open a file for the first time, IDE will also emit DocumentEvent for loading with `isWholeTextReplaced = true`
-        // Added this condition to filter out those events
+        // Added this condition to filter out IDE reloading files
         if (event.isWholeTextReplaced) {
             LOG.debug { "event with isWholeTextReplaced flag: $event" }
-            (event as? DocumentEventImpl)?.let {
-                if (it.initialStartOffset == 0 && it.initialOldLength == event.document.textLength) return
-            }
+            if (event.oldTimeStamp == 0L) return
```
andrewyuq, line 65: Is it guaranteed that, for all the reloading documentChanged events, oldTimeStamp is 0, and for all the normal user paste cases, this value is non-zero?

Will-ShaoHua: >> for all the normal user paste cases 
yes, they will have timestamp different from 0.

>> for all reloading documentChanged events
yes, from what I've observed
- case1: IDE reloading users' change on the document -> timestamp will be 0
- case2: some times IDE will reload the welcome.md (i can't consistent trigger this event), normally it will reload only once(timestamp = 0), but I did observe twice case where second reload didn't have timestamp = 0. However, since the update was done in the first reloading, so even the second reloading event enter `addAndGetTotalTokens()`, the `event.newLength - event.oldLength()` is 0, no tokens will be added.

Will-ShaoHua: >> for all reloading documentChanged events
yes, from what I've observed
- case1: IDE reloading users' change on the document -> timestamp will be 0
- case2: some times IDE will reload the welcome.md (i can't consistent trigger this event), normally it will reload only once(timestamp = 0), but I did observe twice case where second reload didn't have timestamp = 0. However, since the update was done in the first reloading, so even the second reloading event enter `addAndGetTotalTokens()`, the `event.newLength - event.oldLength()` is 0, no tokens will be added.

Will-ShaoHua: using timestamp will be a more solid approach than what I used previously

Will-ShaoHua: Loading cwspr welcome.md - 1
image (1)

Loading cwspr welcome.md - 
image (2)
2

Will-ShaoHua: def add() -> def add(x, y) -> shutdown IDE -> restart IDE
with the current logic, these loading case will pass the condition and get into `addAndGetTotalTokens()`

image

def add() -> (empty) -> shutdown IDE -> restart IDE
image (1)

andrewyuq (approved):

rli (approved):

PR #3230, opened 2022-07-29
[CodeWhisperer] Update %code tracker rangerMarker logic

Will-ShaoHua:
Update acceptedTokens calculation logic. 
Now we use edit distance instead of offset to determine users' modification on CWSPR generated recommendation. With this approach, it can capture more complicated edits/modification done by users.

For example:
- generated recommendation:
```
x, y):
     return x + y
```
- modified recommendation (1):

```
x, y):
"""some comments added by users, we shouldn't count it as one generated by CodeWhisperer"""
    return x + y
```

- modified recommendation (2)
```
xxx, yyy):
    return xxx + yyy
```

-modified recommendation (3)
```
a, b):
    return a + b
```

This PR aims to fix this part.

## Types of changes

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -106,9 +108,16 @@ abstract class CodeWhispererCodeCoverageTracker(
...
-        rangeMarkers.forEach {
-            if (!it.isValid) return@forEach
-            addAndGetAcceptedTokens(it.endOffset - it.startOffset)
+        rangeMarkers.forEach { rangeMarker ->
+            if (!rangeMarker.isValid) return@forEach
+            // if users add more code upon the recommendation generated from CodeWhisperer, we consider those added part as userToken but not CwsprTokens
+            val originalRecommendationLength = rangeMarker.getUserData(key)
+            originalRecommendationLength?.let {
+                addAndGetAcceptedTokens((rangeMarker.endOffset - rangeMarker.startOffset).coerceAtMost(originalRecommendationLength))
+            } ?: run {
```
andrewyuq, line 117: If user deletes some characters in the recommendation and add new characters back into the recommendations to make it as the same length as original, should we take account of the deletion but not the addition to the recommendation length, but here the logic would still treat it as recommendation staying the same.

Will-ShaoHua: yea make sense, the other workaround I think of right now is to use what `usermodification` is using: Levenshtein edit distance, for example, we have the %modification on the recommendation generated by CWSPR, simply take (1 - %modification) and multiply by the original recomm length, then here we get unmodified accepted recommendation. Does this make sense to you @andrewyuq

Will-ShaoHua: but I am still thinking if we have other approach to do this

Will-ShaoHua: have some thoughts, let me check if it's doable and update here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -106,9 +138,21 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+            // if users add more code upon the recommendation generated from CodeWhisperer, we consider those added part as userToken but not CwsprTokens
+            val originalRecommendation = rangeMarker.getUserData(remainingRecommendationKey)
+            val modifiedRecommendation = extractRangeMarkerString(rangeMarker)
+            if (originalRecommendation == null || modifiedRecommendation == null) {
+                LOG.debug {
+                    "failed to get accepted recommendation. " +
+                        "OriginalRecommendation is null: ${originalRecommendation == null}; " +
+                        "ModifiedRecommendation is null: ${modifiedRecommendation == null}"
+                }
+                return
```
andrewyuq, line 152: rangeMarker becomes invalid (all texts are removed in this rangeMaker),modifiedRecommendation will be null here, resulting in an early return. But this is not quite what we want right? If this rangeMarker is invalid, we should continue to other rangeMarkers.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+            originalRecommendationLength < modifiedRecommendationLength ->
+                (modifiedRecommendationLength - editDistance).coerceAtMost(originalRecommendationLength)
+
+            originalRecommendationLength > modifiedRecommendationLength -> originalRecommendationLength - editDistance
+
+            else -> originalRecommendationLength - editDistance
+        }
+    }
+
+    protected open fun getEditDistance(modifiedString: String, originalString: String): Double = levenshteinChecker.distance(modifiedString, originalString)
```
andrewyuq, line 104: break this into two lines if this line is too long

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+        val editDistance = getEditDistance(modifiedRecommendation, originalRecommendation).toInt()
+        val originalRecommendationLength = originalRecommendation.length
+        val modifiedRecommendationLength = modifiedRecommendation.length
+        return when {
+            originalRecommendationLength < modifiedRecommendationLength ->
+                (modifiedRecommendationLength - editDistance).coerceAtMost(originalRecommendationLength)
+
+            originalRecommendationLength > modifiedRecommendationLength -> originalRecommendationLength - editDistance
+
+            else -> originalRecommendationLength - editDistance
```
andrewyuq, line 100: these two branches can be combined?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
         addAndGetTotalTokens(event.newLength - event.oldLength)
     }
 
+    internal fun extractRangeMarkerString(rangeMarker: RangeMarker?): String? = runReadAction {
+        rangeMarker?.range?.let { myRange -> rangeMarker.document.getText(myRange) }
+    }
+
+    internal fun getAcceptedTokensDelta(originalRecommendation: String, modifiedRecommendation: String): Int {
+        val editDistance = getEditDistance(modifiedRecommendation, originalRecommendation).toInt()
```
andrewyuq, line 91: can you add a comment here why using editDistance makes the calculation correct? Comments would help the algorithm used here.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+        val originalRecommendationLength = originalRecommendation.length
+        val modifiedRecommendationLength = modifiedRecommendation.length
```
andrewyuq, line 93: nit: these two variables are not necessary

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+        return when {
+            originalRecommendationLength < modifiedRecommendationLength ->
+                (modifiedRecommendationLength - editDistance).coerceAtMost(originalRecommendationLength)
```
andrewyuq, line 96: This is fine but if everything is done right, you shouldn't need this coerceAtMost(), it should be implictly.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+
+            originalRecommendationLength > modifiedRecommendationLength -> originalRecommendationLength - editDistance
+
+            else -> originalRecommendationLength - editDistance
+        }
```
andrewyuq, line 101: For the whole logic, if you want: `return max(original, modified) - editDistance`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: yep, when a rangeMarker is invalid, it will return@forEach, so will continue with other rangeMarker instead of early return

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: yea I will add some comment or example here for demonstration, but the case is it is still not 100% capturing the users' modification as same as UserModification, but it's the best effort I can think of for now and IMO it can capture most of the cases and provide a good approximate on users' modification.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
andrewyuq: so this should be return@forEach instead of return right?

Will-ShaoHua: oh yes, it's my miss, will update

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +83,26 @@ abstract class CodeWhispererCodeCoverageTracker(
         addAndGetTotalTokens(event.newLength - event.oldLength)
     }
 
+    internal fun extractRangeMarkerString(rangeMarker: RangeMarker?): String? = runReadAction {
```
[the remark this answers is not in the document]
Will-ShaoHua: updated

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: yea good point, that's right.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererCodeCoverageTrackerTest.kt
```diff
@@ -318,22 +321,57 @@ class CodeWhispererCodeCoverageTrackerTest {
         verify(batcher, Times(0)).enqueue(any())
     }
 
+    @Test
```
Will-ShaoHua, line 324: @andrewyuq , except the comments I left at the edit distance logic, here are some test cases you can refer to

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -71,6 +84,23 @@ abstract class CodeWhispererCodeCoverageTracker(
...
+        rangeMarker.range?.let { myRange -> rangeMarker.document.getText(myRange) }
+    }
+
+    // With edit distance, complicate usermodification can be considered as simple edit(add, delete, replace),
+    // and thus the unmodified part of recommendation length can be deducted/approximated
+    internal fun getAcceptedTokensDelta(originalRecommendation: String, modifiedRecommendation: String): Int {
+        val editDistance = getEditDistance(modifiedRecommendation, originalRecommendation).toInt()
+        // ex. (modified > original): originalRecom: foo -> modifiedRecom: fobarbarbaro, distance = 9, delta = 12 - 9 = 3
+        // ex. (modified == original): originalRecom: helloworld -> modifiedRecom: HelloWorld, distance = 2, delta = 10 - 2 = 8
+        // ex. (modified < original): originalRecom: CodeWhisperer -> modifiedRecom: CODE, distance = 12, delta = 13 - 12 = 1
```
andrewyuq, line 97: This comment can be placed above the fuction header

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -137,6 +179,10 @@ abstract class CodeWhispererCodeCoverageTracker(
     }
 
     companion object {
+        @JvmStatic
+        protected val levenshteinChecker = Levenshtein()
+        private const val REMAINING_RECOMMENDATION = "remainingRecommendation"
+        private val remainingRecommendationKey = Key<String>(REMAINING_RECOMMENDATION)
```
andrewyuq, line 185: rename this as KEY_REMAINING_RECOMMENDATION may be better IMO

andrewyuq (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -26,7 +27,10 @@ class CodeWhispererEditorListener : EditorFactoryListener {
                     editor.project?.let { project ->
                         PsiDocumentManager.getInstance(project).getPsiFile(editor.document)?.toProgrammingLanguage() ?. let { languageName ->
                             val language = ProgrammingLanguage(languageName).toCodeWhispererLanguage()
-                            CodeWhispererCodeCoverageTracker.getInstance(language).documentChanged(event)
+                            CodeWhispererCodeCoverageTracker.getInstance(language).apply {
+                                if (!isActive && isTelemetryEnabled()) activateTracker()
```
rli, line 31: would prefer this to be called `activateTrackerIfNotActive`

and move this logic there

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -32,6 +37,8 @@ abstract class CodeWhispererCodeCoverageTracker(
     private val totalTokens: AtomicInteger,
     private val rangeMarkers: MutableList<RangeMarker>
 ) : Disposable {
+    var isActive: Boolean = false
```
rli, line 40: atomic?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
Will-ShaoHua: updated!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
Will-ShaoHua: updated!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -19,6 +23,7 @@ import software.aws.toolkits.jetbrains.services.codewhisperer.popup.CodeWhispere
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererConstants.TOTAL_SECONDS_IN_MINUTE
 import software.aws.toolkits.telemetry.CodewhispererLanguage
 import software.aws.toolkits.telemetry.CodewhispererTelemetry
+import java.lang.Integer.max
```
rli, line 26: nit: use maxOf from kotlin stdlib

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -40,33 +45,59 @@ abstract class CodeWhispererCodeCoverageTracker(
...
                     rangeMarkers.add(rangeMarker)
+                    val originalRecommendation = extractRangeMarkerString(rangeMarker)
+                    originalRecommendation?.let {
+                        rangeMarker.putUserData(KEY_REMAINING_RECOMMENDATION, it)
+                    }
                 }
             }
         )
+        startTime = Instant.now()
+        isActive.set(true)
```
rli, line 70: should be set before listener is subscribed

or just mark the entire method as synchronized so you don’t have to worry about any of that

Will-ShaoHua: hi @rli, just for my own education, could you explain a bit more what's the concern here if we set `isActive = true` after subscription of listener? thanks!

abhikuhikar (approved):

rli (approved):

PR #3232, opened 2022-07-29
Add 212 deprecation message

rli:
## License


manodnyab:
Probably also need to change the MIN_VERSION value in MinimumVersionChange.kt

manodnyab (approved):

PR #3245, opened 2022-08-04
[CodeWhisperer] Send empty telemetry events for empty recommendations

andrewyuq:
1. There are two types of empty recommendations server could
return to the client. The first one is an empty list of recommendations.
The second one is a list of recommendations but some recommendations
are empty (strings). Previously we don't send user decision events
for both cases. This is resulting in a discrepancy between serviceInvocation
event and userDecision event per codewhispererSessionId. We decided to send
userDecision events for both. Additionally, for empty list, we just send
the event and stop early, since there isn't additonal things to render.
For empty recommendations, they will be kept in the state but not shown
to the users, and userDecision events for them will be sent later when
we send events for all recommendations we received on session close.

## Types of changes

## Checklist

## License


Will-ShaoHua (approved):
LGTM

rli (approved):

PR #3265, opened 2022-08-16
[CodeWhisperer] Add missing userDecision events for conflicting popup case

andrewyuq:
1. When CodeWhisperer is invoked but IntelliSense popup is opened
after, we are not showing recommendations due to conflicting popups
in this case previously we rely on disposing the popup from the caller
processCodeWhispererUI(), but the popupListener to send userDecision
event has not yet been registered to the popup. So we send the event
manually just like the caret moving backward case.

## Types of changes

## Testing

Added test, the test was not passing before the fix and it passed after the fix.
## Checklist

## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererTelemetryTest.kt
```diff
@@ -380,6 +380,36 @@ class CodeWhispererTelemetryTest : CodeWhispererTestBase() {
...
+                true
+            }
+        }
+        invokeCodeWhispererService()
+
+        runInEdtAndWait {
+            val metricCaptor = argumentCaptor<MetricEvent>()
+            verify(batcher, atLeastOnce()).enqueue(metricCaptor.capture())
+            assertEventsContainsFieldsAndCount(
+                metricCaptor.allValues, userDecision, pythonResponse.recommendations().size,
```
Will-ShaoHua, line 407: just a nit-pick Andrew, should we make it multi-lines

Will-ShaoHua (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererTelemetryTest.kt
rli: yes, each value should be on its on line if you're breaking it up like this. not sure why our lint rule isn't catching this

rli (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererTelemetryTest.kt
andrewyuq: good catch

PR #3276, opened 2022-08-19
[CodeWhisperer] %Tracker will not be activated by users without enabling CodeWhisperer

Will-ShaoHua:
After deep-dive with the abundance of 0 metrics sent, we found %codeTracker didn't have this condition to filter out users who don't activate CodeWhisperer. Therefore it result in those meaningless 0s because everyone with plugin installed and telemetry enabled will send CodeWhispererPercentage. 

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt
```diff
@@ -52,7 +53,7 @@ abstract class CodeWhispererCodeCoverageTracker(
 
     @Synchronized
     fun activateTrackerIfNotActive() {
-        if (!isTelemetryEnabled() || isActive.get()) return
+        if (!isTelemetryEnabled() || !CodeWhispererExplorerActionManager.getInstance().hasAcceptedTermsOfService() || isActive.get()) return
```
andrewyuq, line 56: In fact this is a problem not only for the calculation of this codePercentage metric but for all CodeWhisperer logics, we want the various listeners to not take effect when CWSPR is disabled. We can put this check in the documentChanged() event in `CodeWhispererEditorListener`. There needs to be other changes but for the scope of this issue this will be enough.

andrewyuq: And replace `isActive.get()` with `isTrackerActive()` here?

Will-ShaoHua: > 

yea make sense

Will-ShaoHua: yea make sense we also need to add this condition in all other cwspr logics. Yes, let me put this in editorListener

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererEditorListener.kt
```diff
@@ -10,13 +10,14 @@ import com.intellij.openapi.editor.event.EditorFactoryListener
...
 import software.aws.toolkits.jetbrains.services.codewhisperer.editor.CodeWhispererEditorUtil.codeWhispererLanguage
+import software.aws.toolkits.jetbrains.services.codewhisperer.explorer.CodeWhispererExplorerActionManager
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererInvocationStatus
 import software.aws.toolkits.jetbrains.services.codewhisperer.telemetry.CodeWhispererCodeCoverageTracker
 
 class CodeWhispererEditorListener : EditorFactoryListener {
     override fun editorCreated(event: EditorFactoryEvent) {
         val editor = (event.editor as? EditorImpl) ?: return
-
+        if (!CodeWhispererExplorerActionManager.getInstance().hasAcceptedTermsOfService()) return
```
andrewyuq, line 20: No I don't mean this one, I mean to put this check only inside the start of the below documentChanged() event. Since doing this will miss for the cases when the users have the editor opened when CodeWhisperer was not enabled yet, but then enabled after. This event is only called when the editor is initially created, thus for all these editors they will miss CodeWhisperer logic. By putting it below we are saying "we are adding this documentListener to the editor anyway but do nothing if CWSPR is not enabled"

Will-ShaoHua: ack make sense

andrewyuq (approved):

rli (approved):

PR #3298, opened 2022-09-12
[codewhisperer]: emit telemetry for perceived latency

zixlin7:
## Types of changes

## Description
To emit telemetry on the new perceivedLatency event: the definition of this event is added in this https://github.com/aws/aws-toolkit-common/pull/366
corresponding VSC PR [here](https://github.com/aws/aws-toolkit-vscode/pull/2882)

Screen Shot 2022-09-12 at 2 48 41 PM

## Checklist

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
@@ -322,8 +322,6 @@ class CodeWhispererService {
             }
         } else if (response.recommendations().isNotEmpty()) {
             updateCodeWhisperer(nextStates, isPopupShowing)
-            val perceivedLatency = CodeWhispererInvocationStatus.getInstance().getTimeSinceDocumentChanged()
-            CodeWhispererTelemetryService.getInstance().sendPerceivedLatencyEvent(requestId, requestContext, responseContext, perceivedLatency)
         }
         return nextStates
     }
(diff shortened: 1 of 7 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -322,6 +322,8 @@ class CodeWhispererService {
             }
         } else if (response.recommendations().isNotEmpty()) {
             updateCodeWhisperer(nextStates, isPopupShowing)
+            val perceivedLatency = CodeWhispererInvocationStatus.getInstance().getTimeSinceDocumentChanged()
+            CodeWhispererTelemetryService.getInstance().sendPerceivedLatencyEvent(requestId, requestContext, responseContext, perceivedLatency)
```
andrewyuq, line 326: Since this is a per-trigger event -- let's put this in `CodeWhispererPopupManager.showPopup()` https://github.com/zixlin7/aws-toolkit-jetbrains/blob/111d4986ee590516c4e7fb9e21d2310a71c34c2f/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt#L296
This is the path that's only reached when it's called first time and we are about to show the popup. Doing this will imply that this event will only be sent in those "shown" cases(Accepted or Rejected), which makes sense since for recommendations that're not shown, the concept of "perceived latency" does not apply.

And we want to make sure for VSC this is also true.

zixlin7: @andrewyuq updated and it's already the case for VSC. Thanks!

andrewyuq (approved):

rli (approved):

PR #3306, opened 2022-09-20
feat(codewhisperer): auto-trigger improvements

zixlin7:
## Types of changes

## Description
Improvements to auto-trigger: replacing the key stroke count based auto-trigger to based on idle time

## Motivation and Context
To make auto-trigger more aggressive so users see more recommendations

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/editor/CodeWhispererTypedHandler.kt
```diff
@@ -4,29 +4,44 @@
...
+                CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()
+            ) {
+                if (!isActive) return@launch
+                delay(CodeWhispererConstants.IDLE_TIME_CHECK_INTERVAL)
+            }
+            runInEdt {
+                if (CodeWhispererInvocationStatus.getInstance().isPopupActive()) return@runInEdt
+                performAutomatedTriggerAction(editor, CodewhispererAutomatedTriggerType.IdleTime)
+                triggerOnIdle?.cancel()
+            }
```
andrewyuq, line 44: There might be a rare race condition issue about when we check `CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()` and when we enter Edt (to invoke CW). However the worst case it can happen is in `performAutomatedTriggerAction` the `hasExistingInvocation` check returns true and returns, so no invocation will happen. It's better we can add a TODO here mentioning this (not a blocker for the PR)

andrewyuq: The `triggerOnIdle?.cancel()` inside Edt may not be necessary, when it's in Edt the coroutine that schedules it should have finished.

zixlin7: updated, thanks!

PR #3315, opened 2022-09-27
Add 223 profile

rli:
## License


resources/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ import software.aws.toolkits.gradle.resources.ValidateMessages
 plugins {
     id("toolkit-kotlin-conventions")
     id("toolkit-testing")
-    id("de.undercouch.download")
+    id("de.undercouch.download") version "5.2.1"
```
manodnyab, line 10: We dont usually specify plugin version numbers, would this break current users?

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
```diff
@@ -248,19 +250,37 @@ class ResizingTextColumnRenderer : ResizingColumnRenderer() {
...
+        null
+    } else {
+        val errors = component.validateCallbacks.mapNotNull { it() }
+        if (errors.isEmpty()) {
+            component.apply()
+        }
+        errors.firstOrNull()
+    }
+
+// backport since removed in 223
```
manodnyab, line 276: Could we change the comment to remove when min is 223

manodnyab (approved):

resources/build.gradle.kts
rli: no this is a build dep

jetbrains-core/src/software/aws/toolkits/jetbrains/utils/ui/UiUtils.kt
rli: this is more like remove when no longer used since this util doesn't exist anymore

PR #3300, opened 2022-09-14
[Codewhisperer] Refactor cwspr programmingLanguage system

Will-ShaoHua:
Refactor programming language system used in cwspr, reducing converting/mapping back and forward to different language types (CodeWhispererLanguage, ProgrammingLanguage etc..)

## Reasons:
- scalability
- maintanability

## Changes:
- use extension points and make it declarative to add programming language (refer to Toolkit serviceNode implementation)

cpp
java
js
jsx
kotlin
plain_text
python
typscript

## Checklist

- [ ] I have read the **README** document
- [ ] I have read the **CONTRIBUTING** document
- [ ] Local run of `gradlew check` succeeds
- [ ] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if change is customer facing in the IDE.
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3318, opened 2022-09-28
[CodeWhisperer] fix and remove incorrect support of .tsx file

Will-ShaoHua:
## Problem
`.tsx` file type is wrongly supported

## Solution
TSX file type name: `TypeScript JSX`
JSX file type name: `JSX harmony`

Ad-hoc fix: Update `CodeWhispererLanguageManager.isLanguageSupported()`
(another refactor PR which will use declarative way for languages supported by CodeWhisperer #3300 will reduce the string coversion back and forward)

#### .tsx

image

#### .jsx

image

 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt
@@ -16,6 +16,7 @@ class CodeWhispererLanguageManager {
     )
 
     fun isLanguageSupported(language: ProgrammingLanguage): Boolean = supportedLanguage.contains(language.languageName)
+
     /**
      * This should be called to map some language dialect to their mother language
      * e.g. JSX -> JavaScript, TypeScript -> JavaScript etc.
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt
@@ -11,14 +11,11 @@ class CodeWhispererLanguageManager {
     private val supportedLanguage = setOf(
         CodewhispererLanguage.Java.toString(),
         CodewhispererLanguage.Python.toString(),
-        CodewhispererLanguage.Javascript.toString()
+        CodewhispererLanguage.Javascript.toString(),
+        "jsx harmony"
     )
 
-    fun isLanguageSupported(language: ProgrammingLanguage): Boolean {
-        val mappedLanguage = getParentLanguage(language)
-        return supportedLanguage.contains(mappedLanguage.languageName)
-    }
-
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt
```diff
@@ -11,13 +11,11 @@ class CodeWhispererLanguageManager {
...
-        CodewhispererLanguage.Javascript.toString()
+        CodewhispererLanguage.Javascript.toString(),
+        "jsx harmony"
     )
 
-    fun isLanguageSupported(language: ProgrammingLanguage): Boolean {
-        val mappedLanguage = getParentLanguage(language)
-        return supportedLanguage.contains(mappedLanguage.languageName)
-    }
+    fun isLanguageSupported(language: ProgrammingLanguage): Boolean = supportedLanguage.contains(language.languageName)
```
zixlin7, line 18: instead of adding "jsx harmony" to the supported language set, can we do:
```
val mappedLanguage = getParentLanguage(language).toCodeWhispererLanguage()
return supportedLanguage.contains(mappedLanguage.toString())
```

zixlin7: and in `getParentLanguage` map "jsx harmony" to javascript?

Will-ShaoHua: yepp, updated with some tweak, we can't call `getParentLanguage()` here since it will map `jsx` -> `javascript`

Will-ShaoHua: instead 
```
fun isLanguageSupported(language: ProgrammingLanguage): Boolean {
        val cwsprLanguage = language.toCodeWhispererLanguage()
        return cwsprLanguage != CodewhispererLanguage.Unknown && cwsprLanguage != CodewhispererLanguage.Plaintext
    }
```

Will-ShaoHua: we check the transformed codewhispererLanguage type directly which have: java, python, javascript, jsx, plaintext, unknown

zixlin7: sorry is there any downside for mapping `jsx` to `javascript` for checking support purpose?

zixlin7: actually if we just check not `unkown` and not `plaintext`, we don't need the supported language set at all. But I think it's safer to still check against the parent

zixlin7 (approved):

Will-ShaoHua:
@rli we might want to get the refactor PR merged https://github.com/aws/aws-toolkit-jetbrains/pull/3300 to decrease the conversion back and forward from different ProgrammingLanguage data class

PR #3328, opened 2022-10-07
Add 213 deprecation message

rli:
## License


manodnyab (commented):
May need to update the MIN_VERSION value in MinimumVersionChange.kt

manodnyab (approved):

PR #3407, opened 2022-12-10
(feat codewhisperer) codewhisperer status widget will be hide when there is no connection

Will-ShaoHua:
https://user-images.githubusercontent.com/96078566/207114098-0ccf17c3-3185-4171-9905-8c2a4df21c2e.mov

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/status/CodeWhispererStatusBarManager.kt
```diff
@@ -0,0 +1,62 @@
...
+import software.aws.toolkits.jetbrains.core.credentials.pinning.CodeWhispererConnection
+import software.aws.toolkits.jetbrains.core.credentials.pinning.ConnectionPinningManagerListener
+import software.aws.toolkits.jetbrains.core.credentials.pinning.FeatureWithPinnedConnection
+import software.aws.toolkits.jetbrains.services.codewhisperer.explorer.isCodeWhispererEnabled
+
+/**
+ * Manager visibility of CodeWhisperer status bar widget, only display it when CodeWhisperer is connected
+ */
+@Service(Service.Level.PROJECT)
+class CodeWhispererStatusBarManager(private val project: Project) {
```
JadenSimon, line 25: I'm not too familiar with this codebase creating a separate top-level class for this functionality seems heavy. 

No need to make any changes, this is just an observation. Personally I would've implemented this logic in `CodeWhispererStatusBarWidgetFactory::createWidget` though maybe it would be harder to test?

Will-ShaoHua: yeaa agree, kinda feel not worth creating another top-level class for this simple functionality, Jaden let me create another SIM tracking this optimization since we wanna roll out this fix tomorrow lol

zixlin7 (approved):

PR #3409, opened 2022-12-11
(feat codewhisperer) reauth UX flow update

Will-ShaoHua:
* make reauth prompt toast notification but not blocking dialog
* invalidate expired connection once user click on re-auth button
* re-factor logoutFromConnection

* case 1: token is expired, refreshable (modify expiredAt then reload the IDE to simulate this case)

https://user-images.githubusercontent.com/96078566/207104845-10b480d8-8d2d-47be-b3d5-fba6686cba99.mov

* case 2: token is expired, irrefreshable (modify expiredAt & refreshToken then reload the IDE to simulate this case)

https://user-images.githubusercontent.com/96078566/207459676-f5213517-502d-4564-a84c-18fcb011d728.mov

 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
```diff
@@ -94,18 +99,39 @@ object CodeWhispererUtil {
...
+        logoutFromSsoConnection(project, connection) {
+            UiTelemetry.click(project, "signout_codewhisperer_expired_connection")
+        }
+        notifyError(
+            message("toolkit.sso_expire.dialog.title", connection.label),
+            message("toolkit.sso_expire.dialog_message"),
+            project,
+            listOf(
+                NotificationAction.create(message("toolkit.sso_expire.dialog.yes_button")) { _, notification ->
+                    runInEdt { ToolkitAddConnectionDialog(project, connection).show() }
```
JadenSimon, line 125: Does this open a dialog for adding a new connection? Would users be confused?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -143,6 +144,18 @@ fun loginSso(project: Project?, startUrl: String, scopes: List<String> = ALL_SON
     }
 }
 
+fun logoutFromSsoConnection(project: Project?, connection: ToolkitConnection, callback: () -> Unit = {}) {
+    if (connection !is AwsBearerTokenConnection) return
```
JadenSimon, line 148: I think it makes more sense to type `connection` as `AwsBearerTokenConnection` in the signature instead of returning early. Especially because the function is called `logoutFromSsoConnection` instead of `logoutFromConnection`

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -143,6 +144,18 @@ fun loginSso(project: Project?, startUrl: String, scopes: List<String> = ALL_SON
...
+    ApplicationManager.getApplication().messageBus.syncPublisher(BearerTokenProviderListener.TOPIC).invalidate(connection.id)
+    ToolkitAuthManager.getInstance().deleteConnection(connection.id)
+
+    project?.let {
+        ToolkitConnectionManager.getInstance(it).switchConnection(null)
+    }
+
+    callback()
```
JadenSimon, line 156: Callback never gets executed if the previous statements throw. This is somewhat surprising IMO.

Will-ShaoHua: ack make sense

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
Will-ShaoHua: this should happen when 
- we fail to refresh the connection at first place
- when user click on the `authenticate` button in the expiry notification, we logout immediately from the stale connection and will fill the metadata(sso url) if it's a sso connection

Will-ShaoHua: does this make sense to you Jaden

Will-ShaoHua: can refer to second screen record 0:54

Will-ShaoHua: I try to avoid the state where users connected with a expired connection, so i log out of them once I detected the connection unrefreshable

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
Will-ShaoHua: but have 1 qq: since here we don't have any API call, can't we expect or assume it will always succeed?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
JadenSimon: I think it's ok to jump the user straight into the browser instead of the add connection dialog.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
JadenSimon: > can't we expect or assume it will always succeed?

No, we should never assume that a statement will succeed. Every line of imperative logic is directly dependent on the success of the preceding line. If you want to guarantee execution then a try/finally is needed.

Will-ShaoHua: gotcha thanks!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
Will-ShaoHua: make sense, let me update this abit

Will-ShaoHua: Jaden I updated the screen recording #2 as well where I modified both `expiredAt` `refrehToken` so that the credential is ir-refreshable

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -143,6 +144,17 @@ fun loginSso(project: Project?, startUrl: String, scopes: List<String> = ALL_SON
     }
 }
 
+fun logoutFromSsoConnection(project: Project?, connection: AwsBearerTokenConnection, callback: () -> Unit = {}) {
+    ApplicationManager.getApplication().messageBus.syncPublisher(BearerTokenProviderListener.TOPIC).invalidate(connection.id)
+
+    try {
+        ToolkitAuthManager.getInstance().deleteConnection(connection.id)
```
JadenSimon, line 151:
```suggestion
    try {
        ApplicationManager.getApplication().messageBus.syncPublisher(BearerTokenProviderListener.TOPIC).invalidate(connection.id)
        ToolkitAuthManager.getInstance().deleteConnection(connection.id)
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
```diff
@@ -94,18 +101,41 @@ object CodeWhispererUtil {
...
+                notifyConnectionExpired(project, connection)
                 callback()
             }
         }
     }
 
+    private fun notifyConnectionExpired(project: Project, connection: BearerSsoConnection?) {
+        connection ?: return
+        logoutFromSsoConnection(project, connection) {
+            UiTelemetry.click(project, "signout_codewhisperer_expired_connection")
```
JadenSimon, line 119: This feels out of place. Has the user clicked anything when this code is executed?

Will-ShaoHua: in fact, not at all, we logout on behalf of them basically, but we don't have other types of telemetry for now, or I can just remove this part?

Will-ShaoHua: as the name suggested, i think i should remove the telemetry report here then add others later

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -76,9 +76,7 @@ class CodeWhispererService {
         val project = editor.project ?: return
         if (!isCodeWhispererEnabled(project)) return
         if (isConnectionExpired(project)) {
-            if (triggerTypeInfo.triggerType == CodewhispererTriggerType.OnDemand) {
-                promptReAuth(project)
-            }
+            promptReAuth(project)
```
JadenSimon, line 79: What's the behavior if the user doesn't interact with the notification but keeps typing anyway?

Will-ShaoHua: yea, since once the notification is shown once, it will logout the stale connection right away, user will be unable to trigger CW service from them on

Will-ShaoHua: `promptReAuth` will try
- refresh the connection
  - if succeed -> nothing will happen
  - if fail -> show the notification telling users the connection is expired and logout from it

Will-ShaoHua: and users can click on the `autehnticate` button on the notification mentioned above which will start the browser flow immediately instead of starting from the login dialog

JadenSimon: I see, this is dropping the persistence entirely. The new behavior might not play nicely with the UX of other features e.g. CodeCatalyst. I don't see too much risk with this change though I think we should check if the CodeCatalyst implementation has any logic to prompt on expiration.

Will-ShaoHua: make sense, i'm also wondering where does CAWS require the connection/credential/access token (e.g. backend API call or something etc.)

Will-ShaoHua: btw, just played around with such case (modified AWS ID `expiredAt` `refreshToken`), it will logout from both CAWS + CWSPR

Will-ShaoHua: but worth verifying this with @rli

Will-ShaoHua: what i found is this https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/sono/SonoCredentialManager.kt#L46

Will-ShaoHua: ^^ https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/sono/SonoCredentialManager.kt#L96 seems do the same thing: start the browser flow directly

Will-ShaoHua: i can change the CWSPR part as same as CAWS looks like

JadenSimon: As long as the CWSPR change doesn't break the CodeCatalyst behavior then it's fine as-is.

JadenSimon (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
Will-ShaoHua: ack, will hold this PR and see if @rli is back tomorrow and if he have any comment on this thanks Jaden

zixlin7 (approved):

PR #3437, opened 2023-01-06
Add new latency related metrics

andrewyuq:
Will paste sample telemetry events soon.

sample telemetry:
DefaultDatum(name=codewhisperer_clientComponentLatency, value=1.0, unit=None, passive=true, metadata={codewhispererCredentialFetchingLatency=0.0, credentialStartUrl=https://view.awsapps.com/start, codewhispererLanguage=java, codewhispererPostprocessingLatency=94.0, codewhispererFirstCompletionLatency=2134.0, codewhispererEndToEndLatency=2277.0, codewhispererCompletionType=Block, codewhispererRequestId=b0936efa-c3f5-4982-8552-336164bfa6a0, codewhispererPreprocessingLatency=48.0, codewhispererAllCompletionsLatency=2833.0, codewhispererSessionId=efa21841-d26d-4004-b380-0a95fc558e3b, codewhispererTriggerType=OnDemand})

Corresponding VSC PR: https://github.com/aws/aws-toolkit-vscode/pull/3090#pullrequestreview-1239209903

Details: The new latency event contains all the fields that are of the scope of a user trigger, that is, there will be only one such event for each user trigger action. So bind that to the `requestContext` which also has a user-trigger scope.

## Types of changes

## Checklist

## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/actions/CodeWhispererRecommendationAction.kt
```diff
@@ -19,11 +20,14 @@ class CodeWhispererRecommendationAction : AnAction(message("codewhisperer.trigge
     }
 
     override fun actionPerformed(e: AnActionEvent) {
+        val latencyContext = LatencyContext()
+        latencyContext.codewhispererPreprocessingStart = System.nanoTime()
+        latencyContext.codewhispererEndToEndStart = System.nanoTime()
```
Will-ShaoHua, line 25: nit: maybe we can put it in the constructor?

Will-ShaoHua (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -306,6 +306,12 @@ class CodeWhispererPopupManager {
         } else {
             WindowManager.getInstance().setAlphaModeRatio(popup.popupWindow, 0.1f)
         }
+        states.requestContext.latencyContext.codewhispererPostprocessingEnd = System.nanoTime()
+        states.requestContext.latencyContext.codewhispererEndToEndEnd = System.nanoTime()
+
+        if (!CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()) {
+            CodeWhispererTelemetryService.getInstance().sendClientComponentLatencyEvent(states)
```
leigaol, line 313: Does this guarantee that the code suggestion is shown to the user before client component latency is sent?

leigaol (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
andrewyuq: yes, when is line is executed, the line of showing popup is guranteed to be executed above. But this reminds me an edge case where scrolling would also enter this path and send the event which we don't want. Will fix it.

andrewyuq: `if (!CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()) {` this is gating for the condition that all paginated response are received.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/actions/CodeWhispererRecommendationAction.kt
rli, line 25: since we end up converting to ms at the end, why not use `currentTimeMillis` directly?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererTelemetryService.kt
```diff
@@ -218,6 +220,42 @@ class CodeWhispererTelemetryService {
...
+            codewhispererRequestId = requestContext.latencyContext.firstRequestId,
+            codewhispererFirstCompletionLatency = requestContext.latencyContext.paginationFirstCompletionTime,
+            codewhispererPreprocessingLatency = TimeUnit.NANOSECONDS.toMillis(
+                requestContext.latencyContext.codewhispererPreprocessingEnd -
+                    requestContext.latencyContext.codewhispererPreprocessingStart
+            ).toDouble(),
+            codewhispererEndToEndLatency = TimeUnit.NANOSECONDS.toMillis(
+                requestContext.latencyContext.codewhispererEndToEndEnd -
+                    requestContext.latencyContext.codewhispererEndToEndStart
+            ).toDouble(),
```
JadenSimon, line 240: nit: adding some extension functions off `LatencyContext` for computing times might help readability slightly. Something like `getCodewhispererEndToEndTime` maybe?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererUIChangeListener.kt
```diff
@@ -92,5 +94,8 @@ class CodeWhispererUIChangeListener : CodeWhispererPopupStateChangeListener {
 
     override fun recommendationAdded(states: InvocationContext, sessionContext: SessionContext) {
         CodeWhispererPopupManager.getInstance().updatePopupPanel(states, sessionContext)
+        if (!CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()) {
+            CodeWhispererTelemetryService.getInstance().sendClientComponentLatencyEvent(states)
+        }
```
JadenSimon, line 99: So this will only get sent after adding a recommendation _and_ pagination has terminated?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -300,6 +300,12 @@ class CodeWhispererPopupManager {
                 states.responseContext,
                 perceivedLatency
             )
+            states.requestContext.latencyContext.codewhispererPostprocessingEnd = System.nanoTime()
+            states.requestContext.latencyContext.codewhispererEndToEndEnd = System.nanoTime()
+
+            if (!CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()) {
+                CodeWhispererTelemetryService.getInstance().sendClientComponentLatencyEvent(states)
+            }
```
JadenSimon, line 308: Why is this event duplicated in CodeWhispererUIChangeListener.kt? Are there two different entrypoints that never join? Emitting stateful events at multiple points in the code could cause issues later as the codebase evolves.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/actions/CodeWhispererRecommendationAction.kt
andrewyuq: I remembered we encountered some issues when using `currentTimeMillis` long time ago but forgot the exact details. Otherwise it's totally fine to use it.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererTelemetryService.kt
andrewyuq: make sense

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
andrewyuq: These two will never join although it's nontrivial to realize it for just reviewing the PR. Agreeing that a single emitting point makes the most sense to be less bug-prone in the future.

andrewyuq: The calling point here corresponds to this [comment](https://github.com/aws/aws-toolkit-jetbrains/pull/3437#discussion_r1065160422), for path 1)

andrewyuq: I can do a small refactor to make it to have a single emitting point.

JadenSimon: If it's a small change then yeah I'd go for it. Otherwise don't worry about it.

JadenSimon (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/actions/CodeWhispererRecommendationAction.kt
andrewyuq: yea we can, slightly preferred not to personally due to the fact that constructing this doesn't necessarily bind to the start time of these two. It just happens to be this case.

JadenSimon (approved):

PR #3441, opened 2023-01-11
Feedback option in gateway

manodnyab:
## License


jetbrains-core/resources/META-INF/plugin.xml
```diff
@@ -624,6 +624,7 @@ with what features/services are supported.
         <group id="aws.toolkit.sono.id.actions">
             <action class="software.aws.toolkits.jetbrains.core.credentials.sono.SonoLogoutAction" id="aws.toolkit.sono.id.logout"/>
             <action class="software.aws.toolkits.jetbrains.services.caws.actions.OpenCawsProfileAction" id="aws.toolkit.caws.profile"/>
+            <action class="software.aws.toolkits.jetbrains.ui.feedback.SubmitFeedbackInGateway" id="aws.toolkit.submit.caws.dev.env.feedback"/>
```
rli, line 627:
```suggestion
            <reference id="aws.toolkit.showFeedback" />
```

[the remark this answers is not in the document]
manodnyab: Not very wide, but we could probably skip the icon? @jparkhurst3 thoughts?

manodnyab: Wanted to avoid using the same class since the project used in both will be different. The project attached to the action event returns a null pointer exception

rli: can always pass in the default project if the project on the event is null

rli (approved):

PR #3460, opened 2023-01-26
Fix CW clientComponentLatency event is emitted multiple times for a session

andrewyuq:
clientComponenLatency event should not be emitted when user is just navigating recommendations or typing according to what the recommendation shows.

## Types of changes

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
@@ -159,7 +159,8 @@ class CodeWhispererPopupManager {
             return
         }
         val typeahead = resolveTypeahead(states, selectedIndex, typeaheadOriginal)
-        sessionContext = SessionContext(typeahead, typeaheadOriginal, selectedIndex, sessionContext.seen)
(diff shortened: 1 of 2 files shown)


manodnyab (approved):

Will-ShaoHua (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -159,7 +159,8 @@ class CodeWhispererPopupManager {
             return
         }
         val typeahead = resolveTypeahead(states, selectedIndex, typeaheadOriginal)
-        sessionContext = SessionContext(typeahead, typeaheadOriginal, selectedIndex, sessionContext.seen)
+        val isFirstTimeShowingPopup = indexChange == 0 && typeaheadChange.isEmpty()
```
zixlin7, line 162: just for my understanding, could there be cases that it's first time showing popup but typeahead is not empty?

andrewyuq: It's not possible by design, the typeahead only tracks the input after some recommendation has already displayed in the editor. The thing that user types before the completion shows is called `userInputSinceInvocation` which is a different thing.

PR #3484, opened 2023-02-15
Add a step to copy the user code in SSO/Builder ID flow

manodnyab:
## Description

Add an additional dialog containing the user verification code before opening the browser. Users must now enter in this code in the browser whereas before it was done automatically. The user code is copied to the clipboard.

 
## License


JadenSimon (approved):

guillaume-fetter:
Sorry to dig this up,but it appears as a downgrade in user experience from my (limited) point of view.
What is the context for this change? Security requirements?

PR #3486, opened 2023-02-16
Add Toolkit specific env var for SAM CLI commands

manodnyab:
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/executables/ExecutableCommon.kt
```diff
@@ -37,6 +38,10 @@ object ExecutableCommon {
                 this["LC_CTYPE"] = "UTF-8"
                 // we're not setting PYTHONIOENCODING because we might break SAM on py2.7
             }
+            if (executableName == "sam") {
```
rli, line 41: does this work on windows?

rli: maybe its cleaner to put the customization directly onto `ExecutableType`

rli (approved):

PR #3522, opened 2023-03-15
Disable CodeWhisperer integration tests on CI

rli:
required credentials not available in CI
 
## License


jetbrains-core/it/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererCodeScanIntegrationTest.kt
```diff
@@ -9,10 +9,8 @@ import org.junit.Test
 import software.aws.toolkits.jetbrains.services.codewhisperer.CodeWhispererTestUtil.cppFileName
 import software.aws.toolkits.jetbrains.services.codewhisperer.CodeWhispererTestUtil.cppTestLeftContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.codescan.sessionconfig.CodeScanSessionConfig
-import software.aws.toolkits.jetbrains.utils.rules.RunWithRealCredentials.RequiresRealCredentials
 import software.aws.toolkits.resources.message
 
-@RequiresRealCredentials
```
andrewyuq, line 15: IIRC, last time I checked if I don't have this annotation, the test will fail, does it matter when we choose to skip it for CI?

rli: oh i see what's happening here.
you need both the rule and annotation. will revert

andrewyuq (approved):

manodnyab (approved):

PR #3535, opened 2023-03-23
feat(codewhisperer): ignore white space in right context handling

zixlin7:
## Types of changes

## Description
To ignore white space and new lines when matching right context to handle more cases like this:
Screenshot 2023-03-22 at 4 27 55 PM

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
```diff
@@ -137,22 +137,23 @@ class CodeWhispererRecommendationManager {
...
-            } else if (rightContextFirstLine.isEmpty()) {
-                overlap(recommendationContent, rightContext)
+
+        // we will ignore white space and new lines when matching right context
+        val overlap = overlap(recommendationContent.trim(), rightContext.trim())
+        val overlapIndex = recommendationContent.lastIndexOf(overlap)
+        val truncated = if (overlapIndex >= 0) {
+            // if the truncated recommendation only contains white space we will return an empty string instead
+            if (recommendationContent.substring(0, overlapIndex).trim().isNotEmpty()) {
+                recommendationContent.substring(0, overlapIndex)
```
andrewyuq, line 147: can save this as a variable

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
```diff
@@ -137,22 +137,23 @@ class CodeWhispererRecommendationManager {
...
             } else {
-                val tempOverlap = overlap(recommendationContent, rightContext)
-                if (recommendationContent.substring(0, recommendationContent.length - tempOverlap.length).none { it == '\n' }) {
-                    tempOverlap
-                } else {
-                    ""
-                }
```
andrewyuq, line 152: actually the old logic can not be simply removed. This if-else logic handles cases like if the first line is not empty, then our right-context algorithm can't simply apply since it's wrong to match a suffix of a multi-line completion to the prefix of the right context when the first line of right context is not empty. There will be display issues.

andrewyuq: The logic should simply be
1. get the overlap with the trimmed completion and right context, using the old if-else logic.
2. get the overlapIndex using lastIndexOf()
3. simply return this truncated one `recommendationContent.substring(0, overlapIndex)`. The last check is not needed since it's checked in the caller.

zixlin7: added back check for first line of right context

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
```diff
@@ -50,10 +50,12 @@ class CodeWhispererRecommendationManager {
         }
         val rangeMarkers = mutableMapOf<RangeMarker, Reference>()
         recommendation.references().forEach {
+            val tempEnd = invocationStartOffset + it.recommendationContentSpan().end()
+            val referenceEnd = if (tempEnd <= endOffset) tempEnd else endOffset
             rangeMarkers[
                 tempDocument.createRangeMarker(
                     invocationStartOffset + it.recommendationContentSpan().start(),
-                    invocationStartOffset + it.recommendationContentSpan().end()
+                    referenceEnd
```
andrewyuq, line 58: This will give you an invalid rangeMarker error when the reference range is truncated entirely if the start of the reference is greater than endOffset, since we are creating the rangeMarker in the tempDocument, we are interested in the recommendation string that's being inserted to this document, which is already reformatted. Thus, in order to create rangeMarkers from this reformatted string, we have to remove all the ones that are outside of the truncated range. This means you have to check the following:
```
val referenceStart = invocationStartOffset + it.recommendationContentSpan().start()
if (referenceStart >= endOffset) return@forEach
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/toolwindow/CodeWhispererCodeReferenceManager.kt
```diff
@@ -70,8 +70,10 @@ class CodeWhispererCodeReferenceManager(private val project: Project) {
         val startOffset = caretPosition.offset
         val relativePath = getRelativePathToContentRoot(editor)
         reformattedDetail.references().forEachIndexed { i, reference ->
+            val recommendationEnd = startOffset + reformattedDetail.content().length
             val start = startOffset + reference.recommendationContentSpan().start()
-            val end = startOffset + reference.recommendationContentSpan().end()
+            val tempEnd = startOffset + reference.recommendationContentSpan().end()
+            val end = if (tempEnd <= recommendationEnd) tempEnd else recommendationEnd
```
andrewyuq, line 76: This is not needed if the reference truncation logic is handled before correctly. The reference.recommendationContentSpan().start() is guaranteed to be smaller than the recommendation length.

andrewyuq (commented):
And also, please add unit tests that covert the scenarios when:
right context is truncated:
1. there are no reference span in the removed range
2. there are reference spans that strictly overlaps the removed range,
3. there are reference spans that are entirely within the removed range

PR #3651, opened 2023-05-03
Fix coverage reporting

rli:
## License


jetbrains-ultimate/src-223+/software/aws/toolkits/jetbrains/services/codewhisperer/importadder/CodeWhispererJSImportUtil.kt
```diff
@@ -1,5 +1,6 @@
 // Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.services.codewhisperer.importadder
```
manodnyab, line 3: Should this be a part of this PR?

manodnyab (approved):

jetbrains-ultimate/src-223+/software/aws/toolkits/jetbrains/services/codewhisperer/importadder/CodeWhispererJSImportUtil.kt
rli: i found it because it showed up as a weird entry in the coverage report

PR #3661, opened 2023-05-10
Remove support for JetBrains Gateway v222 and v223 notification

manodnyab:
## Description

Display notification to customers on JB Gateway v222/223 about deprecation of support in the upcoming releases.

## License


jetbrains-gateway/resources/META-INF/ext-gateway.xml
```diff
@@ -8,7 +8,12 @@
     <extensions defaultExtensionNs="com.intellij">
         <applicationService serviceImplementation="software.aws.toolkits.jetbrains.gateway.connection.SshAgentService" />
         <applicationService serviceImplementation="software.aws.toolkits.jetbrains.gateway.connection.ThinClientTrackerService" />
+        <notificationGroup id="aws.toolkit_deprecation" displayType="BALLOON" key="aws.toolkit_deprecation.title" />
```
rli, line 11: hasn't this notification group been declared already?

jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/GatewayDeprecationNotice.kt
```diff
@@ -0,0 +1,35 @@
...
+package software.aws.toolkits.jetbrains.gateway
+
+import com.intellij.ide.AppLifecycleListener
+import com.intellij.notification.NotificationGroupManager
+import com.intellij.notification.NotificationType
+import com.intellij.openapi.application.ApplicationInfo
+import com.intellij.openapi.application.ApplicationNamesInfo
+import software.aws.toolkits.resources.message
+
+class GatewayDeprecationNotice : AppLifecycleListener {
```
rli, line 13: should we share logic with the one with the regular IDEs?

manodnyab: We can't reuse the logic that we do for the regular IDEs since the functions and classes used eg StartupActivity.runActivity and AppLifecycleListener.welcomeScreenDisplayed() are different and not the first one cannot be applied in Gateway. Most of the other logic used inside for comparing min version stays the same...

jetbrains-gateway/resources/META-INF/ext-gateway.xml
manodnyab: Changed names to avoid confusion, the notification here will have a different message hence creating a new group

rli: it's a different message, but the same category of content, so it's unclear why we need a new one

manodnyab: Makes sense, added the new notification to the original group.

rli (approved):

PR #3670, opened 2023-05-12
feat(codewhisperer): classifier for 4 languages

zixlin7:
## Types of changes

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererClassifierConstants.kt
```diff
@@ -16,221 +16,273 @@ import software.aws.toolkits.telemetry.CodewhispererAutomatedTriggerType
 
 object CodeWhispererClassifierConstants {
     val osMap: Map<String, Double> = mapOf(
-        "Mac OS X" to 0.102638,
-        "win32" to 0.040345,
-        "Windows 10" to 0.289530,
-        "Windows 7" to 0.086803,
+        "Mac OS X" to 0.0741,
```
rli, line 19: have we reconsidered being able to update weights without needing a toolkit release? or does that not make sense because it's also reliant on other client logic?

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererClassifierConstants.kt
zixlin7: yes we will be moving towards that direction, but for now it has to unfortunately

leigaol (approved):

PR #3692, opened 2023-05-25
Add retries to InvalidGrantException

andrewyuq:
The retries are added to all InvalidGrantException that makes sure that all the transient issues that are causing InvalidGrantException should be resolved by this retry. Using default retry stragegy with 2 retries(as the AWSSDK standard)
 that is 3 max attempts. With exponential backoff

Tested with making a request to throw InvalidGrantException and logs showing there are 3 attempts (1st initial attempt + 2 more retries)

The added test does real network calls, although it's designed to return InvalidGrantExceptions, this is actually an integration test so we can if we want to put it outside of unit tests. Mocking this is non-trivial since the retry logic is something deep inside the sdk.

related VSC change:
https://github.com/aws/aws-toolkit-vscode/pull/3498

## Types of changes

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/sso/bearer/BearerTokenProvider.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/sso/bearer/BearerTokenProvider.kt
@@ -9,6 +9,7 @@ import com.intellij.util.containers.orNull
 import software.amazon.awssdk.auth.credentials.AnonymousCredentialsProvider
 import software.amazon.awssdk.auth.token.credentials.SdkToken
 import software.amazon.awssdk.auth.token.credentials.SdkTokenProvider
+import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration
 import software.amazon.awssdk.core.retry.conditions.OrRetryCondition
 import software.amazon.awssdk.core.retry.conditions.RetryOnExceptionsCondition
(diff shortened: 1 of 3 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/sso/bearer/BearerTokenProvider.kt
```diff
@@ -213,5 +216,20 @@ private fun buildUnmanagedSsoOidcClient(region: String): SsoOidcClient =
             Region.of(region),
             clientCustomizer = ToolkitClientCustomizer { _, _, _, _, configuration ->
                 configuration.nullDefaultProfileFile()
+
+                // Get the existing RetryPolicy
+                val existingRetryPolicy = configuration.retryPolicy()
+
+                // Add InvalidGrantException to the RetryOnExceptionsCondition
```
rli, line 223: could we get a test for this?

andrewyuq: sure

rli (approved):

Will-ShaoHua (approved):

justinmk3:
A changelog entry that hints about the improvement ("less frequent re-login" or something like that) will be helpful for customers.

PR #3701, opened 2023-06-03
fix contextMerging not accounting userInput

Will-ShaoHua:
## Before

https://github.com/aws/aws-toolkit-jetbrains/assets/96078566/5af2362a-ec5b-457a-abb0-08e89b111a00

## After

https://github.com/aws/aws-toolkit-jetbrains/assets/96078566/6d24b510-2e8c-485d-a89b-c6c0664d8c4e

## Types of changes

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
```diff
@@ -141,7 +141,7 @@ class CodeWhispererRecommendationManager {
             }
 
             val reformatted = reformat(requestContext, truncated)
-            val isDiscardedByRightContextTruncationDedupe = truncated.content().isEmpty() || !seen.add(reformatted.content())
+            val isDiscardedByRightContextTruncationDedupe = truncated.content().isEmpty() || !seen.add(reformatted.content()) || truncatedContent == userInput
```
andrewyuq, line 144: So what's been done here is to check if the userinput is equal to the suggestion, after truncation. This actually means we should also do this if there's no truncation, since in the above we also just do `!it.content().startsWith(userInput)` but didn't include the strict equal case.

andrewyuq: this is on line 122.

Will-ShaoHua: yes make sense, updated but I think we still need to keep the condition `!it.content().startsWith(userInput)` right? @andrewyuq

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
```diff
@@ -135,13 +135,13 @@ class CodeWhispererRecommendationManager {
             val truncated = it.toBuilder()
                 .content(truncatedContent)
                 .build()
-            val isDiscardedByUserInputForTruncated = !truncated.content().startsWith(userInput)
+            val isDiscardedByUserInputForTruncated = !truncated.content().startsWith(userInput) || truncated.content() == userInput
```
Will-ShaoHua, line 138: so right now, the bug case you encountered will be captured by this statement

Will-ShaoHua: should I remove `tuncatedContent == userInput` at  L144 ?

Will-ShaoHua: The condition `truncated.content().isEmpty()` is safe to remvoe right? So if I understand it correctly, this one is supposed to be the one I just added (`truncatedContent == userInput`)?

andrewyuq: if truncated is empty, `!truncated.content().startsWith(userInput) || truncated.content() == userInput` this must be true

andrewyuq (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererRecommendationManager.kt
Will-ShaoHua: I see

PR #3779, opened 2023-07-26
CodeWhisperer CodeScan - Handle file outside project

abhikuhikar:
Handle case where the file a user wants to scan is outside the project root. Currently scanned file path relative to the project root is sent for code scan. When this file is outside the project root, relative path contains `..` which fails at CG processing. This is handled in this PR.    

## Types of changes

## Description

* When the file is outside the project root, include just that file in the zip archive. 
* Change project root to the parent of this above file as the file path relative to the project root is included in the zip archive. 
   * This will make sure that the relative path is correct (does not include `..`)
   * The updated project root will be correctly used when displaying the results back in the panel.   

File outside project
https://github.com/aws/aws-toolkit-jetbrains/assets/16646509/54c2ddd7-25b0-42d6-a244-205eb917a129

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/sessionconfig/CodeScanSessionConfig.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/sessionconfig/CodeScanSessionConfig.kt
@@ -65,7 +65,7 @@ sealed class CodeScanSessionConfig(
 
         LOG.debug { "Creating payload. File selected as root for the context truncation: ${selectedFile.path}" }
 
-        val (includedSourceFiles, payloadSize, totalLines, _) = when(selectedFile.path.startsWith(projectRoot.path)) {
+        val (includedSourceFiles, payloadSize, totalLines, _) = when (selectedFile.path.startsWith(projectRoot.path)) {
             true -> includeDependencies()
             false -> {
                 // Set project root as the parent of the selected file.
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/sessionconfig/CodeScanSessionConfig.kt
```diff
@@ -64,7 +65,14 @@ sealed class CodeScanSessionConfig(
 
         LOG.debug { "Creating payload. File selected as root for the context truncation: ${selectedFile.path}" }
 
-        val (includedSourceFiles, payloadSize, totalLines, _) = includeDependencies()
+        val (includedSourceFiles, payloadSize, totalLines, _) = when (selectedFile.path.startsWith(projectRoot.path)) {
+            true -> includeDependencies()
+            false -> {
+                // Set project root as the parent of the selected file.
+                projectRoot = selectedFile.parent
+                includeFileOutsideProjectRoot()
```
andrewyuq, line 73: did we align this behavior that we will scan only the current folder in this case?

abhikuhikar: It's only one file. Not even the entire folder. This is consistent with the current [VSC behavior](https://github.com/aws/aws-toolkit-vscode/blob/bdb0a832be5b48d1aae519bfc6602ef1a3229c13/src/codewhisperer/util/dependencyGraph/pythonDependencyGraph.ts#L180).

andrewyuq (approved):

PR #3780, opened 2023-07-26
fix(codewhisperer): revise condition for refresh token expiration in CW node

leigaol:
## Types of changes

## Description

After 1 hour of AWS Builder ID login, if the IDE is not restarted, the AWS Toolkit will show a yellow warning icon, claiming the connection is invalid & expired. 

But the connection is actually in needs refresh state and it can be automatically refreshed. 

## Analysis 

In the AWS Toolkit node, the connection expiration icon is shown based on the return from`lazyIsUnauthedBearerConnection`, which ultimately converts to `provider.state() != BearerTokenAuthState.AUTHORIZED or provider.currentToken() == null`.

However, this is not aligned with the connection expiration state condition CodeWhisperer node. 
In CW node, `checkActiveCodeWhispererConnectionType` is the check for connection expiration. 

```
val connectionType = CodeWhispererExplorerActionManager.getInstance().checkActiveCodeWhispererConnectionType(project)
        when (connectionType) { CodeWhispererLoginType.Sono -> {
                presentation.addText(message("codewhisperer.explorer.root_node.login_type.aws_builder_id"), SimpleTextAttributes.GRAY_ATTRIBUTES)
            }
        }
```

It then depends on 

```
isAccessTokenExpired(project) || isRefreshTokenExpired(project) -> CodeWhispererLoginType.Expired
```

while 

```
fun isRefreshTokenExpired(project: Project): Boolean {
        val state = tokenProvider.state()
        return state == BearerTokenAuthState.NOT_AUTHENTICATED
    }
```

## The Fix 

The desired behavior is: When access token needs a refresh, it should refresh by itself. 

When the state becomes NEEDS_REFRESH, the AWS Toolkit node should still show valid connection. 

In the latest commit, before the toolkit node is about to show connection expired, we ran a refresh in the background thread, if the state is still not authorized after the refresh, then the warning icon is showed.

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
@@ -207,7 +207,7 @@ fun AwsBearerTokenConnection.lazyIsUnauthedBearerConnection(): Boolean {
         }
 
         // or state is not authorized
-        return provider.state() != BearerTokenAuthState.AUTHORIZED
+        return provider.state() == BearerTokenAuthState.NOT_AUTHENTICATED
     }
 
     // not a bearer token provider
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
@@ -120,7 +120,7 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
 
     fun checkActiveCodeWhispererConnectionType(project: Project) = when {
         actionState.token != null -> CodeWhispererLoginType.Accountless
-        isAccessTokenExpired(project) || isRefreshTokenExpired(project) -> CodeWhispererLoginType.Expired
+        isRefreshTokenExpired(project) -> CodeWhispererLoginType.Expired
         else -> {
             val conn = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(CodeWhispererConnection.getInstance())
             if (conn != null) {
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
@@ -20,7 +20,6 @@ import software.aws.toolkits.jetbrains.core.explorer.refreshDevToolTree
 import software.aws.toolkits.jetbrains.services.codewhisperer.credentials.CodeWhispererLoginType
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererConstants
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererUtil.getConnectionStartUrl
-import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererUtil.isAccessTokenExpired
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererUtil.isRefreshTokenExpired
 import software.aws.toolkits.telemetry.AwsTelemetry
 import java.time.LocalDateTime


andrewyuq:
>It then depends on
> isAccessTokenExpired(project) || isRefreshTokenExpired(project) -> CodeWhispererLoginType.Expired

`isAccessTokenExpired()` checks if the state is `NEEDS_REFRESH`. So the warning icon will show as long as it's not fully authenticated (need refresh or reauth through browser flow).

Will-ShaoHua:
anotehr root cause I feel we didn't refresh the conneciton somewhere

Will-ShaoHua:
@leigaol 
i feel we might miss both 
- UI update
- refresh connection

leigaol:
> So regarding this, the toolkit icon has the same updating condition as the CodeWhisperer status icon. I think somewhere we are missing sending such an event to the Toolkit status when the accessToken gets successfully refreshed.

When toolkit shows the warning, I can do a manual trigger in CWSPR and then the yellow warning goes away. 

I think what confuses user is the 1 hour yellow warning icon. They don't expect to see it per hour.

rli (approved):
We can't tell if a refresh token is valid or not, so i think it's acceptable to assume it's valid until time of use.

Maybe we need to have a sentinel value to tell if a refresh token is definitely invalid (because we attempted to use it and got failure)

andrewyuq:
>> So regarding this, the toolkit icon has the same updating condition as the CodeWhisperer status icon. I think somewhere we are missing sending such an event to the Toolkit status when the accessToken gets successfully refreshed.

> When toolkit shows the warning, I can do a manual trigger in CWSPR and then the yellow warning goes away.
I think what confuses user is the 1 hour yellow warning icon. They don't expect to see it per hour.

If we do this, we should also make the change to the following code:

```
override fun getIcon(): Icon =
        if (isCodeWhispererExpired(project)) {
            AllIcons.General.BalloonWarning
        } else if (CodeWhispererInvocationStatus.getInstance().hasExistingInvocation()) {
            AnimatedIcon.Default()
        } else {
            AllIcons.Actions.Commit
        }
```

This code for CodeWhisperer icon needs to align with the change in this PR. it should change to show the Balloon warning only if the refresh token is expired.

Will-ShaoHua:
@rli are we able to refresh the connections at @ `DefaultToolkitAuthManager.loadState()`, not sure if we can cuz it's a JB API and we probably should not call network API within there?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ConnectionSettingsMenuBuilder.kt
```diff
@@ -80,7 +87,22 @@ class ConnectionSettingsMenuBuilder private constructor() {
...
+                                            } catch (error: Exception) {
+                                                LOG.info {"Failed to refresh access token $error"}
+                                            }
+                                        }
+                                        // show warning icon if connection is in not authorized state.
+                                        if (provider?.state() != BearerTokenAuthState.AUTHORIZED) {
+                                            e.presentation.icon = AllIcons.General.Warning
+                                        }
+
+                                    }
```
Will-ShaoHua, line 105:
```suggestion
                                    ApplicationManager.getApplication().executeOnPooledThread {
                                        val provider = (value.getConnectionSettings().tokenProvider.delegate as? BearerTokenProvider)
                                        if (provider != null && provider.state() == BearerTokenAuthState.NEEDS_REFRESH) {
                                            try {
                                                provider.resolveToken()
                                                BearerTokenProviderListener.notifyCredUpdate(provider.id)
                                            } catch (error: Exception) {
                                                LOG.info {"Failed to refresh access token $error"}
                                            }
                                        }
                                    }.get()
                                    
                                    if (provider?.state() != BearerTokenAuthState.AUTHORIZED) {
                                            e.presentation.icon = AllIcons.General.Warning
                                     }
```

Will-ShaoHua: I think it's something like this? @andrewyuq @rli

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ConnectionSettingsMenuBuilder.kt
```diff
@@ -80,7 +87,22 @@ class ConnectionSettingsMenuBuilder private constructor() {
...
+                                                provider.resolveToken()
+                                                BearerTokenProviderListener.notifyCredUpdate(provider.id)
+                                            } catch (error: Exception) {
+                                                LOG.info {"Failed to refresh access token $error"}
+                                            }
+                                        }
+                                        // show warning icon if connection is in not authorized state.
+                                        if (provider?.state() != BearerTokenAuthState.AUTHORIZED) {
+                                            e.presentation.icon = AllIcons.General.Warning
+                                        }
```
Will-ShaoHua, line 103: we should move L102 out of `executeOnPoolThread`

Will-ShaoHua: and call a `get()` like this `ApplicationManager.getApplication().executeOnPooledThread {}.get()`

Will-ShaoHua: since `executeOnPooledThread` returns a `Future`(similar to async/await) so we just await here by calling `get()`

andrewyuq: @Will-ShaoHua will `get()` block the `update` execution?

rli: this can run once a second so this is probably not the right place for this

Will-ShaoHua: maybe we should just did what you previously did @leigaol 
```suggestion
                                        if (provider?.state() == BearerTokenAuthState.NOT_AUTHENTICATED) {
                                            e.presentation.icon = AllIcons.General.Warning
                                        }
```

Will-ShaoHua: and add a `tryRefresh` somewhere else

Will-ShaoHua: when do we update this connectionMenu?

Will-ShaoHua: I think let's stick with the original change here in this PR
```
if (provider?.state() != BearerTokenAuthState.AUTHORIZED) {
                                            e.presentation.icon = AllIcons.General.Warning
                                        }
```

andrewyuq: This is the place where the "More" action icon in the aws toolkit panel `...` updates its icon.
```
class MoreConnectionActionsAction : DumbAwareAction(AllIcons.Actions.MoreHorizontal) {
    override fun update(e: AnActionEvent) {
        e.presentation.icon = if (lazyGetUnauthedBearerConnections().isEmpty()) AllIcons.Actions.MoreHorizontal else AllIcons.General.Warning
    }
    ```

the place here where your change has made is where the "Balloon Warning" icon can appear
```
override fun update(e: AnActionEvent) {
                                super.update(e)
                                if (value.lazyIsUnauthedBearerConnection()) {
                                    e.presentation.icon = AllIcons.General.Warning
```

So, it seems like these two places both use `lazyIsUnauthedBearerConnection()` to check if it needs to have a balloon warning icon. I would make one of the two changes below:
1. in each of the above logic, do manual refresh once(non-blocking for EDT) and update the icon to warning only after the refresh fails. 
2. like what you have initially changed, relax the condition such that if it's accessToken expired it would still show ok icon instead of the warning icon. And let the applications refresh token on usage(CW etc.)

Will-ShaoHua: besides the change above, can we spread one more `CodeWhispererUtil.promptReAuth` at `CodeWhispererService`?

Will-ShaoHua: IIRC we used to have this when user tries to trigger getRecommendation

Will-ShaoHua: right now we only have this at projectStartup

andrewyuq: we still have this on line 91 of the file CodeWhispererService.kt if that is what you are talking about.

Will-ShaoHua: oh ok found it

leigaol: OK I just reverted it back to 

> like what you have initially changed, relax the condition such that if it's accessToken expired it would still show ok icon instead of the warning icon. And let the applications refresh token on usage(CW etc.)


Doing an API call in a UX update function is expensive and not ideal.

andrewyuq: can you also change the relax CW checking condition then so it matches the changes here?

leigaol: Do you mean this code? 
`
    fun checkActiveCodeWhispererConnectionType(project: Project) = when {
        actionState.token != null -> CodeWhispererLoginType.Accountless
        isAccessTokenExpired(project) || isRefreshTokenExpired(project) -> CodeWhispererLoginType.Expired`

andrewyuq: yes

andrewyuq (approved):

rli (approved):

PR #3785, opened 2023-07-28
Deprecation notification for Gateway 231

manodnyab:
## License


rli (approved):
let's drop 222 for regular IDEs too?

PR #3800, opened 2023-08-08
use extension point for language specific functionality

Will-ShaoHua:
## Types of changes

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/languages/CodeWhispererJavaScript.kt
```diff
@@ -4,10 +4,13 @@
 package software.aws.toolkits.jetbrains.services.codewhisperer.language.languages
 
 import software.aws.toolkits.jetbrains.services.codewhisperer.language.CodeWhispererProgrammingLanguage
+import software.aws.toolkits.jetbrains.services.codewhisperer.util.FileCrawler
+import software.aws.toolkits.jetbrains.services.codewhisperer.util.JavascriptCodeWhispererFileCrawler
 import software.aws.toolkits.telemetry.CodewhispererLanguage
 
 class CodeWhispererJavaScript private constructor() : CodeWhispererProgrammingLanguage() {
     override val languageId: String = ID
+    override val fileCrawler: FileCrawler = JavascriptCodeWhispererFileCrawler
```
rli, line 13: all the JS-related extensions and file crawlers should live in jetbrains-ultimate

rli (approved):

andrewyuq (approved):
change looks good but not sure if they are related to extension points.

Will-ShaoHua:
> change looks good but not sure if they are related to extension points.

I found out if I move all JS relevant code path to Ultimate, the file diff will be 40 files+, will hold this PR @andrewyuq

Will-ShaoHua:
The reason we need file crawler to live in CodeWhispererProgrammingLangauge(which is an extension point) is because if we don't do that, `PyFile` in https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/PythonCodeWhispererFileCrawler.kt#L32 will not exist in the class path and it will cause runtime error. That being said if users are using intellij community + not install python plugin, the plugin will throw class `PyFile` not found error. More than that, I think it will cause CW to crash unless users restart the IDE

Will-ShaoHua:
Therefore I am thinking to disable L132 at https://github.com/aws/aws-toolkit-jetbrains/blob/7eda113804f1d2fdc180ec08675c200bf748ef01/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererFileCrawler.kt#L132C115-L132C115 as a short term mitigation to fix this issue. And we will restructure and make all languages related code path live in its corresponding module. (`CodeWhispererLanguageManager.getLanguage()` `JsCodeScan` and relevant tests)

Will-ShaoHua:
> change looks good but not sure if they are related to extension points.

I think my previous statement didn't answer your question, yea the change is to put `FileCrawler` under `CodeWhispererProgrammingLanguage` which is already an extension point therefore `PythonFileCrawler` should only appears when `CodeWhispererPython` lang exists which is the case if users' IDE recognize `.py` file

andrewyuq:
> > change looks good but not sure if they are related to extension points.
> 
> oh my previous previous didn't answer your question, yea the change is to put `FileCrawler` under `CodeWhispererProgrammingLanguage` which is already an extension point therefore `PythonFileCrawler` should only appears when `CodeWhispererPython` lang exists which is the case if users' IDE recognize `.py` file

yes, this sounds good.

Will-ShaoHua:
will open a followup PR to 
- add a `FallbackFileCrawler` which will not use language specific API like `ClassOwner` or `PytFile`
- Move all FileCrawler logic into `language` package

PR #3832, opened 2023-08-28
Prompt users  to continue their Dev Environment session

manodnyab:
## Description

Ask users if they would like to continue working in their dev env after specified minutes of inactivity and extend the duration of the Dev Env session.

## License


jetbrains-ultimate/tst/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcherTest.kt
```diff
@@ -0,0 +1,7 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.caws
+
+class DevEnvStatusWatcherTest {
```
rli, line 6: test?

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+    companion object {
+        fun getInstance(project: Project) = project.service<DevEnvStatusWatcher>()
+    }
+
+    override fun runActivity(project: Project) {
+        val connection = SonoCredentialManager.getInstance(project).getConnectionSettings()
+            ?: error("Failed to fetch connection settings from Dev Environment")
+        val envId = System.getenv(CawsConstants.CAWS_ENV_ID_VAR) ?: error("envId env var null")
+        val org = System.getenv(CawsConstants.CAWS_ENV_ORG_NAME_VAR) ?: error("space env var null")
+        val projectName = System.getenv(CawsConstants.CAWS_ENV_PROJECT_NAME_VAR) ?: error("project env var null")
```
rli, line 36: should we make sure this only runs on dev envs? we don't want to error out or run this in a normal IDE environment

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+
+                        if (ans) {
+                            notifyBackendOfActivity()
+                            lastControllerActivity = actualInactivityDuration
+                        }
+                    } catch (e: Exception) {
+                        notifyError(e.message.toString())
+                    }
+                }
+                Thread.sleep(30000)
```
rli, line 83: if we're using coroutines, use `delay` instead of Thread#sleep

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+                            }
+                            return@runBlocking continueWorking
+                        }
+
+                        if (ans) {
+                            notifyBackendOfActivity()
+                            lastControllerActivity = actualInactivityDuration
+                        }
+                    } catch (e: Exception) {
+                        notifyError(e.message.toString())
```
rli, line 80: log error too

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+        val coroutineScope = projectCoroutineScope(project)
+        coroutineScope.launch(getCoroutineBgContext()) {
+            val initialEnv = client.getDevEnvironment {
+                it.id(envId)
+                it.spaceName(org)
+                it.projectName(projectName)
+            }
+            val inactivityTimeout = initialEnv.inactivityTimeoutMinutes()
+            if (inactivityTimeout == 0) {
+                return@launch
```
rli, line 47: log

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+            }
+            val inactivityTimeoutInSeconds = inactivityTimeout * 60
+            var actualInactivityDuration: Long = 0
+            var lastControllerActivity: Long = 0
+            while (true) {
+                val statusJson = UnattendedStatusUtil.getStatus()
+                val lastActivityTime = statusJson.projects?.first()?.secondsSinceLastControllerActivity ?: 0
+                actualInactivityDuration = if (lastActivityTime > lastControllerActivity) {
```
rli, line 55: what if we queried the environment API for lastActivityTime instead? that is the ultimate source of truth anyways

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+                }
+                Thread.sleep(30000)
+            }
+        }
+    }
+
+    private fun notifyBackendOfActivity() {
+        val request = UpdateActivityRequest(
+            timestamp = System.currentTimeMillis().toString()
+        )
```
rli, line 91: we don't need this since the environment api will automatically infer the timestamp

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: We are updating the environment API only when there is activity.
This lastActivityTime is the one returned by the IDE

jetbrains-ultimate/tst/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcherTest.kt
manodnyab: Added tests

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,109 @@
...
+
+class DevEnvStatusWatcher : StartupActivity {
+
+    companion object {
+        fun getInstance(project: Project) = project.service<DevEnvStatusWatcher>()
+        private val LOG = getLogger<DevEnvStatusWatcher>()
+    }
+
+    override fun runActivity(project: Project) {
+        if (!isRunningOnRemoteBackend()) {
```
rli, line 36: this isn't quite right -- we only want to run in a CodeCatalyst environment

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,109 @@
...
+            var lastControllerActivity: Long = 0
+            while (true) {
+                val statusJson = UnattendedStatusUtil.getStatus()
+                val lastActivityTime = statusJson.projects?.first()?.secondsSinceLastControllerActivity ?: 0
+                actualInactivityDuration = getActualInactivityDuration(lastActivityTime, lastControllerActivity)
+
+                if (actualInactivityDuration >= (inactivityTimeoutInSeconds - 300)) {
+                    try {
+                        val inactivityDurationInMinutes = actualInactivityDuration / 60
+                        val ans = runBlocking {
```
rli, line 68: we shouldn't need runBlocking here

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
rli: i'm saying we can query the environment API for the actual duration since last activity instead of trying to calculate it

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,109 @@
...
+                        LOG.error(preMessage + ": " + e.message)
+                        notifyError(preMessage, e.message.toString())
+                    }
+                }
+                delay(30000)
+            }
+        }
+    }
+
+    fun getActualInactivityDuration(
```
rli, line 93: this needs a better name because "get" functions don't typically have side effects

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,109 @@
...
+                                    message("caws.devenv.continue.working.after.timeout.title"),
+                                    message("caws.devenv.continue.working.after.timeout", inactivityDurationInMinutes)
+                                ).ask(project)
+                            }
+                            return@runBlocking continueWorking
+                        }
+
+                        if (ans) {
+                            notifyBackendOfActivity()
+                            lastControllerActivity = actualInactivityDuration
```
rli, line 80: won't this cause some desync since we're sending the environment API one time, but storing a slightly different one?

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,109 @@
...
+                        }
+                    } catch (e: Exception) {
+                        val preMessage = "Error while checking if Dev Environment should continue working"
+                        LOG.error(preMessage + ": " + e.message)
```
rli, line 84:
```suggestion
                        LOG.error(e) { preMessage }
```

jetbrains-ultimate/resources/META-INF/ext-codewithme.xml
```diff
@@ -10,6 +10,9 @@
 
         <postStartupActivity implementation="software.aws.toolkits.jetbrains.services.caws.DevfileWatcher"/>
         <applicationService serviceImplementation="software.aws.toolkits.jetbrains.services.caws.DevfileWatcher"/>
+        <postStartupActivity implementation="software.aws.toolkits.jetbrains.services.caws.DevEnvStatusWatcher"/>
+        <projectService serviceImplementation="software.aws.toolkits.jetbrains.services.caws.DevEnvStatusWatcher"/>
```
rli, line 14: this doesn't need to be a service. just instantiate it directly for tests

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: Yeah, it may cause a couple of millisecond desync, let me change that to one value

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: the runBlocking is reqd to execute withContext. We need to run the dialog in ui context and then pass the result of the dialog as other actions will depend on this result

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: Got it, Adding a check for the CodeCatalyst env id  constant

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: Changed the implementation to query value off the environment api

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: It does not infer the timestamp correctly, value needs to be provided here

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: This function has been removed in the new implementation

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,112 @@
...
+import java.time.Instant
+
+class DevEnvStatusWatcher : StartupActivity {
+
+    companion object {
+        private val LOG = getLogger<DevEnvStatusWatcher>()
+    }
+
+    override fun runActivity(project: Project) {
+        if (!isRunningOnRemoteBackend() && System.getenv(CawsConstants.CAWS_ENV_ID_VAR) == null) {
```
rli, line 36:
```suggestion
        if (System.getenv(CawsConstants.CAWS_ENV_ID_VAR) == null) {
```
is sufficient

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,112 @@
...
+            if (inactivityTimeout == 0) {
+                LOG.info("Dev environment inactivity timeout is 0, not monitoring")
+                return@launch
+            }
+            val inactivityTimeoutInSeconds = inactivityTimeout * 60
+
+            // ensure the JetBrains inactivity tracker and the activity api are in sync
+            val jbActivityStatusJson = UnattendedStatusUtil.getStatus()
+            val jbActivityStatus = jbActivityStatusJson.projects?.first()?.secondsSinceLastControllerActivity ?: 0
+            notifyBackendOfActivity((System.currentTimeMillis() - (jbActivityStatus * 1000)).toString())
```
rli, line 62: let's get rid of all this math to make this easier to read
for example:
```kt
Instant.now().minus(jbActivityStatus, ChronoUnit.SECONDS)
```
```

File: jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt [hunk 16]
```diff
@@ -0,0 +1,112 @@
...
+            while (true) {
+                val statusJson = UnattendedStatusUtil.getStatus()
+                val lastActivityTime = statusJson.projects?.first()?.secondsSinceLastControllerActivity ?: 0
+                if (lastActivityTime < secondsSinceLastControllerActivity) {
+                    // update the API in case of any activity
+                    notifyBackendOfActivity((System.currentTimeMillis() - (lastActivityTime * 1000)).toString())
+                }
+                secondsSinceLastControllerActivity = lastActivityTime
+
+                val lastRecordedActivityTime = CawsEnvironmentClient.getInstance().getActivity().timestamp ?: System.currentTimeMillis().toString()
```
rli, line 74: might be a bad idea to silently ignore here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/GetActivityResponse.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.caws.envclient.models
+
+data class GetActivityResponse(
+    val timestamp: String?
```
rli, line 7: should this be nullable?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/CawsEnvironmentClient.kt
```diff
@@ -88,6 +91,27 @@ class CawsEnvironmentClient(
         return objectMapper.readValue(response.entity.content)
     }
 
+    fun getActivity(): GetActivityResponse = try {
+        val request = HttpGet("$endpoint/activity")
+        val response = execute(request)
+        objectMapper.readValue<GetActivityResponse>(response.entity.content)
+    } catch (e: Exception) {
+        LOG.error(e) { "Couldn't parse response from /activity API" }
+        GetActivityResponse(timestamp = null)
```
rli, line 100: if we can't parse the response, we might not want to run any heartbeat logic?
one case where we can't parse is when the environment is not in heartbeat timeout mode

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: We need a runBlocking here to be able to return the response correctly

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: Added a log statement and `return` if we don't have a response from the api

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/GetActivityResponse.kt
manodnyab: No, changed it

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/CawsEnvironmentClient.kt
manodnyab: Returning a null here in case we cant parse a response which is then handled in the main code

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,120 @@
...
+            val inactivityTimeoutInSeconds = inactivityTimeout * 60
+
+            // ensure the JetBrains inactivity tracker and the activity api are in sync
+            val jbActivityStatusJson = UnattendedStatusUtil.getStatus()
+            val jbActivityStatus = jbActivityStatusJson.projects?.first()?.secondsSinceLastControllerActivity ?: 0
+            notifyBackendOfActivity((getActivityTime(jbActivityStatus).toString()))
+            var secondsSinceLastControllerActivity = jbActivityStatus
+
+            while (true) {
+                val statusJson = UnattendedStatusUtil.getStatus()
```
rli, line 68: can we have unit tests on this inner loop

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,120 @@
...
+                                return@withContext MessageDialogBuilder.okCancel(
+                                    message("caws.devenv.continue.working.after.timeout.title"),
+                                    message("caws.devenv.continue.working.after.timeout", inactivityDurationInMinutes)
+                                ).ask(project)
+                            }
+                            return@runBlocking continueWorking
+                        }
+
+                        if (ans) {
+                            notifyBackendOfActivity(getActivityTime(0).toString())
```
rli, line 99: nit: not obvious what the `0` means. maybe make 0 the default value and rename the parameter name to "secondsInThePast" or sometihng

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,120 @@
...
+                        LOG.error(e) { preMessage }
+                        notifyError(preMessage, e.message.toString())
+                    }
+                }
+                delay(30000)
+            }
+        }
+    }
+
+    private fun notifyBackendOfActivity(timestamp: String = Instant.now().toEpochMilli().toString()) {
```
rli, line 112: i'd probably just make this take a `Long` and always do `toString()`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/GetActivityResponse.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.caws.envclient.models
+
+data class GetActivityResponse(
+    val timestamp: String
```
rli, line 7: deserialization should automatically cast to Long?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/UpdateActivityRequest.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.caws.envclient.models
+
+data class UpdateActivityRequest(
+    val timestamp: String? = null
```
rli, line 7: does this actually need to be a string or will the environment API automatically accept it as a Long?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/GetActivityResponse.kt
manodnyab: The API takes a string value so no cast required

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/models/UpdateActivityRequest.kt
manodnyab: The API takes a string value so no cast required

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
manodnyab: Done, changed the implementation to have zero as the default value along with a change in the variable name

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/CawsEnvironmentClient.kt
```diff
@@ -88,6 +91,30 @@ class CawsEnvironmentClient(
         return objectMapper.readValue(response.entity.content)
     }
 
+    fun getActivity(): GetActivityResponse? = try {
+        val request = HttpGet("$endpoint/activity")
+        val response = execute(request)
+        if (response.statusLine.statusCode == 400) {
+            throw Exception("Inactivity tracking may not enabled")
```
rli, line 98: prefer log + return null here instead of exception-driven logic

jetbrains-ultimate/tst/software/aws/toolkits/jetbrains/services/DevEnvStatusWatcherTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        val devEnvStatusWatcher = spy<DevEnvStatusWatcher>(sut) {
+            doReturn(600.toLong()).whenever(it).getJbRecordedActivity()
+            doReturn(null).whenever(it).getLastRecordedApiActivity()
+        }
+        val response = devEnvStatusWatcher.checkHeartbeat(0, 0, projectRule.project)
+        assertThat(response.first).isTrue()
+    }
+
+    @Test
+    fun `API is called if user extends the timeout 5 minutes before inactivity timeout`() {
```
rli, line 36: API is not called if user rejects?

jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/caws/DevEnvStatusWatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+                secondsSinceLastControllerActivity = response.second
+            }
+        }
+    }
+
+    fun checkHeartbeat(
+        secondsSinceLastControllerActivity: Long,
+        inactivityTimeoutInSeconds: Int,
+        project: Project
+    ): Pair<Boolean, Long> {
```
rli, line 78: add a doc string on what these return values signify

rli (approved):

jetbrains-ultimate/tst/software/aws/toolkits/jetbrains/services/DevEnvStatusWatcherTest.kt
manodnyab: No, it isn't cause at that point we dont want to update the API. The API at this point will have the last recorded activity. Since the user doesn't want to continue. the session if they reject it, we needn't update the timestamp.

manodnyab: Added a test for the condition "if the user rejects"

PR #3856, opened 2023-09-08
Always pin connection on connection change to reduce user confusion

rli:
Per PM feedback

## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/pinning/ConnectionPinningManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/pinning/ConnectionPinningManager.kt
@@ -3,28 +3,19 @@
 
 package software.aws.toolkits.jetbrains.core.credentials.pinning
 
-import com.intellij.icons.AllIcons
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
-import com.intellij.openapi.application.ModalityState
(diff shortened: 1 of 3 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/pinning/ConnectionPinningManager.kt
```diff
@@ -93,32 +90,27 @@ class DefaultConnectionPinningManager :
...
             "${features.dropLast(1).joinToString(",") { it.featureName }} and ${features.last().featureName}"
         }
 
-        var connectionToPin = if (oldConnection is AwsBearerTokenConnection) oldConnection else newConnection
-        if (showDialogIfNeeded(oldConnection, newConnection, featuresString)) {
-            features.forEach {
-                setPinnedConnection(it, connectionToPin)
-            }
-            notifyInfo(message("credentials.switch.notification.title", featuresString, connectionToPin.label))
+        val connectionToPin = if (oldConnection is AwsBearerTokenConnection) oldConnection else newConnection
```
manodnyab, line 100: In the case where both Identity Center and Builder ID can be used and we are switching connections from Builder ID to Identity Center, should we still pin the connection?

rli: right, we keep the connection they started with

manodnyab (approved):

PR #3877, opened 2023-09-22
Populate Identity Center connection entries directly from profile file

rli:
In addition to adding a lot of tests that were previously missing from the `sso-session` logic, there are a couple of things we need to accomplish.

* CredentialManager is now aware of `sso-session`. This was done to open the door to combining `ToolkitAuthManager` and `CredentialManager` in the future, and also to ensure that both systems have the same view of the available `sso-session` profiles, which...
* allows us to reuse the same connection across the sigv4 and bearer token features.
* Mutating a `sso-session` invalidates all sigv4 profiles referencing it for consistency, though the scope upgrade logic will be tackled separately

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/DefaultToolkitAuthManager.kt
```diff
@@ -30,11 +34,68 @@ class DefaultToolkitAuthManager : ToolkitAuthManager, PersistentStateComponent<T
             )
         }
 
-        factoryConnections.toList()
+        factoryConnections.toMutableSet()
+    }
+
+    init {
+        // initial load then subscribe to bus for future changes
```
manodnyab, line 41: Would adding this as a listener work instead?

manodnyab (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/DefaultToolkitAuthManager.kt
rli: yes, the listener does this, but we don't want to worry about the order of which service gets initialized first

PR #3881, opened 2023-09-25
fix expire connection able to call recommendation API issue

panshao612:
While connection expired(session token expired).If restart IDE or after one hour of the time pass, prompt showed to CX that need reconnect to CodeWhisperer. If CX don’t reconnect, they won’t see the prompt but they are able to manual or auto trigger recommendation request. 
Root cause found that auth logic in recommendation request call bind with show reAuth notification logic. After first time showing the connection expire notification, CX are able to trigger recommendation with expired connection. 

https://t.corp.amazon.com/V1009966626/communication

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
@@ -21,6 +21,7 @@ import software.aws.toolkits.jetbrains.core.credentials.pinning.CodeWhispererCon
 import software.aws.toolkits.jetbrains.core.credentials.sono.isSono
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.BearerTokenAuthState
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.BearerTokenProvider
+import software.aws.toolkits.jetbrains.core.explorer.refreshDevToolTree
 import software.aws.toolkits.jetbrains.services.codewhisperer.actions.CodeWhispererLoginLearnMoreAction
 import software.aws.toolkits.jetbrains.services.codewhisperer.actions.CodeWhispererSsoLearnMoreAction
 import software.aws.toolkits.jetbrains.services.codewhisperer.actions.ConnectWithAwsToContinueActionError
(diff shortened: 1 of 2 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
```diff
@@ -196,12 +196,13 @@ object CodeWhispererUtil {
...
     fun promptReAuth(project: Project, isPluginStarting: Boolean = false): Boolean {
-        if (CodeWhispererService.hasReAuthPromptBeenShown()) return false
         if (!isCodeWhispererExpired(project)) return false
         val tokenProvider = tokenProvider(project) ?: return false
         return maybeReauthProviderIfNeeded(project, tokenProvider) {
             runInEdt {
-                notifyConnectionExpiredRequestReauth(project)
+                if (!CodeWhispererService.hasReAuthPromptBeenShown()) {
+                    notifyConnectionExpiredRequestReauth(project)
+                }
```
Will-ShaoHua, line 205: I prefer adding `if (isCodeWhispererExpired(project)) return` directly at `CodeWhispererService.kt` right after where we invoke `promptReauth` 

The reasoning is that our goal here is to handle cases when
1. users have seen "reconnect" UI and didn't not take action to the expired connection
2. in such schenario, we want to block upcoming triggers from invoking the real service

I think the `promptReauth` the function name itself suggest its functionality is to determine if we need to prompt the UI but not to terminate the logic from proceeding.

panshao612: Ditto. I also feel like adding a seperate check out side of promptReAuth suit the common sense to me. @andrewyuq What's you thought on this issue?

Will-ShaoHua: yes that works too, there are only 2 places calling this function, not a blocker.

Will-ShaoHua: you can also leave it as is, I am fine with that. Either way, we need to test it. Try edit the access token file in the sso/cache, try 
1. edit access token expiration date
2. edit access token expiration date + refresh token 

that's it then we are good to go

panshao612: Add project.refreshDevToolTree() and did functional test:
Connection expired prompt show up once at restart IDE, also show up at the first time trigger recommendation.
Later when trigger recommendation at expired state, abort the request and show no message to user.

Will-ShaoHua (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
andrewyuq: > I prefer adding if (isCodeWhispererExpired(project)) return directly at CodeWhispererService.kt right after 
> where we invoke promptReauth

this works for me.

andrewyuq: `project.refreshDevToolTree()` inside `maybeReauthProviderIfNeeded` makes sense to me. Since in all "fail to refresh" cases, we are calling the callback inside `maybeReauthProviderIfNeeded`.

panshao612: status bar show expired already before adding refreshDevToolTree

andrewyuq: I think this is good to go.

andrewyuq (approved):

PR #3884, opened 2023-09-26
Remove unusued dsl v1 components

manodnyab:
## Description

With the migration from dsl v1 to v3 we can remove unused dsl v2 components

## License


rli (approved):
Did you want to ban `com.intellij.ui.layout.*` while you're at it?

PR #3891, opened 2023-09-29
Show auth success/fail banner and 'Got it' tooltip after auth flow

rli:
image

image

image

 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsToolkitExplorerToolWindow.kt
```diff
@@ -68,52 +65,27 @@ class AwsToolkitExplorerToolWindow(private val project: Project) : SimpleToolWin
...
+        if (component != null) {
+            tabPane.selectedComponent = tabPane.getComponentAt(index)
 
-        // otherwise select the tab
-        content?.let {
-            tabPane.selectedIndex = index
+            return component
         }
+
+        return null
```
manodnyab, line 88: We won't get this null case but are we handling it anywhere on the off chance that we do?

manodnyab (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AwsToolkitExplorerToolWindow.kt
rli: no, let's handle when a consumer actually cares about the result

PR #3915, opened 2023-10-13
Save last selected explorer tab

rli:
Persist the last selected tab so the user sees the same thing on IDE restart
 
## License


jetbrains-core/tst/software/aws/toolkits/jetbrains/core/explorer/AwsToolkitExplorerToolWindowTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        assertThat(sut.state.selectedTab).isEqualTo(AwsToolkitExplorerToolWindow.DEVTOOLS_TAB_ID)
+    }
+
+    @Test
+    fun `handles loading invalid state`() {
+        (ToolWindowManager.getInstance(projectExtension.project) as ToolWindowHeadlessManagerImpl)
+            .doRegisterToolWindow(AwsToolkitExplorerFactory.TOOLWINDOW_ID)
+        val sut = runInEdtAndGet { AwsToolkitExplorerToolWindow(projectExtension.project) }
+
+        sut.loadState(
```
manodnyab, line 64: Should this assert that sut.loadState does not throw exceptions

rli: test will automatically fail if it throws an exception

manodnyab (approved):

PR #3928, opened 2023-10-18
Move 'Got it' popup to appropriate tree entry

rli:
`GotitTooltip#show` takes a component and a point transformer to determine the location to draw the popup. JTree does not expose the underlying JComponent for a given tree entry, however, we can determine the bounds of a node and calculate its offset to the tree.

image

 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/devToolsTab/DevToolsToolWindow.kt
```diff
@@ -154,8 +159,17 @@ class DevToolsToolWindow(private val project: Project) : SimpleToolWindowPanel(t
...
 
-    fun makeServiceChildrenVisible() {
-        TreeUtil.expand(tree, 2)
+    fun showGotIt(node: String?, tooltip: GotItTooltip) {
+        TreeUtil.promiseExpand(tree, 2).onSuccess {
+            node ?: return@onSuccess
+            treeModel.invoker.invoke {
+                val path = treeModel.asSequence().firstOrNull { (it.userObject as? AbstractTreeNode<*>)?.value == node }?.path ?: return@invoke
+                runInEdt {
+                    val pointToRight = tree.getPathBounds(TreePath(path))?.let { Point(it.x + it.width, it.y + it.height / 2) } ?: return@runInEdt
```
manodnyab, line 168: Would the width of the panel affect the got it tooltip?

manodnyab:
Does it make sense to have a similar one for the Explorer?

jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/devToolsTab/DevToolsToolWindow.kt
rli: no the bounding box is determined by the length of the node text

rli:
> Does it make sense to have a similar one for the Explorer?

there is already a tooltip for the explorer

manodnyab (approved):

PR #3932, opened 2023-10-19
Flip erronous state in auth support matrix

rli:
Resource Explorer does not support Builder ID

image

 
## License


manodnyab (approved):

manodnyab:
The Bulder ID text should be in disabled text formal

PR #3959, opened 2023-10-31
fix codewhispererUtil.isConnectionExpired return true when the state is "NEEDS_REFRESH"

Will-ShaoHua:
## Types of changes

token_expired

## Result
The unexpected behavior will result in"promptReauth" returning early and will not try refresh the token

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitConnectionImpls.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitConnectionImpls.kt
@@ -7,6 +7,7 @@ import com.intellij.openapi.Disposable
 import com.intellij.openapi.util.Disposer
 import software.aws.toolkits.core.TokenConnectionSettings
 import software.aws.toolkits.core.credentials.ToolkitBearerTokenProvider
+import software.aws.toolkits.jetbrains.core.credentials.sso.DiskCache
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.BearerTokenProvider
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.InteractiveBearerTokenProvider
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.ProfileSdkTokenProviderWrapper
@@ -16,6 +17,7 @@ class ManagedBearerSsoConnection(
     val startUrl: String,
     val region: String,
     override val scopes: List<String>,
+    cache: DiskCache = diskCache
 ) : BearerSsoConnection, Disposable {
     override val id: String = ToolkitBearerTokenProvider.ssoIdentifier(startUrl, region)
     override val label: String = ToolkitBearerTokenProvider.ssoDisplayName(startUrl)
@@ -25,7 +27,8 @@ class ManagedBearerSsoConnection(
             InteractiveBearerTokenProvider(
                 startUrl,
                 region,
-                scopes
+                scopes,
+                cache
             ),
             region
         )
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
@@ -13,14 +13,16 @@ import com.intellij.util.xmlb.annotations.Property
 import org.jetbrains.annotations.ApiStatus.ScheduledForRemoval
 import software.aws.toolkits.core.utils.getLogger
 import software.aws.toolkits.core.utils.warn
+import software.aws.toolkits.jetbrains.core.credentials.AwsBearerTokenConnection
+import software.aws.toolkits.jetbrains.core.credentials.ToolkitConnection
 import software.aws.toolkits.jetbrains.core.credentials.ToolkitConnectionManager
 import software.aws.toolkits.jetbrains.core.credentials.pinning.CodeWhispererConnection
 import software.aws.toolkits.jetbrains.core.credentials.sono.isSono
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.BearerTokenAuthState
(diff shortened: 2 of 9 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
```diff
@@ -196,18 +195,6 @@ object CodeWhispererUtil {
         listOf(CodeWhispererSsoLearnMoreAction(), ConnectWithAwsToContinueActionError(), DoNotShowAgainActionError())
     )
 
-    fun isAccessTokenExpired(project: Project): Boolean {
-        val tokenProvider = tokenProvider(project) ?: return false
```
Will-ShaoHua, line 200: I think `?:` case here should return `true` instead of `false`

Will-ShaoHua: previous we were not checking `isAccessTokenExpired`, which is wrong

Will-ShaoHua: that's why we've observed cases where client registration, refresh token are still valid but customers are seeing InvalidGrant

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -158,6 +159,12 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
         actionState.accountlessErrorTimestamp = state.accountlessErrorTimestamp
     }
 
+    private fun isConnectionExpired(project: Project): Boolean {
+        val tokenProvider = CodeWhispererUtil.tokenProvider(project) ?: return true
```
andrewyuq, line 163: In what condition would tokenProvider be null, again?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
andrewyuq: can we add tests here, I think this has been touched recently and it seems there's a regression.

Will-ShaoHua: yep, agree, will add them.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
Will-ShaoHua: logically, shouldn't be the case, however since function `tokenProvider` use `as?` to safely cast type therefore the null case.

jetbrains-core/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitConnectionImpls.kt
```diff
@@ -25,7 +27,8 @@ class ManagedBearerSsoConnection(
             InteractiveBearerTokenProvider(
                 startUrl,
                 region,
-                scopes
+                scopes,
+                cache
```
Will-ShaoHua, line 31: for testing purpose to inject the dependency

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -119,22 +122,27 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
...
+
+                BearerTokenAuthState.NOT_AUTHENTICATED -> {
+                    CodeWhispererLoginType.Logout
                 }
-            } else {
-                CodeWhispererLoginType.Logout
             }
-        }
-    }
+        } ?: CodeWhispererLoginType.Logout
```
Will-ShaoHua, line 145: unify this API with Toolkit `BearerTokenAuthState` interface

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -198,10 +206,17 @@ interface CodeWhispererActivationChangedListener {
...
 fun isCodeWhispererEnabled(project: Project) = with(CodeWhispererExplorerActionManager.getInstance()) {
     checkActiveCodeWhispererConnectionType(project) != CodeWhispererLoginType.Logout
 }
 
 fun isCodeWhispererExpired(project: Project) = with(CodeWhispererExplorerActionManager.getInstance()) {
-    checkActiveCodeWhispererConnectionType(project) == CodeWhispererLoginType.Expired
+    if (checkActiveCodeWhispererConnectionType(project) == CodeWhispererLoginType.Logout) {
+        return@with null
+    }
+    return@with checkActiveCodeWhispererConnectionType(project) == CodeWhispererLoginType.Expired
```
Will-ShaoHua, line 221: return `null` intentionally when the connection state is `un_authenticate`

Will-ShaoHua: as a followup, I'm thinking it would be better to remove both `isCodeWhispererEnabled` and `isCodeWhispererExpired` and only use `checkActiveCodeWhispererConnectionType`. But will open another PR for that purpose for smaller code change surface

Will-ShaoHua:
replace this one with #3961 , closing this PR

PR #3961, opened 2023-11-01
fix codewhisperer auth

Will-ShaoHua:
## Demo of the issue
As the attached video, I modified credential access token `expiredAt` field to be time point in the past and put break points at function `promptReauth` and make sdk `refreshToken` doing nothing. The `promptReauth` won't be really executed because `isCodeWhispererExpired` will return false anyway as long as the `refreshToken != null`. 

TL;DR: the fallback mechanism of manual refresh is broken and we fully rely on SDK refresher

https://github.com/aws/aws-toolkit-jetbrains/assets/96078566/cb97bb01-111f-432a-82b1-877c392a301d

token_expired

## Reason
There're cases where aws java sdk has not refreshed the token and leave the token in `NEEDS_REFRESH` state. 
And changes from #3780 change the definition of `isCodewhispererExpired` by only checking 
```
fun isRefreshTokenExpired(project: Project): Boolean {
        val state = tokenProvider.state()
        return state == BearerTokenAuthState.NOT_AUTHENTICATED
    }
```
which equivalently is saying if the refresh token is non null, it's not expired, and we will not manually refresh the token when it's required, leaving the UI stuck in a state where all CW UI suggest CW is working however the token is expired.

## Types of changes

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
@@ -110,10 +110,10 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
     @Deprecated("Accountless credential will be removed soon")
     @ScheduledForRemoval
     // Will keep it for existing accountless users
-        /**
-         * Will be called from CodeWhispererService.showRecommendationInPopup()
-         * Caller (e.x. CodeWhispererService) should take care if null value returned, popup a notification/hint window or dialog etc.
-         */
+    /**
+     * Will be called from CodeWhispererService.showRecommendationInPopup()
+     * Caller (e.x. CodeWhispererService) should take care if null value returned, popup a notification/hint window or dialog etc.
+     */
     fun resolveAccessToken(): String? {
         if (actionState.token == null) {
             LOG.warn { "Logical Error: Try to get access token before token initialization" }
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
@@ -110,10 +110,10 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
     @Deprecated("Accountless credential will be removed soon")
     @ScheduledForRemoval
     // Will keep it for existing accountless users
-    /**
-     * Will be called from CodeWhispererService.showRecommendationInPopup()
-     * Caller (e.x. CodeWhispererService) should take care if null value returned, popup a notification/hint window or dialog etc.
-     */
+        /**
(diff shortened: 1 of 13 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -201,7 +207,3 @@ interface CodeWhispererActivationChangedListener {
 fun isCodeWhispererEnabled(project: Project) = with(CodeWhispererExplorerActionManager.getInstance()) {
     checkActiveCodeWhispererConnectionType(project) != CodeWhispererLoginType.Logout
 }
-
-fun isCodeWhispererExpired(project: Project) = with(CodeWhispererExplorerActionManager.getInstance()) {
-    checkActiveCodeWhispererConnectionType(project) == CodeWhispererLoginType.Expired
-}
```
Will-ShaoHua, line 207: remove since it has nuance that if "logout" is expired or not. use `checkActiveCodeWhispererConnectionType` instead and check if the type is `Expired`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -119,21 +121,25 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
...
                 }
-            } else {
-                CodeWhispererLoginType.Logout
+
+                BearerTokenAuthState.NEEDS_REFRESH -> CodeWhispererLoginType.Expired
+
+                BearerTokenAuthState.NOT_AUTHENTICATED -> CodeWhispererLoginType.Logout
             }
-        }
+        } ?: CodeWhispererLoginType.Logout
```
Will-ShaoHua, line 142: unify this API with Toolkit auth API

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
andrewyuq: instead of removing it, can we just add a comment here to explain the nuance? I see duplicated code otherwise. Or rename it to something like "isCodeWhispererRefreshNeeded"

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererUtil.kt
```diff
@@ -196,18 +195,6 @@ object CodeWhispererUtil {
...
-        val state = tokenProvider.state()
-        return state == BearerTokenAuthState.NEEDS_REFRESH
-    }
-
-    fun isRefreshTokenExpired(project: Project): Boolean {
-        val tokenProvider = tokenProvider(project) ?: return false
-        val state = tokenProvider.state()
-        return state == BearerTokenAuthState.NOT_AUTHENTICATED
-    }
-
```
Will-ShaoHua, line 210: removed since it's not being used elsewhere, move the logic into CodewhispererExplorerActionManager

andrewyuq (approved):

rli (approved):

PR #3974, opened 2023-11-07
Don't allow setting up authentication for CW in Dev Env

manodnyab:
Disable authentication if user is trying to setup auth in dev env
 
## License


resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
```diff
@@ -964,6 +964,7 @@ gettingstarted.builderid.description=Personal profile for builders
 gettingstarted.codecatalyst.open.explorer=Open CodeCatalyst menu
 gettingstarted.codecatalyst.panel.create.space=Create a CodeCatalyst space
 gettingstarted.codecatalyst.panel.setup=Set up CodeCatalyst
+gettingstarted.codewhisperer.remote=Codewhisperer is unavailable in JetBrains Gateway
```
rli, line 967:
```suggestion
gettingstarted.codewhisperer.remote=CodeWhisperer is unavailable in JetBrains Gateway
```

jetbrains-core/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
```diff
@@ -82,6 +83,10 @@ fun requestCredentialsForCodeWhisperer(
     isFirstInstance: Boolean = false,
     connectionInitiatedFromExplorer: Boolean = false
 ): Boolean {
+    if (isRunningOnRemoteBackend()) {
```
rli, line 86: ideally we disable the buttons too

rli (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
manodnyab: We won't be able to disable the link for logging in with Idc, which is why added the mesage here

PR #3978, opened 2023-11-08
CodeWhisperer: Add feature fetching component

andrewyuq:
1. Add a component to fetch feature assignments every 30 mins, calling ListFeatureEvaluations API and cache values in memory.
2. Add a field in UserTriggerDecision codewhispererFeatureEvaluations to record all the feature configs in a string, which can be queried on Kibana

## Types of changes

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererFeatureConfigService.kt
```diff
@@ -0,0 +1,80 @@
...
+import software.aws.toolkits.core.utils.getLogger
+import software.aws.toolkits.jetbrains.services.codewhisperer.credentials.CodeWhispererClientAdaptor
+import software.aws.toolkits.jetbrains.services.codewhisperer.explorer.isCodeWhispererExpired
+
+class CodeWhispererFeatureConfigService {
+    private val featureConfigs = mutableMapOf<String, FeatureContext>()
+
+    @RequiresBackgroundThread
+    fun fetchFeatureConfigs(project: Project) {
+        if (isCodeWhispererExpired(project)) return
```
Will-ShaoHua, line 20:
```suggestion
        if (!isCodeWhispererEnabled(project) || isCodeWhispererExpired(project)) return
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/startup/CodeWhispererProjectStartupActivity.kt
```diff
@@ -113,6 +123,16 @@ class CodeWhispererProjectStartupActivity : StartupActivity.DumbAware {
             parsedLastShown.plusDays(7) <= LocalDateTime.now()
         } ?: true
     }
+
+    // Start a job that runs every 30 mins
+    private fun initFeatureConfigPollingJob(project: Project) {
+        projectCoroutineScope(project).launch {
+            while (isActive) {
+                CodeWhispererFeatureConfigService.getInstance().fetchFeatureConfigs(project)
```
Will-ShaoHua, line 131: given it's a app level config, shouldn't we put this part of logic in a app level component instead?
think this way the network overhead could be slightly reduced

Will-ShaoHua: nm, either way we will still need "project" to check connection validity. But we need some debounce logic maybe.

PR #3979, opened 2023-11-08
Updated metric name for first startup

manodnyab:
## Description

Metric name is capitalized in one location. It needs to be lowercase

 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedOnStartup.kt
```diff
@@ -38,15 +38,15 @@ class GettingStartedOnStartup : StartupActivity {
                 GettingStartedPanel.openPanel(project, firstInstance = true, connectionInitiatedFromExplorer = false)
                 AuthTelemetry.addConnection(
                     project,
-                    source = SourceOfEntry.FIRST_STARTUP.name,
+                    source = SourceOfEntry.FIRST_STARTUP.toString(),
```
rli, line 41: i'm starting to think we need to centralize the enum -> toString() logic somewhere

rli (approved):

PR #3981, opened 2023-11-09
Add deprecation messages for IDE 223 / Gateway 232

rli:
This is earlier than we would normally do this, and we likely will not do this immediately. However, we want the agility in case we need to drop these builds last minute.

## License


G-Rath:
hey so I'm now getting a deprecation warning about being on 2023.2.5 even though there is no higher non-EAP version to upgrade to - I'm guessing that's related to this.

It seems very premature to start telling people they need to upgrade when there isn't a version available to upgrade to?

PR #3997, opened 2023-11-23
feat(codecatalyst): Use MDE endpoint set by environment variable

rli:
MDE requires use of different endpoint in some VPC configurations.

## Types of changes

## Description
Impl copied from https://github.com/aws/aws-toolkit-vscode/pull/4077

## Checklist
 
## License


jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/CawsEnvironmentClient.kt
```diff
@@ -30,7 +30,7 @@ import software.aws.toolkits.jetbrains.utils.notifyError
 import software.aws.toolkits.resources.message
 
 class CawsEnvironmentClient(
-    private val endpoint: String = "http://127.0.0.1:1339",
+    private val endpoint: String = System.getenv(CawsConstants.CAWS_ENV_AUTH_TOKEN_VAR) ?: CawsConstants.DEFAULT_CAWS_ENV_API_ENDPOINT,
```
manodnyab, line 33: I think the env var will have the default endpoint too, so will there be a situation that this is null?

manodnyab (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/caws/envclient/CawsEnvironmentClient.kt
rli: i believe they said this will only be set if there's an override?

PR #4028, opened 2023-12-01
WIP tests

fjnoyp:
General unit tests for CWC chat.  Covers most of the major functionality with the remainign minor classes to be covered in a separate PR.


jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
```diff
@@ -0,0 +1,132 @@
...
+     */
+
+    private lateinit var fileContextExtractor: FileContextExtractor
+
+    @Before
+    fun setUp() {
+        fileContextExtractor = FileContextExtractor(mockFqnWebviewAdapter, mockProject, mockLanguageExtractor)
+
+        // Editor
+        mockkStatic(FileEditorManager::class)
```
mattcreaser, line 51: You need to match every call to `mockkStatic`, `mockkObject`, and `mockkConstructor` with a matching call to `unmockkStatic`/`unmockkObject`/`unmockkConstructor`, or simply do an `unmockkAll`. This is because these are static mocks, not just mock instances, and can leak across tests.

This is why I prefer using the `mockkStatic(class) { .... }` form of these functions where possible, which automatically unmocks at the end of the lambda. Since you're doing it in the `@Before` here though you need to unmock in the `@After`

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
```diff
@@ -0,0 +1,132 @@
...
+        coEvery { MatchPolicyExtractor.extractMatchPolicyFromCurrentFile(any(), any(), any(), any()) } returns null
+
+        // Act
+        val result = runBlocking {
+            fileContextExtractor.extract()
+        }
+
+        // Assert
+        assertNotNull(result)
+        assertEquals(testFileLanguage, result?.fileLanguage)
```
mattcreaser, line 111:
```suggestion
        assertThat(result?.fileLanguage).isEqualTo(testFileLanguage)
```
Replace instances of junit assertions

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
```diff
@@ -0,0 +1,132 @@
...
+        val testFileText = "public class Test {}"
+        every { mockEditor.document.text } returns testFileText
+
+        mockkStatic(PsiDocumentManager::class)
+        val mockPsiFile = mockk<PsiFile>(relaxed = true)
+        every { PsiDocumentManager.getInstance(any()).getPsiFile(any()) } returns mockPsiFile
+        val testFilePath = "/path/to/file"
+        every { mockPsiFile.virtualFile.path } returns testFilePath
+
+        mockkObject(MatchPolicyExtractor)
```
mattcreaser, line 101: Needs to be unmocked

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
```diff
@@ -0,0 +1,132 @@
...
+        coEvery { MatchPolicyExtractor.extractMatchPolicyFromCurrentFile(any(), any(), any(), any()) } returns null
+
+        // Act
+        val result = runBlocking {
+            fileContextExtractor.extract()
+        }
+
+        // Assert
+        assertNotNull(result)
```
mattcreaser, line 110: This assertion is kind of redundant with the others

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
```diff
@@ -0,0 +1,794 @@
...
+        every { BrowserUtil.browse(any<String>()) } just Runs
+
+        // Act
+        runBlocking {
+            chatController.processLinkClick(testMessage)
+        }
+
+        // Assert
+        verify { BrowserUtil.browse(testUrl) }
+        coVerify { telemetryHelper.recordInteractWithMessage(match { it == testMessage }) }
```
mattcreaser, line 789:
```suggestion
        coVerify { telemetryHelper.recordInteractWithMessage(testMessage) }
```

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
```diff
@@ -0,0 +1,794 @@
...
+        // Act
+        runBlocking {
+            chatController.processContextMenuCommand(testMessage, testTriggerId)
+        }
+
+        // Assert
+
+        val formattedCodeSelection = "\n```\n${testCodeSelection}\n```\n"
+
+        coEvery {
```
mattcreaser, line 742:
```suggestion
        coVerify {
```

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
```diff
@@ -0,0 +1,794 @@
...
+                        it.message == formattedCodeSelection &&
+                        it.command == testMessage.command.actionId &&
+                        it.triggerId == testTriggerId
+                }
+            )
+        }
+
+        val prompt = "${testMessage.command} the following part of my code for me: $formattedCodeSelection"
+
+        coEvery {
```
mattcreaser, line 755:
```suggestion
        coVerify {
```

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
```diff
@@ -0,0 +1,794 @@
...
+            // Act
+            chatController.processOnboardingPageInteraction(testMessage, testTriggerId)
+        }
+
+        // Assert
+
+        // Check call to sendOnboardingPageInteractionMessage
+        coVerify {
+            messagePublisher.publish(
+                match {
```
mattcreaser, line 679: I would think about extracting some of these matchers into custom functions to make the tests more readable. Something like the following (not tested, just an outline):

```kotlin
fun MockKMatcherScope.matchOnboardingInteractionMessage(
  triggerId: String,
  message: StaticPrompt = StaticPrompt.OnboardingHelp,
  interactionType: OnboardingInteractionType = OnboardingInteractionType.CwcButtonClick
): AmazonQMessage = match {
  it is OnboardingPageInteractionMessage && ...
}

// usage
coVerify { messagePublisher.publish(matchOnboardingPageInteractionMessage(testTriggerId)) }
```

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
```diff
@@ -0,0 +1,794 @@
...
+        runBlocking {
+            chatController.processCodeWasCopiedToClipboard(testMessage)
+        }
+
+        // Assert
+        coVerify { telemetryHelper.recordInteractWithMessage(match { it == testMessage }) }
+    }
+
+    @Test
+    fun `processInsertCodeAtCursorPosition handles code insertion`() {
```
mattcreaser, line 450: This tests is excessively long, with a lot of mocking. That is a pretty good indicator that the underlying code should be refactored.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/auth/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    @After
+    fun tearDown() {
+        unmockkAll()
+    }
+
+    @Test
+    fun `NotConnected returns AuthNeededState`() {
+        // Arrange
+        every { checkBearerConnectionValidity(any(), BearerTokenFeatureSet.Q) } returns ActiveConnection.NotConnected
```
mattcreaser, line 44: You can make a helper function to avoid repeating all of these calls.

```kotlin
private fun mockConnectionValidity(
  qConnection: ActiveConnection = ActiveConnection.NotConnected,
  cwConnection: ActiveConnection = ActiveConnection.NotConnected
) {
    every { checkBearerConnectionValidity(any(), BearerTokenFeatureSet.Q) } returns qConnection
    every { checkBearerConnectionValidity(any(), BearerTokenFeatureSet.CODEWHISPERER) } returns cwConnection
}
```

mattcreaser (changes requested):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cwc/controller/ChatController.kt
```diff
@@ -477,13 +471,27 @@ class ChatController private constructor(
...
+        internal const val NO_TAB_AVAILABLE = "no-available-tabs"
 
         val objectMapper: ObjectMapper = jacksonObjectMapper()
             .registerModule(JavaTimeModule())
             .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
             .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
             .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
             .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+
+        fun create(context: AmazonQAppInitContext): ChatController {
```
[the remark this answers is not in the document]
fjnoyp: Unfortunately our original implementation had the telemetry helper created internally making it impossible to mock.  To respect the auto telemetryHelper creation I created this create method that is called while the internal constructor is called from our test code.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/auth/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        val authController = AuthController()
+        val authNeededState = authController.getAuthNeededState(mockProject)
+
+        // Assert
+        assertThat(authNeededState).isNotNull
+        assertThat(authNeededState?.authType).isEqualTo(AuthFollowUpType.FullAuth)
+    }
+
+    /*
+    // Broken due to Kotlin glitch preventing mocking of outside method "GettingStartedAuthUtilsKt"
```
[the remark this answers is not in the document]
fjnoyp: I am unable to mock this method that is called with default credentials.  Unsure if it's related to that or because it's in a different project scope.  Regardless this causes the 'real' method to continue being called, thereby breaking the unit test. 

`// Broken here: cannot mock outside method
        mockkStatic("software.aws.toolkits.jetbrains.core.gettingstarted.GettingStartedAuthUtilsKt")
        every {
            requestCredentialsForQ(
                any(), // project
                any(), // initialConnectionCount, using any() since we can't replicate default logic
                any(), // initialAuthConnections, using any() for the same reason
                any(), // isFirstInstance
                any(), // connectionInitiatedFromExplorer
                eq(true) // connectionInitiatedFromQChatPanel, with specific value
            )
        } returns true`

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
fjnoyp: Wouldn't I just need to un mock static?  A mockObject would never leak between tests right - unless it's explicitly stored in an instance variable?

I called unmockkAll

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
fjnoyp: With this call this is implicitly umocked now right?

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/editor/context/file/util/FileContextExtractorTest.kt
```diff
@@ -0,0 +1,132 @@
...
+            firstArg<() -> Unit>().invoke()
+        }
+        every { EdtUtility.runReadAction<String> (any()) } answers {
+            firstArg<() -> String>().invoke()
+        }
+    }
+
+    @After
+    fun tearDown() {
+        clearMocks(
```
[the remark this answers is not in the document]
fjnoyp: Switched to unmockkAll instead to handle mock object and static.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
fjnoyp: How would you propose refactoring the underlying code though?  The actual `processInsertCodeAtCursorPosition` method is about 30 lines of code, but its reliance on various editor/static classes which thereby increase the complexity.  

The scope of the underlying method itself also doesn't seem too big - it's main functionality is to insert code at cursor position and not much else. 

If I were to break that underlying method into several methods, those methods should be private (so unsure how to mock).  If it is possible to mock them and then to write separate tests for them that could reduce the scope here, but unsure if that's even possible.

ppiliaev (approved):

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/auth/AuthControllerTest.kt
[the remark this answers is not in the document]
rli:
```kt
mockkStatic(::requestCredentialsForQ)
```

is the invocation you want

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/auth/AuthControllerTest.kt
```diff
@@ -0,0 +1,195 @@
...
+import org.junit.Test
+import software.aws.toolkits.jetbrains.core.gettingstarted.editor.ActiveConnection
+import software.aws.toolkits.jetbrains.core.gettingstarted.editor.BearerTokenFeatureSet
+import software.aws.toolkits.jetbrains.core.gettingstarted.editor.checkBearerConnectionValidity
+import software.aws.toolkits.jetbrains.core.gettingstarted.reauthenticateWithQ
+import software.aws.toolkits.jetbrains.services.cwc.controller.chat.telemetry.TelemetryHelper
+import software.aws.toolkits.jetbrains.services.cwc.utility.EdtUtility
+import software.aws.toolkits.telemetry.CwsprChatCommandType
+
+class AuthControllerTest {
```
rli, line 27: should really be doing this in JUnit5, but no point in rewriting them now

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/auth/AuthControllerTest.kt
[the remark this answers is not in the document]
fjnoyp: Unfortunately this doesn't work - are we sure it's even possible to mock these standalone static methods in mockito?  They do not have an associated class which makes this impossible.  From my understanding having standalone methods is not ideal either - I feel like we need to have a separate PR to refactor the `GettingStartedAuthUtils` to be a proper static class before being able to mock it in this test file.

jetbrains-core/tst/software/aws/toolkits/jetbrains/services/cwc/controller/ChatControllerTest.kt
fjnoyp: Hmm unsure how this would reduce code or make things more readable?  

We have 8 coVerify tests on messagePublisher with around 4 different message types being matched on which each have different data schemas.  

Furthermore, we compare against actual test values in the code, if we were to consolidate it would still require creating a 'dummy' test message that we'd have to create and then match against in this helper method right?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/cwc/utility/EdtUtility.kt
```diff
@@ -0,0 +1,12 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cwc.utility
+
+object EdtUtility {
```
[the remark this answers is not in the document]
fjnoyp: Is there another way to do this then?  I created this to allow mocking of the runInEdt and runReadAction - since these methods come from openapi libs I was unable to mock them in my tests.  Without being able to mock the tests would fail due to improper environment.

PR #4048, opened 2023-12-08
fix(cwc): update telemetry

div5yesh:
- update telemetry as per new models
- send time between chunks
 
## License


leigaol:
Can you fix the required Github CI? Thanks!

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/credentials/CodeWhispererClientAdaptor.kt
```diff
@@ -334,6 +370,81 @@ open class CodeWhispererClientAdaptorImpl(override val project: Project) : CodeW
...
+        userIntent: UserIntent?,
+        hasCodeSnippet: Boolean?,
+        programmingLanguage: String?,
+        activeEditorTotalCharacters: Int?,
+        timeToFirstChunkMilliseconds: Double?,
+        timeBetweenChunks: List<Double>?,
+        fullResponselatency: Double?,
+        requestLength: Int?,
+        responseLength: Int?
+    ): SendTelemetryEventResponse = bearerClient().sendTelemetryEvent { requestBuilder ->
```
rli, line 385: chat should avoid `CodeWhispererClientAdaptor` since it's using the connection associated with suggestions

div5yesh: What is the alternative?

rli: grab a client instance with QConnection
```kt
val connection = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance())
val client = AwsClientManager.getInstance().getClient<CodeWhispererRuntimeClient>(connection.getConnectionSettings())
```

saggarn: Why do we need a separate client here? This API is common for all clients and will be used for recording events for code completions, chat, WB and Gumby. For reference, we also reuse the CW client in VS code https://github.com/aws/aws-toolkit-vscode/pull/4200/files

PR #4053, opened 2023-12-12
Improve pom detection algorithm

arangatang:
## Types of changes

## Description
Using `VfsUtil.collectChildrenRecursively` can be prohibitively slow for very large projects making the validation take a very long time. This pr contains these changes to speed this up:

1. Uses kotlins walkTopDown which allows for early stopping of recursion when `pom.xml` is found.
2. Limits the max depth we search for `pom.xml` to some limit (currently 5 dirs deep, previously unlimited)
3. Avoids running the search for `pom.xml` twice as was done before. 

These changes makes a ~200 MB with depth of 12 take <1 second to scan instead of ~10 before.

### Before

https://github.com/aws/aws-toolkit-jetbrains/assets/9272897/da6bb10c-660e-4d70-989b-db6bae91c6ad

### After

https://github.com/aws/aws-toolkit-jetbrains/assets/9272897/7b86eb12-aff3-4e22-a659-46b1f4b21cca

## Checklist
- [X] I have added tests to cover my changes
 
## License


dhasani23 (commented):
Just double check if we want to search for pom.xml in root dir only

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -638,6 +639,29 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
     }
 
+    private fun findBuildFiles(sourceFolder: File): List<File> {
+        /**
+         * For every dir, check if any supported build files (pom.xml etc) exists.
+         * If found store it and stop further recursion.
+         */
+        val buildFiles = mutableListOf<File>()
+        sourceFolder.walkTopDown()
+            .maxDepth(5)
```
rli, line 649: is this a good value for the depth parameter?

rli (approved):

arangatang:
@dhasani23 
We should be searching deeper than the root only, but the uploaded zip should have a pom in the root directory.
Further if the poms are nested then the top level pom should be the only option for customers.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
damntrecky: I'm not a good source for Java/Kotlin so I don't know what is a reasonable depth for these types of projects. Where would be a good source for this?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -638,6 +639,29 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
         managerState.flags.putAll(state.flags)
     }
 
+    private fun findBuildFiles(sourceFolder: File): List<File> {
```
damntrecky, line 642: Question: Do we have a test to make sure this doesn't find the file after N depth?

Would this be too much effort to create and not a lot of value proposition in that test?

damntrecky (approved):
Approved with a follow up on the maxDepth and comment explaining why we chose the depth.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
arangatang: Not much value imo, more importantly is that we have tests for nested and poms not in the root which we both have.

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
arangatang: This is just a best guess atm, we do not have statistics for this atm.

Will-ShaoHua (approved):

PR #4074, opened 2023-12-20
Fix: add supportPatching flag for code transform

tincheng:
## Types of changes

## Description

Default `supportPatching` to true for code transform

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/ZipManifest.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/ZipManifest.kt
@@ -9,5 +9,6 @@ import software.aws.toolkits.jetbrains.services.codemodernizer.ZIP_SOURCES_PATH
 data class ZipManifest(
     val sourcesRoot: String = ZIP_SOURCES_PATH,
     val dependenciesRoot: String? = null,
-    val version: String = UPLOAD_ZIP_MANIFEST_VERSION.toString()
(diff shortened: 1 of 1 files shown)


jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/ZipManifest.kt
```diff
@@ -9,5 +9,6 @@ import software.aws.toolkits.jetbrains.services.codemodernizer.ZIP_SOURCES_PATH
 data class ZipManifest(
     val sourcesRoot: String = ZIP_SOURCES_PATH,
     val dependenciesRoot: String? = null,
-    val version: String = UPLOAD_ZIP_MANIFEST_VERSION.toString()
+    val version: String = UPLOAD_ZIP_MANIFEST_VERSION.toString(),
+    val supportPatching: Boolean = true
```
damntrecky, line 13: This is an API level change to fix a customer facing issue. Will require a release outside the normal release cycle.

damntrecky: Did we verify this fix has worked E2E from backend to IDE?

The expected behavior should from adding "supportPatching" should be to not add source code to the zip file on return, reducing download size. This change is specific to IntelliJ because it supports the "patch file" diff viewer and source code is not needed.

damntrecky (approved):
Please verify E2E test/change has worked.

If we can't release today, we can focus on writing a test case for this.

tincheng:
Canceled per team decision. Issue will be handled on the service side.

PR #4079, opened 2023-12-21
Add support for opening CodeCatalyst dev env using Idc

juanbenj:
## License


jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
```diff
@@ -98,8 +103,11 @@ class CawsConnectionProvider : GatewayConnectionProvider {
             return null
         }
 
+        val ssoSettings = connectionParams.ssoSettings ?: SsoSettings(SONO_URL, SONO_REGION)
+
         val connectionSettings = try {
-            CodeCatalystCredentialManager.getInstance().getSettingsAndPromptAuth()
+            loginSso(null, ssoSettings.startUrl, ssoSettings.region, CODECATALYST_SCOPES)
```
rli, line 109: does this do the right thing if you're currently logged into builder id, but the requested environment lies in an SSO tenant?

jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
```diff
@@ -98,8 +103,11 @@ class CawsConnectionProvider : GatewayConnectionProvider {
             return null
         }
 
+        val ssoSettings = connectionParams.ssoSettings ?: SsoSettings(SONO_URL, SONO_REGION)
```
rli, line 106: if we're defaulting to builder id, then https://github.com/aws/aws-toolkit-jetbrains/blob/4f5f56d08077653ce216b829d5354a86ad12d1f7/jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/welcomescreen/WorkspaceDetails.kt#L140-L149 needs to be amended to include the sso url / region when connecting from the gateway UI

jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
juanbenj: Yes. loginSso handles it by default. It will sign you out of the BuilderID connection and sign you into Idc based on the parameters.

jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/welcomescreen/ExistingWorkspaces.kt
```diff
@@ -54,13 +57,20 @@ class ExistingWorkspaces(
                 }
         }
 
+        val ssoSettings: SsoSettings? = ToolkitAuthManager.getInstance().getConnection(connectionSettings.providerId)?.let {
+            if (it is AwsBearerTokenConnection) {
+                return@let SsoSettings(it.startUrl, it.region)
+            }
+            return@let null
```
rli, line 64: nit:
```suggestion
            if (it is AwsBearerTokenConnection) {
                SsoSettings(it.startUrl, it.region)
            } else {
                null
            }
```

rli, line 64: maybe a LOG.warn when this is null

rli (approved):

manodnyab (approved):

PR #4087, opened 2024-01-03
Add JobId to Code Transform History

tincheng:
## Types of changes

## Description

Add Job Id to Code Transform Job History Panel
Screenshot 2024-01-03 at 2 48 13 PM

## Checklist
 
## License

### DIFF
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerSession.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerSession.kt
@@ -116,7 +116,7 @@ class CodeModernizerSession(
                 return CodeModernizerStartJobResult.Cancelled
             }
             val startJobResponse = startJob(uploadId)
-            state.putJobHistory(sessionContext, "STARTED", startJobResponse.transformationJobId())
+            state.putJobHistory(sessionContext, TransformationStatus.STARTED, startJobResponse.transformationJobId())
             state.currentJobStatus = TransformationStatus.STARTED
             CodeModernizerStartJobResult.Started(JobId(startJobResponse.transformationJobId()))
         } catch (e: AlreadyDisposedException) {
@@ -124,7 +124,7 @@ class CodeModernizerSession(
             return CodeModernizerStartJobResult.Disposed
         } catch (e: Exception) {
             val errorMessage = "Failed to start job"
-            state.putJobHistory(sessionContext, "FAILED TO START")
+            state.putJobHistory(sessionContext, TransformationStatus.FAILED)
             state.currentJobStatus = TransformationStatus.FAILED
             CodetransformTelemetry.logGeneralError(
                 codeTransformApiErrorMessage = errorMessage,
@@ -161,7 +161,7 @@ class CodeModernizerSession(
             state.currentJobStatus = new
             sessionContext.project.refreshCwQTree()
             val instant = Instant.now()
-            state.updateJobHistory(sessionContext, new.name, instant)
+            state.updateJobHistory(sessionContext, new, instant)
             setCurrentJobStopTime(new, instant)
             setCurrentJobSummary(new)
             jobTransitionHandler(new, plan)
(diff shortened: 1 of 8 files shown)


dhasani23 (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/state/CodeModernizerSessionState.kt
```diff
@@ -29,23 +29,24 @@ class CodeModernizerSessionState {
     var transformationSummary: TransformationSummary? = null
     var currentJobId: JobId? = null
 
-    private fun getJobItemId(sessionContext: CodeModernizerSessionContext) = Path(sessionContext.configurationFile.path).toAbsolutePath().toString()
-    fun putJobHistory(sessionContext: CodeModernizerSessionContext, status: String, startedAt: Instant = Instant.now()) {
-        val id = getJobItemId(sessionContext)
+    private fun getJobModuleName(sessionContext: CodeModernizerSessionContext) = Path(sessionContext.configurationFile.path).toAbsolutePath().toString()
+    fun putJobHistory(sessionContext: CodeModernizerSessionContext, status: String, jobId: String = "", startedAt: Instant = Instant.now()) {
+        val module = getJobModuleName(sessionContext)
```
damntrecky, line 34: nit: `module` should be called something like `moduleName` since it reads like its an object. But its a string.

resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
```diff
@@ -590,6 +590,7 @@ codemodernizer.toolwindow.scan_in_progress.stopping=Stopping the job...
 codemodernizer.toolwindow.scan_in_progress.transforming=Amazon Q is transforming your code. Details will appear soon.
 codemodernizer.toolwindow.stop_scan=Stop job
 codemodernizer.toolwindow.table.header.date=Date
+codemodernizer.toolwindow.table.header.job_id=Job Id
```
damntrecky, line 593: Question: Is there supposed to be a space here or should it be `JobId`? I think there are arguments for both Job Id and JobId, don't want to over complicate something simple.

damntrecky (approved):

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerSession.kt
```diff
@@ -250,7 +250,7 @@ class CodeModernizerSession(
     /**
      * This will resume the job, i.e. it will resume the main job loop kicked of by [createModernizationJob]
      */
-    fun resumeJob(startTime: Instant) = state.putJobHistory(sessionContext, "Started", startTime)
+    fun resumeJob(startTime: Instant, jobId: JobId) = state.putJobHistory(sessionContext, "Started", jobId.id, startTime)
```
manodnyab, line 253: The status in the putJobHistory function is in upperCase [here](https://github.com/aws/aws-toolkit-jetbrains/pull/4087/files#diff-8b0ea0f291f1c20406e922c82de118911f271b383adaa9520b9b3bf521abaf43R119). Should this be changed to ensure parity? Is it a good idea to make this an enum instead of a string?

resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
tincheng: @damntrecky I used "Job Id" because there are spaces in "Module Name" and "Job running time". On that note, VSCode uses different display names: "Id", "Module", and "Duration". While the UIs do not need to look identical across IDEs, this is a straightforward change, and I personally prefer the titles in IntelliJ more as they give more context. Thoughts on changing VSCode to match?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerSession.kt
tincheng: great find, will fix this

manodnyab (approved):

PR #4109, opened 2024-01-18
feature(QTransform): Detect and add telemetry for 1p dependencies

carolabadeer:
## Types of changes

## Description
This PR adds detection for 1p dependencies (private dependencies not available on maven central) and creates new telemetry to help the Q Transform team get more granulated information on the types of dependencies in applications submitted through the transformation. 

This change includes: 
- Running a new maven command that generates a report on dependencies that have newer versions available in maven central. This report is parsed for dependencies that are reported to be on the latest version (which can be available on maven but being used in the application at the latest version available, or be a private dependency). The small subset of dependencies parsed from this report are then checked against maven central. 
    - A few notes on latency: 
        - This approach adds about 3-5 seconds of latency for large multi-module applications tested. The `-DonlyProjectDependencies=true` flag reduced this number from ~1 minute and 40 seconds to only a few seconds. 
        - This logic runs in a new coroutine to ensure that the overall transformation time is not impacted on the customer's end. 
- This change also refactors the `runMavenCommand` function, which was strictly running the `mvn copy-dependencies` command, to generalize it and allow for passing in any `commandList`

## Checklist
- [ ] I have added tests to cover my changes (tests conducted for this change consisted of end-to-end transformations on the IntelliJ IDE, and tested telemetry generation locally)
 
## License

### DIFF
--- a/jetbrains-core/build.gradle.kts
+++ b/jetbrains-core/build.gradle.kts
@@ -139,9 +139,11 @@ dependencies {
 
     implementation(project(":mynah-ui"))
     implementation(libs.aws.crt)
+    implementation(libs.aws.crt)
     implementation(libs.bundles.jackson)
     implementation(libs.zjsonpatch)
     implementation(libs.commonmark)
+    implementation(libs.jsoup)
 
     testImplementation(project(path = ":core", configuration = "testArtifacts"))
     testImplementation(libs.mockk)
--- a/jetbrains-core/build.gradle.kts
+++ b/jetbrains-core/build.gradle.kts
@@ -139,7 +139,6 @@ dependencies {
 
     implementation(project(":mynah-ui"))
     implementation(libs.aws.crt)
-    implementation(libs.aws.crt)
     implementation(libs.bundles.jackson)
     implementation(libs.zjsonpatch)
     implementation(libs.commonmark)
--- a/jetbrains-core/resources/telemetryOverride.json
+++ b/jetbrains-core/resources/telemetryOverride.json
@@ -207,6 +207,11 @@
                 "q",
                 "codewhispererQ"
             ]
+        },
+        {
+            "name": "has1pdependency",
+            "type": "boolean",
+            "description": "Project contains at least 1 dependency not available in maven central"
         }
     ],
     "metrics": [
@@ -505,6 +510,24 @@
                 {
                     "type": "cwsprChatCommandName",
                     "required": false
+                }
+            ]
+        },
+        {
+            "name": "codeTransform_projectContains1pDependencies",
+            "description": "When a project contains dependencies not available on maven central",
+            "metadata": [
+                {
+                    "type": "codeTransformSessionId",
+                    "required": true
+                },
+                {
+                    "type": "codeTransformJobId",
+                    "required": true
+                },
+                {
+                    "type": "has1pdependency",
+                    "required": true
                 }
             ]
         }
--- a/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
+++ b/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
@@ -74,14 +74,9 @@ import java.time.Instant
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.swing.Icon
 import org.jsoup.Jsoup;
-import org.jsoup.nodes.Element;
-import org.jsoup.select.Elements;
-import java.net.URI;
-import java.net.http.HttpClient;
-import java.net.http.HttpRequest;
-import java.net.http.HttpResponse;
 import kotlinx.serialization.json.Json;
 import kotlinx.serialization.json.JsonObject;
+import software.aws.toolkits.jetbrains.core.getTextFromUrl;
 
 @State(name = "codemodernizerStates", storages = [Storage("aws.xml", roamingType = RoamingType.PER_OS)])
 class CodeModernizerManager(private val project: Project) : PersistentStateComponent<CodeModernizerState>, Disposable {
@@ -387,19 +382,10 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
     dependencies
      */
     private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
-        val httpClient = HttpClient.newHttpClient();
-        val request =  HttpRequest.newBuilder()
-            .uri(URI.create("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json"))
-            .GET()
-            .build();
-
         try{
-            val response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
-
-            if(response.statusCode() == 200){
-                val jsonResponse = response.body();
+            val response = getTextFromUrl("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json")
                 try {
-                    val json: JsonObject = Json.parseToJsonElement(jsonResponse) as JsonObject;
+                    val json: JsonObject = Json.parseToJsonElement(response) as JsonObject;
                     val responses = json["response"] as? JsonObject;
                     val numFound = responses?.get("numFound");
 
@@ -415,9 +401,6 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
                 } catch (e: Exception){
                     LOG.warn{"Error parsing JSON response from GET request to maven central: ${e.printStackTrace()}"}
                 }
-            } else {
-                LOG.warn{"Error making request to maven central. Response code: ${response.statusCode()}"}
-            }
         } catch(e: Exception){
             LOG.warn{"Exception while making GET request to maven central: ${e.printStackTrace()}"}
         }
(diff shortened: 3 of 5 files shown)


jetbrains-core/build.gradle.kts
```diff
@@ -139,9 +139,11 @@ dependencies {
 
     implementation(project(":mynah-ui"))
     implementation(libs.aws.crt)
+    implementation(libs.aws.crt)
```
rli, line 142: ?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +381,113 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+        }
+        return false;
+    }
+
+
+    /**
+     * Coroutine to find and generate metrics on 1p dependencies (dependencies
+     * not available in maven central)
+     */
+    fun generate1pDependencyMetrics(jobId: JobId, session: CodeModernizerSession) = projectCoroutineScope(project).launch {
```
rli, line 432: there should definitely be a way to get this as structured output instead of scraping HTML, which tends to be fragile

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +381,113 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+    helper function to generate 1p dependency metrics
+    makes GET requests to maven central to check for 1p dependencies for a small subset of the project's
+    dependencies
+     */
+    private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
+        val httpClient = HttpClient.newHttpClient();
+        val request =  HttpRequest.newBuilder()
+            .uri(URI.create("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json"))
+            .GET()
+            .build();
```
rli, line 394: use our `HttpUtils` or `HttpRequests`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +381,113 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
         managerState.flags[StateFlags.IS_ONGOING] = false
     }
 
+    /**
+    helper function to generate 1p dependency metrics
+    makes GET requests to maven central to check for 1p dependencies for a small subset of the project's
+    dependencies
+     */
+    private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
```
rli, line 389: we need a better way to determine if this is a 1p dependency -- this is customer data and we shouldn't be making API calls to a 3p service

carolabadeer: No concerns with this approach after internal discussions

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
carolabadeer: done

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -73,6 +73,10 @@ import java.io.File
 import java.time.Instant
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.swing.Icon
+import org.jsoup.Jsoup;
+import kotlinx.serialization.json.Json;
+import kotlinx.serialization.json.JsonObject;
```
rli, line 78: please switch to jackson to maintain consistency with the rest of the codebase

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+    makes GET requests to maven central to check for 1p dependencies for a small subset of the project's
+    dependencies
+     */
+    private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
+        try{
+            val response = getTextFromUrl("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json")
+                try {
+                    val json: JsonObject = Json.parseToJsonElement(response) as JsonObject;
+                    val responses = json["response"] as? JsonObject;
+                    val numFound = responses?.get("numFound");
```
rli, line 390: prefer modeling the response with a data class, instead of calling getters on nested objects

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
     }
 
+    /**
+    helper function to generate 1p dependency metrics
+    makes GET requests to maven central to check for 1p dependencies for a small subset of the project's
+    dependencies
+     */
+    private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
+        try{
+            val response = getTextFromUrl("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json")
```
rli, line 386: nit: doesn't group/artifact/version form a unique key? why do we need 20 rows?

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -73,6 +73,10 @@ import java.io.File
 import java.time.Instant
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.swing.Icon
+import org.jsoup.Jsoup;
```
rli, line 76: we don't need yet another deserialization library

take a look at JDOM in `com.intellij.configurationStore` (preferred)
or pull in jackson-xml instead

jetbrains-core/resources/telemetryOverride.json
```diff
@@ -507,6 +512,24 @@
                     "required": false
                 }
             ]
+        },
+        {
+            "name": "codeTransform_projectContains1pDependencies",
```
rli, line 517: new metrics should be declared in https://github.com/aws/aws-toolkit-common

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+    private fun makeGETRequests(groupId: String, artifactId: String, version: String, jobId: JobId): Boolean {
+        try{
+            val response = getTextFromUrl("https://search.maven.org/solrsearch/select?q=g:$groupId%20AND%20a:$artifactId%20AND%20v:$version%20AND%20p:jar&rows=20&wt=json")
+                try {
+                    val json: JsonObject = Json.parseToJsonElement(response) as JsonObject;
+                    val responses = json["response"] as? JsonObject;
+                    val numFound = responses?.get("numFound");
+
+
+                    if(numFound.toString() == "0"){
```
rli, line 393: letting the deserializer coerce the response into a data class will allow you to do an int comparison on an int instead of a string comparison

rli: if you're intentionally interacting with raw `JsonElement`s, you generally want to coerce the JsonElement into the proper type instead of just doing `toString()`

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+                        CodetransformTelemetry.projectContains1pDependencies(
+                            codeTransformSessionId = CodeTransformTelemetryState.instance.getSessionId(),
+                            codeTransformJobId = jobId.id,
+                            has1pdependency = true
+                        )
+                        return true;
+                    }
+                } catch (e: Exception){
+                    LOG.warn{"Error parsing JSON response from GET request to maven central: ${e.printStackTrace()}"}
```
rli, line 402: don't print stack traces directly

```suggestion
                    LOG.warn(e) {"Error parsing JSON response from GET request to maven central"}
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+                }
+        } catch(e: Exception){
+            LOG.warn{"Exception while making GET request to maven central: ${e.printStackTrace()}"}
+        }
```
rli, line 406: you can avoid some nesting here:

```kt
val response = try {
    getTextFromUrl(...)
} catch (e: Exception) {
    LOG.warn/error...
    return false
}

...
...
if (numFound != 0) {
    ...
    return true
}

return false
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -372,6 +376,84 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
...
+
+        try{
+            session.sessionContext.runMavenShellCommand(File(dir), commandList);
+            var aggregateReport = Jsoup.parse(File("$dir/target/dependency-updates-aggregate-report.xml"), "UTF-8");
+
+            var found1pDependency = false;
+            val dependencies = aggregateReport.select("DependencyUpdatesReport dependencies dependency");
+
+            for ((index, dependency) in dependencies.withIndex()){
+                val status = dependency.select("status").text();
```
rli, line 431: same comment about using a data class here

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
rli, line 1: tests? there seems to be a substantial amount of logic introduced

rli (commented):
in general, please try to make sure the required tests are passing before requesting a review

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/CodeModernizerSessionContext.kt
```diff
@@ -328,10 +315,36 @@ data class CodeModernizerSessionContext(
...
+        return shouldTryMvnCommand;
+    }
 
+    /**
+     * @description
+     * this command is used to run the maven commmand which copies all the dependencies to a temp file which we will use to zip our own files to
+     */
+    fun runMavenCommand(sourceFolder: File): File? {
+        val currentTimestamp = System.currentTimeMillis()
+        val destinationDir = Files.createTempDirectory("transformation_dependencies_temp_" + currentTimestamp)
```
rli, line 327: nit
```suggestion
        val destinationDir = Files.createTempDirectory("transformation_dependencies_temp_${System.currentTimeMillis()})
```

jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
rli, line 386: i'd also like to reiterate the same comment i (and our appsec engineer) had:

why can't we figure this out on the service side?
it seems like we're trying to push a lot of business logic to the client side which in turn needs to be duplicated across all the client implementations

carolabadeer: There are future enhancements based on the data from this metric that will be done on the IDE side, which is why this logic is being done on the client side. If you'd like to see specifics, I can share our roadmap with you internally.

jetbrains-core/resources/telemetryOverride.json
carolabadeer: PR to add this metric in toolkit-common 
https://github.com/aws/aws-toolkit-common/pull/683

PR #4132, opened 2024-02-12
Restructure repo layout to clarify feature vertical boundaries

rli:
Along the lines of https://github.com/aws/aws-toolkit-vscode/pull/4342,

We split the structure of the repo to prepare for multiple extensions out of a single monorepo.
This is only a subset of changes to hopefully keep merge conflicts minimal as we migrate buildscripts and further move code between directories.

## License


build.gradle.kts
```diff
@@ -63,8 +68,8 @@ if (idea.project != null) { // may be null during script compilation
         project {
             settings {
                 taskTriggers {
-                    afterSync(":sdk-codegen:generateSdks")
-                    afterSync(":jetbrains-core:generateTelemetry")
+                    afterSync(":plugin-core:sdk-codegen:generateSdks")
+                    afterSync(":plugin-toolkit:jetbrains-core:generateTelemetry")
```
manodnyab, line 72: Should this also be a part of plugin-core?

plugins/toolkit/jetbrains-ultimate/src/test/testData/vgo/src/test/testData/mockGoPath/folderKeep
manodnyab, line 1: It's an empty file, is this required?

rli: i believe the path needs to exist you can't commit an empty folder
this should probably handled a different way but is not in scope for this PR

build.gradle.kts
rli: that is more involved so will happen later

.gitignore
```diff
@@ -17,22 +17,19 @@ stale_outputs_checked
 global.json
 
 # local dev files
-unload-PythonCore-*
-unload-aws.toolkit-*
+unload-*
 jetbrains-core/bin/
 jetbrains-ultimate/bin/
 resources/bin/
+jetbrains-rider/**/NuGet.config
```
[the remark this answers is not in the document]
rli: we can adjust again if we change the paths again, but it shouldn't be necessary

PR #4148, opened 2024-02-23
Bump to IJ EAP7 / Rider EAP5

rli:
## License


gradle.properties
```diff
@@ -8,7 +8,7 @@ toolkitVersion=2.11-SNAPSHOT
 publishToken=
 publishChannel=
 
-ideProfileName=2023.3
+ideProfileName=2024.1
```
manodnyab, line 11: do we want to keep this 2023.3 until we move to 241 stable?

manodnyab (approved):

gradle.properties
rli: sure

PR #4156, opened 2024-02-28
Gumby bug fix - transform complete message shown before job is complete

mk-fan:
## Types of changes

## Description

- Fix a bug that network change will cause wrong "transform complete" notification be shown
  - Now we will instead show "transform failed" notification with a reason of "connection timeout"
- When transformation is cancelled, shown "job is stopped" status in the transformation plan detail step panel (right panel)

Screenshot 2024-02-27 at 10 53 30 PM

## Checklist
 
## License


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeTransformUtils.kt
```diff
@@ -168,15 +168,13 @@ suspend fun JobId.pollTransformationStatusAndPlan(
...
         waitUntil(
-            succeedOn = { state in succeedOn },
-            failOn = { state in failOn },
+            succeedOn = { result -> result in succeedOn },
+            failOn = { result -> result in failOn },
             maxDuration = maxDuration,
             exceptionsToStopOn = setOf(
                 InternalServerException::class,
                 ValidationException::class,
-                AccessDeniedException::class,
```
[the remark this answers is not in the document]
damntrecky: Second Davids comment?

Can we catch the `AccessDeniedException` and `SdkClientException` and then do a subsequent check to make sure we want to continue or cancel? Any metadata fields from the exceptions to look at?

This may require a more in depth review.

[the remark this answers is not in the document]
mk-fan: We will only throw AccessDeniedException if we retry refreshing token 10 times. It shouldn't be part of exceptionsToStopOn because it will cause waitUntil to exit without actually throwing the exception.

I fixed the type and added another catch block outside, so that AccessDeniedException will return a failed PollingResult.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeTransformUtils.kt
```diff
@@ -214,15 +213,19 @@ suspend fun JobId.pollTransformationStatusAndPlan(
...
+                return@waitUntil state
             } catch (e: AccessDeniedException) {
                 if (numRefreshes++ > maxRefreshes) throw e
                 refreshToken(project)
+                return@waitUntil state
             } finally {
                 sleep(sleepDurationMillis)
             }
         }
     } catch (e: WaiterUnrecoverableException) {
```
dhasani23, line 225: Is there a reason we can't just catch any `Exception` at once here?

mk-fan: Changed to generic Exception so that we can handle all exceptions thrown.

Also made the fix so that the status will update when job failed due to exceptions.

damntrecky (approved):

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeTransformUtils.kt
```diff
@@ -214,14 +211,18 @@ suspend fun JobId.pollTransformationStatusAndPlan(
...
             } catch (e: AccessDeniedException) {
                 if (numRefreshes++ > maxRefreshes) throw e
                 refreshToken(project)
+                return@waitUntil state
             } finally {
                 sleep(sleepDurationMillis)
             }
         }
-    } catch (e: WaiterUnrecoverableException) {
+    } catch (e: Exception) {
```
arangatang, line 223: `WaiterUnrecoverableException` means that we ended up in a `failOn` state. 
But other exceptions could happen here like `AccessDeniedException` or other unhandled exceptions. 

I'm not sure we want to hide other possible exceptions here, so how about: 

```
catch (e: Exception) {
  when (ex):
     is WaiterUnrecoverableException, AccessDeniedException -> {what we do now}
     else -> throw
}
```

arangatang: Also could we log the error / stack trace, would ease debugging i recon

mk-fan: Changed to throwing exception if it's not WaiterUnrecoverableException or AccessDeniedException.

The logging is already done in `pollUntilJobCompletion`.

rli (approved):

PR #4183, opened 2024-03-12
Split out CodeWhisperer and core extension declarations

rli:
This PR pulls out the relevant core services into the `core` module and partially moves the CodeWhisperer implementations into the `amazonq/codewhisperer` module.

We expose the plugin.xml for both as something that can be included independently to allow us to emit the same artifact we publish to the marketplace today, as well as prepare for testing it as an independent entity.

As part of this, some minor changes were required to uncouple clearly toolkit-only bits from shared functionality. There are still some toolkit-only bits that do not belong in core, but any non-trivial changes are out of scope for this PR.
 
The idea of using `testFixtures` instead of emitting a `testJar` also appears promising and has cleaned up some annoying boilerplate that we don't need to transfer to the new submodules. During this investigation, there was another discovery that we can replace the default output artifact with the `instrumentedJar` variant. This allows us clean up some other boilerplate and should make the cross-module development experience slightly better now that the IDE no longer sometimes navigates you to the instrumented classes instead of the raw Kotlin.

## License


plugins/amazonq/codewhisperer/jetbrains-community/build.gradle.kts
```diff
@@ -12,5 +12,7 @@ intellijToolkit {
 }
 
 dependencies {
-    compileOnly(project(":plugin-amazonq:shared:jetbrains-community"))
+    implementation(project(":plugin-amazonq:shared:jetbrains-community"))
+
+    compileOnly(project(":plugin-core:jetbrains-community"))
```
manodnyab, line 17: Would this also be required for implementation for the common bits?

plugins/amazonq/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,17 @@
...
+
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude">
+    <id>plugin-amazonq</id>
+    <version>1.0</version>
+    <idea-version since-build="231" until-build="999"/>
+
+    <depends>core</depends>
+    <xi:include href="/META-INF/module-amazonq.xml" />
+
+    <extensions defaultExtensionNs="com.intellij">
```
manodnyab, line 12: does this block need to be defined here if it's currently empty or will there be additions later?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/AwsResourceCache.kt
```diff
@@ -287,39 +282,6 @@ class ClientBackedCachedResource<ReturnType, ClientType : SdkClient>(
     override fun toString(): String = "ClientBackedCachedResource(id='$id')"
 }
 
-class ExecutableBackedCacheResource<ReturnType, ExecType : ExecutableType<*>>(
```
manodnyab, line 290: Should this continue to be here?

plugins/amazonq/codewhisperer/jetbrains-community/build.gradle.kts
rli: we want to be able to compile but avoid pulling in the jars at runtime because of the plugin classloader issue when the same class has been loaded by another plugin

plugins/amazonq/src/main/resources/META-INF/plugin.xml
rli: this will go away later

plugins/core/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,21 @@
+<!-- Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude">
+    <id>core</id>
+    <name>core</name>
+    <description>corecorecorecorecorecorecorecorecorecore</description>
```
manodnyab, line 7: nit: Common functionalities required for AWS Toolkit and Amazon Q plugins

manodnyab (approved):

PR #4194, opened 2024-03-18
fix(CodeTransformation): remove play/stop buttons from bottom panel

dhasani23:
## Types of changes

## Description

Removed the play/stop buttons from the Transformation Hub since we're moving entirely to the chat.

## Checklist
 
## License


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -240,7 +240,7 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
         // New projectDetails metric should always be fired whether the project was valid or invalid
         CodetransformTelemetry.projectDetails(
             codeTransformSessionId = CodeTransformTelemetryState.instance.getSessionId(),
-            result = if (!validationResult.valid) Result.Failed else Result.Unknown,
+            result = if (!validationResult.valid) Result.Failed else Result.Succeeded,
```
dhasani23, line 243: Dhruvi mentioned this cannot be unknown, should be Failed or Succeeded.

.changes/next-release/removal-6d1b2bf1-ed21-467c-9333-13d062442906.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type" : "removal",
+  "description" : "CodeTransformation: remove play/stop buttons from Transformation Hub"
```
rli, line 3: elaborate on what the replacement is from a customer perspective

rli (approved):

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
damntrecky: Interesting, I did not know that.. Its undefined for VSCode which is a valid value. Maybe null would be better here. Its not a true "success metric"

plugins/toolkit/jetbrains-core/resources/META-INF/ext-java.xml
```diff
@@ -48,13 +48,6 @@
     </extensions>
     <actions>
         <group id="aws.toolkit.codemodernizer.toolbar">
-            <action
-                id="codemodernizer.toolbar.startmodernizer"
-                class="software.aws.toolkits.jetbrains.services.codemodernizer.actions.CodeModernizerStartModernizerAction"/>
-            <action
```
damntrecky, line 54: Question: Why are we getting rid of the "stop" action? The only stop button would be in the chat and until we get rid of the bottom panel progress window, I don't think we should remove the stop action from there.

damntrecky (approved):
Tentative approval, I would like more context on why we are removing the stop button.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
dhasani23: Pretty sure `result` has to always be present though

plugins/toolkit/jetbrains-core/resources/META-INF/ext-java.xml
dhasani23: Added it back, you're right that we should keep it

PR #4198, opened 2024-03-19
Split out CodeTransform and Q Chat extension declarations

rli:
Continuation of concept in #4183, but for CodeTransform and chat
 
## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeTransformUtils2.kt
```diff
@@ -0,0 +1,14 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
manodnyab, line 3: Do we want to change this file name or include it in the already available utils file?

manodnyab (approved):

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeTransformUtils2.kt
rli: if you keep the same name then you get a runtime error because there are 2 classes with the same name

PR #4239, opened 2024-04-05
Add Github action to generate prerelease builds

manodnyab:
## Description

This generates prerelease builds for all 3 plugins

## Checklist
 
## License


justinmk3 (approved):
Nice

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,166 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'corretto'
+          java-version: '17'
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '6.x'
```
rli, line 82: core an q don't need dotnet

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Generate artifact
+        run: |
+          ./gradlew -PideProfileName=${{ matrix.supported_versions }} :plugin-toolkit:intellij:buildPlugin --exclude-task buildResharperPlugin
```
rli, line 33: let's also do `intellij-standalone`

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: plugin-toolkit-${{ matrix.supported_versions }}
+          path: ./plugins/toolkit/intellij/build/distributions/*.zip
+          retention-days: 1
+
+  generate_artifact_q:
```
rli, line 41: as a future optimization, we can combine these all into a single job and just run `./gradlew buildPlugin`
as-is we're repeatedly downloading tens of gigs of dependencies

we will need to disable that task on the subprojects where it does not make sense

manodnyab: Yeah, that sounds good, wanted to keep this separate currently cause:
1) Their creation is independent of each other 
2) the running out of disk space issue

.github/workflows/prerelease.yml
manodnyab: Added

rli (approved):

PR #4252, opened 2024-04-09
Add support for OAuth login flow using authorization_grant with PKCE

rli:
This change allows the PCKE login flow to be used after enabling a developer flag. `sdk-codegen` is untouched other than bringing in the new service model from internal and the existing SDK customization from public GitHub.

During testing, a classpath conflict on Netty was discovered and it appears that we can safely exclude it from our plugin distribution.

On the token cache side, we enforce distinction during de/serialization since the registration/access token pairs are not compatible with each other (PKCE requires new client registration instances).

Otherwise the entirety of the PKCE grant can be expressed in `ToolkitOAuthService.kt` by overriding the relevant abstract methods in the standard IDE base classes.
 
## License


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/SsoAccessTokenProvider.kt
```diff
@@ -63,37 +85,79 @@ class SsoAccessTokenProvider(
...
+
+        return registeredClient
+    }
+
+    private fun registerPkceClient(): PKCEClientRegistration {
+        loadPkceClientRegistration()?.let {
+            return it
+        }
+
+        if (!ssoUrl.contains("identitycenter")) {
```
manodnyab, line 129: Should we have a test for this if we are doing text parsing?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/pkce/ToolkitOAuthService.kt
```diff
@@ -0,0 +1,170 @@
...
+        if (toolkitRequest.csrfToken != callbackState) {
+            request.result.completeExceptionally(RuntimeException("Invalid CSRF token"))
+            return false
+        }
+
+        return super.handleServerCallback(path, parameters)
+    }
+
+    override fun revokeToken(token: String) {
+        TODO("Not yet implemented")
```
manodnyab, line 69: nit: Should this be commented?

manodnyab (approved):

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/SsoAccessTokenProvider.kt
rli: it's just a warning, service is right place to make the final call

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/pkce/ToolkitOAuthService.kt
rli: could not figure out what calls this

PR #4316, opened 2024-04-19
fix(amazonq): remove single quotes from messages

dhasani23:
## Types of changes

## Description

When using the `message` function on a String, it is supposed to replace the `{0}` in the String with a provided value, however this does not work when the String contains a single quote in it (such as `couldn't`). 

We could escape this, but our PM does not want contractions in messages anyway, so solving this by removing the single quotes.

## Checklist
 
## License


plugins/toolkit/resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
```diff
@@ -541,9 +541,9 @@ codemodernizer.chat.message.button.view_summary=View summary
 codemodernizer.chat.message.error_request=Request failed
 codemodernizer.chat.message.follow_up.new_transformation=Start a new transformation
 codemodernizer.chat.message.local_build_begin=I'm building your module. This can take up to 10 minutes, depending on the size of your module.
-codemodernizer.chat.message.local_build_failed=Sorry, I couldn't run the Maven clean install command to build your module.
```
mk-fan, line 544: Is there no way to escape the single quote character?

damntrecky: Have we asked the toolkit team about this. Usually for JS/TS you have string escape delimiters for special characters.


-    \\ – Backslash
 -   \' – Single quote
   -  \" – Double quote
    - \n – Newline
-    \r – Carriage return
  -  \t – Tab

damntrecky: Have we asked the toolkit team about this. Usually for JS/TS you have string escape delimiters for special characters.


-    \\ – Backslash
-   \' – Single quote
   -  \" – Double quote
    - \n – Newline
-    \r – Carriage return
  -  \t – Tab

mk-fan: https://stackoverflow.com/questions/4449639/apostrophe-doesnt-get-translated-properly-when-placed-in-a-resource-bundle

dhasani23: Spoke to PM and we don't want contractions in our messages anyway. So `could not` is preferred over `couldn't`.

manodnyab (approved):

PR #4344, opened 2024-04-23
Remove profile field from new login UX

rli:
Per internal testing and prior UX discussion, the "profile" field should be inferred from the input instead of requiring users to choose

## Description
Screenshot 2024-04-23 at 1 00 00 PM
 
## License


andrewyuq (approved):

plugins/toolkit/core/tst/software/aws/toolkits/core/credentials/SsoUrlIdentifierTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.ArgumentsProvider
+import org.junit.jupiter.params.provider.ArgumentsSource
+import java.util.stream.Stream
+
+class SsoUrlIdentifierTest {
+    companion object {
+        class ValidArgs : ArgumentsProvider {
+            override fun provideArguments(context: ExtensionContext?) = Stream.of(
+                Arguments.of("https://test.awsapps.com/start", "test"),
```
manodnyab, line 19: Can we add a test for the situation where there are similar start urls?

manodnyab: Discussed offline, for the same start url we'll update it. Part of another PR

manodnyab (approved):

PR #4362, opened 2024-04-24
Use reflection to determine product name and version for Telemetry

breedloj:
## Types of changes

## Description
Uses a mix of JB internal utilities and additional logic to determine the current plugin via its entry point on the stack to allow product name and version to be dynamically identified. 

* Stack is traversed from deepest frame first, filtered by Toolkit prefixed FQCNs, until a plugin ID can be identified (should always be on the first attempt). JB's built in utilities tend to go shallow->deep which result in core classes being resolved against.
* For product resolution default is always `AWS_TOOLKIT_FOR_JET_BRAINS` unless the Q extension is explicitly identified. This can be amended as needed in the future.
* Version returns an explicit `unknown` literal in case of resolution failure - this is a case that should never occur but previously would have resulted in an NPE. Please let me know if there is another agreed upon convention for this (e.g. omission or explicit null).
* Unit tests are added for the new resolver class - integration testing end-to-end through each plugin has thus far proven difficult due to the reliance of this logic on the `PluginAwareClassLoader` which is only present when a plugin is properly loaded in the IDE.

## Checklist
 
## License


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt
```diff
@@ -90,8 +90,9 @@ open class AwsClientManager : ToolkitClientManager(), Disposable {
         val userAgent: String by lazy {
             val platformName = tryOrNull { ApplicationNamesInfo.getInstance().fullProductNameWithEdition.replace(' ', '-') }
             val platformVersion = tryOrNull { ApplicationInfoEx.getInstanceEx().fullVersion.replace(' ', '-') }
-            val pluginVersion = tryOrNull { PluginManager.getPluginByClass(this::class.java)?.version }
-            "AWS-Toolkit-For-JetBrains/$pluginVersion $platformName/$platformVersion ClientId/${AwsSettings.getInstance().clientId}"
+            val pluginName = PluginResolver.getInstance().product.toString().replace(" ", "-")
+            val pluginVersion = tryOrNull { PluginResolver.getInstance().version }
+            "$pluginName/$pluginVersion $platformName/$platformVersion ClientId/${AwsSettings.getInstance().clientId}"
```
rli, line 95: i think the SDKs are mostly independent between extensions where it's not really a big problem, but there's also the case where a given client is used by both extensions so we'd need to have the user agent be dynamic on the caller

plugins/core/sdk-codegen/codegen-resources/telemetry/service-2.json
rli, line 1: maybe the normal variant would have been a smaller diff

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/telemetry/ClientMetadata.kt
```diff
@@ -3,16 +3,15 @@
...
 import com.intellij.openapi.application.ApplicationNamesInfo
 import com.intellij.openapi.util.SystemInfo
 import software.amazon.awssdk.services.toolkittelemetry.model.AWSProduct
 import software.aws.toolkits.jetbrains.settings.AwsSettings
 
 data class ClientMetadata(
-    val productName: AWSProduct = AWSProduct.AWS_TOOLKIT_FOR_JET_BRAINS,
-    val productVersion: String = PluginManager.getPluginByClass(this::class.java)?.version.toString(),
+    val productName: AWSProduct = PluginResolver.getInstance().product,
+    val productVersion: String = PluginResolver.getInstance().version,
```
rli, line 14: wouldn't we need to inspect on each call?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt
breedloj: Good point. Still working out a way to provide a UA at the request level. As of the current change they will be generated per SDK client instance.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/telemetry/ClientMetadata.kt
breedloj: Addressed in the latest revision - each metric event captures the caller's plugin product/version and these are grouped at time of telemetry publish.

rli (approved):

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/AwsClientManager.kt
breedloj: Addressed this

PR #4444, opened 2024-05-07
Move remaining CodeWhisperer code into Q plugin

rli:
Resolves hacks left in place to allow plugins to be built while the repository was partially split. Additionally allows us to stop bundling the entire toolkit with the Q plugin.
 
## License


rli:
Failing windows UTG-related test is fixed in https://github.com/aws/aws-toolkit-jetbrains/pull/4488

Will-ShaoHua (approved):

PR #4452, opened 2024-05-08
Fix flaky test `mavenRootPomHandlerBaseDirIsCorrect `

vchikoti1998:
## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Description
The test case mavenRootPomHandlerBaseDirIsCorrect() was updated to use `runBlocking` instead of `runTest`.  `runBlocking` is a Kotlin coroutines function that helps bridge regular blocking code (code that stops the current thread) 
with coroutine based code (code that suspends execution and allows other tasks to run concurrently). 
Using `runBlocking` resolved the issue with uncaught exceptions that was occurring when using the more specialized 
`runTest` function.
## Checklist
 
## License


plugins/toolkit/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
```diff
@@ -61,7 +61,7 @@ class JavaLambdaBuilderTest {
     }
 
     @Test
-    fun mavenRootPomHandlerBaseDirIsCorrect() = runTest {
+    fun mavenRootPomHandlerBaseDirIsCorrect(): Unit = runBlocking {
```
rli, line 64: tests should use runTest

gradle/libs.versions.toml
```diff
@@ -17,7 +17,7 @@ junit5 = "5.10.2"
 # https://kotlinlang.org/docs/releases.html#release-details
 kotlin = "1.9.23"
 # set in <root>/settings.gradle.kts
-kotlinCoroutines = "1.7.3"
+kotlinCoroutines = "1.8.0"
```
rli, line 20: coroutines can't be safely bumped like this
https://github.com/aws/aws-toolkit-jetbrains/blob/main/kotlinResolution.settings.gradle.kts

vchikoti1998: Sure

plugins/toolkit/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
vchikoti1998: will revert it

rli: discussed offline, the issue is with `runTest` and the fix is in a later version of coroutines-test

rli, line 64: it would also be good to add a comment in the source code explaining why we're using `runBlocking`
```suggestion
    fun mavenRootPomHandlerBaseDirIsCorrect() = runBlocking {
```

rli (approved):

plugins/toolkit/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/lambda/java/JavaLambdaBuilderTest.kt
vchikoti1998: sure Definitely

PR #4454, opened 2024-05-09
fix flaky behaviour [ Test load tab state]

vchikoti1998:
## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Description
The test is failing because the loadState is being called out of runInEdt. When runInEdt is added, thread safety is enforced and the test passes.
## Checklist
 
## License


plugins/toolkit/jetbrains-core/tst/software/aws/toolkits/jetbrains/core/explorer/AwsToolkitExplorerToolWindowTest.kt
```diff
@@ -40,22 +40,23 @@ class AwsToolkitExplorerToolWindowTest {
...
-            }
-        )
-        assertThat(sut.state.selectedTab).isEqualTo(AwsToolkitExplorerToolWindow.Q_TAB_ID)
+            sut.loadState(
+                AwsToolkitExplorerToolWindowState().apply {
+                    selectedTab =
+                        AwsToolkitExplorerToolWindow.DEVTOOLS_TAB_ID
+                }
+            )
+            assertThat(sut.state.selectedTab).isEqualTo(AwsToolkitExplorerToolWindow.Q_TAB_ID)
```
manodnyab, line 58: Should this test change now that user can dismiss the Q tab?

vchikoti1998: In this we are just testing whether Q_TAB is selected or not. Actually this test passes but sometimes fails on CI and local. So, running this test in  `runInEdtAndGet {}  ` block. To avoid flakiness.
> Should this test change now that user can dismiss the Q tab?

I think this is an another test case, isn't it?

rli (approved):

plugins/toolkit/jetbrains-core/tst/software/aws/toolkits/jetbrains/core/explorer/AwsToolkitExplorerToolWindowTest.kt
rli: per discussion with Ankit, leave default to be Q_TAB

vchikoti1998: Default tab selected should be Q_TAB, checked with @ankit-sharma90

PR #4471, opened 2024-05-14
fix(AmazonQ FeatureDev): throws an error when the repo is bigger than 200MB

gasolima:
Problem: Backend can't handle repo size larger than 200MB
Solution: Throws an error to give the user better experience

## Types of changes

## Description

Screenshot 2024-05-14 at 11 03 08

## Checklist
 
## License


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/controller/FeatureDevController.kt
```diff
@@ -430,10 +431,10 @@ class FeatureDevController(
             }
         } catch (err: Exception) {
             logger.warn(err) { "Encountered ${err.message} for tabId: $tabId" }
-            if (err is ContentLengthError) {
+            if (err is RepoSizeLimitError || err is ContentLengthError) {
                 messenger.sendError(
                     tabId = tabId,
-                    errMessage = err.message,
+                    errMessage = err.message!!,
```
gasolima, line 437: I don't know why i need to use `!!` in this case, since the 2 errors has message. But if i removed it the compiler complain saying that the message can be optional and it's expected to be required

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/Session.kt
```diff
@@ -39,7 +40,7 @@ class Session(val tabID: String, val project: Project) {
     var isAuthenticating: Boolean
 
     init {
-        context = FeatureDevSessionContext(project)
+        context = FeatureDevSessionContext(project, maxProjectSizeBytes)
```
sannicm, line 43: question: why adding the size to the context object as an input?

You could just add the value as a private val inside the class and use it.

.changes/next-release/bugfix-6871c109-4229-4247-a897-debfba52d980.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "Amazon Q Feature Development: Update error message when repo size larger than 200 megabytes"
```
sannicm, line 3: issue: I would say this is more a feature as you are adding a check for the repo size prior to uploading the artifacts regardless of using a new error message or not.

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/controller/FeatureDevController.kt
sannicm: Most probably the type is inferring from RepoSizeLimitError has a possible undefined message. And I guess it is becuase it just extends RuntimeException(). Maybe the root cause is on that.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -92,16 +94,24 @@ class FeatureDevSessionContext(val project: Project) {
...
-                        return true
-                    }
                     return runBlocking {
-                        !ignoreFile(file, this)
+                        if (file.isFile && !ignoreFile(file, this)) {
+                            totalSize += file.length
+                            files.add(file)
+
+                            if (maxProjectSizeBytes != null && totalSize > maxProjectSizeBytes) {
+                                throw RepoSizeLimitError(message("amazonqFeatureDev.content_length.error_text"))
```
sannicm, line 109: question: did you check if this is not breaking another team's functionality?

gasolima: if `maxProjectSizeBytes` isn't send to this class it will not do anything

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/Session.kt
gasolima: Maybe another team want to use this class with different `maxProjectSizeBytes`

.changes/next-release/bugfix-6871c109-4229-4247-a897-debfba52d980.json
gasolima: if other people see it as a feature, i can change it as well. Don't have a strict opinion about it

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/controller/FeatureDevController.kt
matboros: does it work in a switch statement?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -92,16 +94,24 @@ class FeatureDevSessionContext(val project: Project) {
...
+                            }
+                            return@runBlocking true
+                        } else {
+                            return@runBlocking !ignoreFile(file, this)
```
sarayev, line 113: With this implementation, we'll run `ignoreFile(file, this)` function twice for the files that should be ignored. Can we - either store result of `ignoreFile()` and use it here
- or add a check to this else statement that checks if it is not file, then call this function?

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/controller/FeatureDevController.kt
gasolima: But both of them are extended from `RuntimeException()`, why it's working fine with both of them separate in different if conditions?

Anyway if you know how to fix it, let's huddle

gasolima: > does it work in a switch statement?

Tried didn't work :(

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
gasolima: I can put it in a variable and reuse it

sarayev: Asking because, the ignoreFile function is doing pattern matching based on the regexp and this implementation can increase the duration needed to check files.

gasolima: > Asking because, the ignoreFile function is doing pattern matching based on the regexp and this implementation can increase the duration needed to check files.

Yes will make it slower, by how much not sure. But adding another variable isn't a big deal so i can do it anyway

.changes/next-release/bugfix-6871c109-4229-4247-a897-debfba52d980.json
sannicm: Well, it is a feature right? This check didn't exist in the logic before.

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/Session.kt
sannicm: I would doubt so, this is a specific class from our feature, very improbable that it will be a shared class.

gasolima: > very improbable that it will be a shared class.

At least this class now not under FeatureDev folder, so it's already in a shared space, even no one is using it

.changes/next-release/bugfix-6871c109-4229-4247-a897-debfba52d980.json
gasolima: I don't see it as a feature, since we were always were failing to treat those repos. But before this CR the user was seeing another error after calling the execute planning API, with no buttons. So this CR is only fixing the Error to be appropriate(with button) + fail fast. 

Let me know if you still see it as a feature, i can change it to feature

sannicm: I see, thanks for the explanation.

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevConstants.kt
```diff
@@ -13,3 +13,6 @@ const val CODE_GENERATION_RETRY_LIMIT = 3
 
 // The default retry limit used when the session could not be found
 const val DEFAULT_RETRY_LIMIT = 0
+
+// Max allowed size for a repository in bytes
+const val maxProjectSizeBytes: Long = 200 * 1024 * 1024
```
sarayev, line 18: nit: Can we have this constant name in ALL_CONSTANT, like MAX_PROJECT_SIZE_IN_BYTES? That seems the pattern for all constants in this file.

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/Session.kt
sarayev: I think that class should be renamed at some point. It is being used by anothe team and is not located in /featureDev folder.

sannicm: I see the file has also a TODO that also shows it is an intended shared functionality

.changes/next-release/bugfix-6871c109-4229-4247-a897-debfba52d980.json
sarayev: To me it also looks like a bugfix - previously there was a bug that, repos larger than 200 MB were being sent to BE and we were failing to generate a plan/code for it. Now, it is fixed, and FE doesn't send that repo to BE and fails gracefully.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -31,7 +31,12 @@ import java.util.zip.ZipOutputStream
 import kotlin.io.path.Path
 import kotlin.io.path.relativeTo
 
-class FeatureDevSessionContext(val project: Project) {
+interface RepoSizeError {
```
sannicm, line 34: praise for the message!! fix.

sannicm (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevConstants.kt
gasolima: sure, my bad

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -92,18 +97,26 @@ class FeatureDevSessionContext(val project: Project) {
...
             object : VirtualFileVisitor<Unit>() {
-                override fun visitFile(file: VirtualFile): Boolean {
-                    if (file.isFile) {
-                        files.add(file)
-                        return true
-                    }
-                    return runBlocking {
-                        !ignoreFile(file, this)
+                override fun visitFile(file: VirtualFile): Boolean =
+                    runBlocking {
```
manodnyab, line 106: Do we need to block the thread for the entire function? We should try to avoid using runBlocking in non-test functions. Im guessing we need it here for ignoreFile(). maybe we could change it to
val isFileIgnored = runBlocking {return ignoreFile....}
if(....)

gasolima: Yeah good tip, i did it in the next rev

sarayev (approved):

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -92,17 +97,24 @@ class FeatureDevSessionContext(val project: Project) {
...
             object : VirtualFileVisitor<Unit>() {
                 override fun visitFile(file: VirtualFile): Boolean {
-                    if (file.isFile) {
-                        files.add(file)
-                        return true
-                    }
-                    return runBlocking {
-                        !ignoreFile(file, this)
-                    }
+                    val isFileIgnored = runBlocking { return@runBlocking ignoreFile(file, this) };
```
manodnyab, line 106:
```suggestion
                    val isFileIgnored = runBlocking {  ignoreFile(file, this) }
```

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -92,17 +97,24 @@ class FeatureDevSessionContext(val project: Project) {
...
+                        if (file.isFile && !isFileIgnored) {
+                            totalSize += file.length
+                            files.add(file)
+
+                            if (maxProjectSizeBytes != null && totalSize > maxProjectSizeBytes) {
+                                throw RepoSizeLimitError(message("amazonqFeatureDev.content_length.error_text"))
+                            }
+                            return true;
```
manodnyab, line 114: nit: Unnecessary semicolon

manodnyab (approved):

sarayev (approved):

dogusata (approved):

PR #4493, opened 2024-05-20
Security scans: Increase in payload size limit from 500MB to 1GB

laileni-aws:
## Types of changes
- **Security scans:** Increase in payload size limit from 500MB to 1GB.

## Checklist
 
## License


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/sessionconfig/CodeScanSessionConfig.kt
```diff
@@ -79,7 +79,7 @@ class CodeScanSessionConfig(
         }
 
         // Fail fast if the selected file size is greater than the payload limit.
-        if (selectedFile != null && selectedFile.length > getPayloadLimitInBytes()) {
+        if ((selectedFile != null && selectedFile.length > getPayloadLimitInBytes()) || (getProjectPayloadMetadata().payloadSize > getPayloadLimitInBytes())) {
```
ashishrp-aws, line 82: We can't throw this here. It will never pass because we truncate the payload in `getProjectPayloadMetadata`. Instead of truncation, let's break project payload metadata and throw error there.

laileni-aws: I can revert this commit and do this in separate PR as this is not just adding a check.
We need to revert entire code with `isProjectTruncated`.

ctlai95 (approved):

ashishrp-aws (approved):

PR #4507, opened 2024-05-23
stop using cwsprConnection

Will-ShaoHua:
Only usage after changes is migration path

image

## Checklist
 
## License


plugins/toolkit/jetbrains-core/src/migration/software/aws/toolkits/jetbrains/services/codewhisperer/explorer/CodeWhispererExplorerActionManager.kt
```diff
@@ -38,7 +38,7 @@ class CodeWhispererExplorerActionManager : PersistentStateComponent<CodeWhispere
     }
 
     private fun getCodeWhispererConnectionStartUrl(project: Project): String {
-        val connection = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(CodeWhispererConnection.getInstance())
+        val connection = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance())
```
rli, line 41: needs to handle case where inline suggestions continue to work if user only has inline scopes

Issue #4509, opened 2024-05-24
Amazon Q does not open the browser when requesting Confirmation Code


tbjuhasz:
**Describe the bug**
Amazon Q plugin does not open the window to get the confirmation code. 

![image](https://github.com/aws/aws-toolkit-jetbrains/assets/139776071/5fc33457-3fac-43e6-ae4a-c883321c9652)

**To reproduce**
1. Install Amazon Q 
2. Click on Use for free
3. Click Continue
4. Its waiting on the browser to open
**Expected behavior**
It should open a browser with and provide the confirmation code. 

**Screenshots**
![image](https://github.com/aws/aws-toolkit-jetbrains/assets/139776071/2af976f8-6f5e-41b1-a5ea-7abe30f49df4)
**Your Environment**
- OS: MacOS
- JetBrains product: PyChar
- JetBrains product version:PyCharm 2024.1.1
- AWS Toolkit version: 3.5-241
- SAM CLI version:
- JVM/Python version: 3.10.1

**Additional context**

Will-ShaoHua:
hey @tbjuhasz, thanks for reporting, could you help confirm your `~/.aws/config` file is following the format where there is no space character.
```
[default]
aws_access_key_id=fooooooooooooooooo
aws_secret_access_key=barrrrrrrrr
```

also would you mind share debug logs with us?

```
pycharm | Help | Diagnostic tools | debug log settings
paste software.aws.toolkits
pycharm | Help | show log in finder
```

PR #4513, opened 2024-05-24
surface error message to users if plugin can't read aws/config file

Will-ShaoHua:
## Problem
When the plugin can't read config file (possible root cause is invalid format like space characters) or file R/W access), the login flow will return and we're not surfacing the error to users so that the login flow seems to be ir-responsive after users clicking login button.

#4509 

## Solution

surface the error to users

image

## Types of changes

## Checklist
 
## License


rli (approved):

Will-ShaoHua:
should figure out a better way to do this, maybe a toast notification

PR #4520, opened 2024-05-28
Q Code Transform: Ignore jobs started more than 24h ago when resuming jobs.

arangatang:
## Types of changes
- [X] New feature (non-breaking change which adds functionality)

### Problem
We noticed in our telemetry that the resumeJob functionality of QCT executed for very old jobs which are no longer available in our systems. This happens as the state was not cleaned until a new job was submitted or if a successful resume job attempt was made. However when a user did not open the project before the backend retention period expired then we could never resume the job successfully as the GetTransformationJob api would reject the request every time.

### Solution
1. Store start time of job into state
2. Use start time to expire jobs stored in the state from being resumed after 24h

## Checklist
 
## License


plugins/amazonq/codetransform/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/codemodernizer/CodeWhispererCodeModernizerTest.kt
```diff
@@ -156,4 +164,103 @@ class CodeWhispererCodeModernizerTest : CodeWhispererCodeModernizerTestBase() {
         )
         assertEquals(expectedResult, result)
     }
+
```
damntrecky, line 167: Comment: Adding tests is always an improvement!

plugins/amazonq/codetransform/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
```diff
@@ -477,7 +479,15 @@ class CodeModernizerManager(private val project: Project) : PersistentStateCompo
             }
 
             LOG.info { "Attempting to resume job, current state is: $managerState" }
-            if (!managerState.flags.getOrDefault(StateFlags.IS_ONGOING, false)) return@launch
+            if (!managerState.isJobOngoing()) return@launch
+
+            val lastJobId = managerState.getLatestJobId()
+            val isJobExpired = managerState.getLatestJobStartTime().plusMillis(JOB_EXPIRY_TIME_DAYS.days.inWholeMilliseconds) < Instant.now()
+            if (isJobExpired) {
+                LOG.info { "Last job is expired, won't attempt to resume job with Id $lastJobId anymore" }
```
damntrecky, line 487: Blocker/Question:

Just need to finalize the message string and if this message needs to:
1. Be displayed to users
2. Logged to .idea logs
3. Other

damntrecky (changes requested):
Address the messaging question for the users.

plugins/amazonq/codetransform/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codemodernizer/CodeModernizerManager.kt
arangatang: 1. Currently this is a silent operation and we only inform customers on success, so don't think we should notify users here
2. Probably we don't have to log this, i'll remove this
3. Not sure what Other means :D

PR #4532, opened 2024-05-30
fix(amazonq): use generic patch applier

ctlai95:
## Types of changes

### Problem
`PlainSimplePatchApplier` is too strict and does not allow any differences in the diff patch and actual code. This is a problem because the document would likely change after a security scan, either by manual edits or by previous fixes being applied.

### Solution
Use `GenericPatchApplier.applySomehow` which 1) tries to use `PlainSimplePatchApplier` first anyways so we are not risking anything with this change and 2) is much more lenient in incorrect line numbers or differing neighboring lines from the location of the change.

## Checklist
 
## License

### DIFF
--- a/.changes/next-release/bugfix-04f91235-8b99-45c3-a4bd-096c530739af.json
+++ b/.changes/next-release/bugfix-04f91235-8b99-45c3-a4bd-096c530739af.json
@@ -1,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "Security Scan: Improved method of applying security fixes"
+}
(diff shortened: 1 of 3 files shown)


laileni-aws (approved):

.changes/next-release/bugfix-04f91235-8b99-45c3-a4bd-096c530739af.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "Security Scan: Improved method of applying security fixes"
```
manodnyab, line 3: nit: Is the effect for the user going to be the same, if yes, then do we need the changelog?

manodnyab (approved):

.changes/next-release/bugfix-04f91235-8b99-45c3-a4bd-096c530739af.json
ctlai95: The effect is the same, the difference is that it is less likely to fail with this change. Do you suggest a different changelog message?

PR #4535, opened 2024-05-31
fix(amazonq): fix issue underline peristing after apply fix

ctlai95:
## Types of changes

### Problem
1. After auto-scan rescans, the yellow underline is not removed even if the issue node is removed.
2. Sometimes yellow underline is not removed when modifying the text range that overlaps with the issue.
3. After the apply fix action, the issue that was fixed should be immediately removed (if auto-scanning is enabled).

### Solution
1. For each auto-scan render, always remove all underlines by clearing the rangeHighlighter
2. Switch the order so that text attributes are unset before disposing the rangeHighlighter.
3. Add a method to remove an issue node by findingId and refresh the UI

## Checklist
 
## License

### DIFF
--- a/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/CodeWhispererCodeScanManager.kt
+++ b/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/CodeWhispererCodeScanManager.kt
@@ -412,29 +412,8 @@ class CodeWhispererCodeScanManager(val project: Project) {
     fun getOverlappingScanNodes(file: VirtualFile, range: TextRange): List<DefaultMutableTreeNode> = synchronized(scanNodesLookup) {
         scanNodesLookup[file]?.mapNotNull { node ->
             val issue = node.userObject as CodeWhispererCodeScanIssue
-            if (issue.textRange?.overlaps(range) == true) node else null
+            if (issue.textRange?.overlaps(range) == true && !issue.isInvalid) node else null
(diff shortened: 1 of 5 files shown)


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/CodeWhispererCodeScanManager.kt
```diff
@@ -412,29 +412,8 @@ class CodeWhispererCodeScanManager(val project: Project) {
     fun getOverlappingScanNodes(file: VirtualFile, range: TextRange): List<DefaultMutableTreeNode> = synchronized(scanNodesLookup) {
         scanNodesLookup[file]?.mapNotNull { node ->
             val issue = node.userObject as CodeWhispererCodeScanIssue
-            if (issue.textRange?.overlaps(range) == true) node else null
-        } ?: listOf()
-    }
-
-    fun updateScanNodesForIssuesOutOfTextRange(fixedIssue: CodeWhispererCodeScanIssue) {
```
ashishrp-aws, line 419: This function updates issue scan nodes when the Applied Fix is out of issue text range(apply fix text change is not in yellow line area). Have you checked for this case

ctlai95: Yes in this case we are now removing the issue directly by findingId

ashishrp-aws (approved):

PR #4590, opened 2024-06-17
Properly resolve plugin in worker thread tasks

breedloj:
## Types of changes

## Description
This change addresses an issue where tasks run in worker threads would be unable to properly resolve the original plugin calling context since the original call stack is lost. This resulted in bad telemetry in these scenarios when running the Amazon Q plugin.

To address this, this change refactors the PluginResolver class to allow each thread to have its own `PluginResolver` instance, while also supporting the ability for child threads to use the instance created by the parent thread. By using ThreadLocal, each thread has its own context for storing and retrieving the PluginResolver instance. A new threading utility function, `pluginAwareExecuteOnPooledThread`, wraps the original pooled thread execution logic and injects a resolved `PluginResolver` instance from the parent thread on the worker thread. This results in the worker inheriting the parent's correctly resolved plugin identifier.

## Checklist
 
## License

### DIFF
--- a/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/QLoginWebview.kt
+++ b/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/QLoginWebview.kt
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.DataContext
-import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.components.service
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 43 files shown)


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/QLoginWebview.kt
```diff
@@ -164,7 +164,7 @@ class QWebviewBrowser(val project: Project, private val parentDisposable: Dispos
             "reauth" -> {
                 ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance())?.let { conn ->
                     if (conn is ManagedBearerSsoConnection) {
-                        ApplicationManager.getApplication().executeOnPooledThread {
+                        pluginAwareExecuteOnPooledThread {
```
rli, line 167: maybe worth adding a lint rule to disallow `executeOnPooledThread` and its variants

rli (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/QLoginWebview.kt
breedloj: Ack - this is a good idea. I'll follow up with this in a separate PR.

rbart:
looks good to me! 😁

PR #4617, opened 2024-06-26
Decrease failure noise by increase flaky test failure threshold

rli:
Sometimes we get exactly 5 failing flaky tests

## License


buildSrc/src/main/kotlin/toolkit-testing.gradle.kts
```diff
@@ -66,7 +66,7 @@ tasks.withType<Test>().all {
     ciOnly {
         retry {
             failOnPassedAfterRetry.set(false)
-            maxFailures.set(5)
+            maxFailures.set(50)
```
breedloj, line 69: If we aren't already, can we log these retries or emit them as a metric somewhere so we can keep track of offending tests? I'd hate for us to paper over issues with effectively unbounded retries which may also significantly degrade our CI times. At the same time, I understand this is better than failing the entire suite.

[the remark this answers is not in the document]
rli: it is still max 2 retries per test

rli: we can query them through cloudwatch insights
there's an existing dashboard that needs to be updated for the new release infrastructure

breedloj (approved):

PR #4623, opened 2024-06-27
Update codeowners file to requie Dev and Transformation review for Chat ui

manodnyab:
## License


.github/CODEOWNERS
```diff
@@ -7,4 +7,4 @@ codemodernizer/ @aws/elastic-gumby
 codetransform/ @aws/elastic-gumby
 codewhisperer/ @aws/codewhisperer-team
 
-mynah-ui/ @aws/dexp
```
rli, line 10: i think we wanted this reviewer change to be specific to the NPM package updates instead of everything in here as it'll cause friction with partners

rli (approved):

.github/CODEOWNERS
```diff
@@ -8,3 +8,5 @@ codetransform/ @aws/elastic-gumby
 codewhisperer/ @aws/codewhisperer-team
 
 mynah-ui/ @aws/dexp
+mynah-ui/package.json @aws/dexp @aws/elastic-gumby @aws/earlybird
```
rli, line 11: would be nice if we could do:
```suggestion
mynah-ui/{package,package-lock}.json @aws/dexp @aws/elastic-gumby @aws/earlybird
```

but doesn't seem like that's supported

PR #4632, opened 2024-06-28
Fix duplicated IdC failure metrics

wularr:
## Types of changes

## Description

Currently, attempting to sign in through the Q by clicking "Use with Pro License" then cancelling results in duplicate browser login and addConnection metrics. This is because we are submitting metrics in the IdC-specific login function and in the general login function (the IdC-specific login function calls the general one).

This change will move all metrics decisions to the IdC-specific login function. The general login function does the actual submission through a hook.

## Checklist
- [X] I have added metrics for my changes (if required)
 
## License

### DIFF
--- a/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
+++ b/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
@@ -115,17 +115,19 @@ abstract class LoginBrowser(
         else -> ""
     }
 
-    private fun isReAuth(scopes: List<String>): Boolean = if (scopes.toSet().intersect(Q_SCOPES.toSet()).isNotEmpty()) {
-        ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance()) != null
-    } else if (scopes.toSet().intersect(CODECATALYST_SCOPES.toSet()).isNotEmpty()) {
-        ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(CodeCatalystConnection.getInstance()) != null
-    } else {
-        ToolkitConnectionManager.getInstance(project).activeConnection().let {
-            val ccCon = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(CodeCatalystConnection.getInstance())
-            val qCon = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance())
-            it != null && it != ccCon && it != qCon
+    private fun isReAuth(scopes: List<String>): Boolean = ToolkitConnectionManager.getInstance(project)
+        .let {
+            if (scopes.toSet().intersect(Q_SCOPES.toSet()).isNotEmpty()) {
+                it.activeConnectionForFeature(QConnection.getInstance()) != null
+            } else if (scopes.toSet().intersect(CODECATALYST_SCOPES.toSet()).isNotEmpty()) {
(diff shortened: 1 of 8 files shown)


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
```diff
@@ -145,6 +146,7 @@ abstract class LoginBrowser(
     }
 
     open fun loginIdC(url: String, region: AwsRegion, scopes: List<String>) {
+        val isReAuth = ToolkitConnectionManager.getInstance(project).activeConnection() != null
```
manodnyab, line 149: Would this be a proper indicator of reauth, cause this is not feature-specific. So, could there be a scenario where it's authing for one feature but reauthing for another eg in the case of CC in GW and Q in IDE?

wularr: New revision inspects scopes to distinguish between Q and non-Q

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
```diff
@@ -145,6 +150,19 @@ abstract class LoginBrowser(
     }
 
     open fun loginIdC(url: String, region: AwsRegion, scopes: List<String>) {
+        // assumes scopes contains either Q or non-Q permissions but not both
+        val isReAuth = if (scopes.toSet().intersect(Q_SCOPES.toSet()).isNotEmpty()) {
+            ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance()) != null
```
manodnyab, line 155: Should we also compare the start url and region to the active connection's start url and region, cause if they are different then technically it's a new auth?

manodnyab: Can we also move the isReauth logic to a new function?

wularr: > Should we also compare the start url and region to the active connection's start url and region, cause if they are different then technically it's a new auth?

Gonna check in with analytics to see if this is considered new auth. There could be a world where the following can be considered a continuous string of reauths:

User relogs with acc A -> A credentials expire -> user relogs with acc B -> B credentials expire -> user relogs with acc C -> C credentials expire -> ...

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
```diff
@@ -110,6 +115,18 @@ abstract class LoginBrowser(
         else -> ""
     }
 
+    private fun isReAuth(scopes: List<String>): Boolean = if (scopes.toSet().intersect(Q_SCOPES.toSet()).isNotEmpty()) {
+        ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance()) != null
```
manodnyab, line 119: nit: can we initialize all the connections before so we can reuse the variables?

wularr: Pulled `ToolkitConnectionManager.getInstance(project)` to the top. Connections are branch dependent so initializing at the top is undesirable.

manodnyab (approved):

PR #4650, opened 2024-07-08
Add a metric on the continue button on login screen

manodnyab:
## License


rli (approved):

rli:
does the toolkit panel need changes too?

plugins/core/webview/src/q-ui/components/loginOptions.vue
```diff
@@ -2,8 +2,8 @@
 <!-- SPDX-License-Identifier: Apache-2.0 -->
 
 <template>
-    <QOptions v-if="app === 'AMAZONQ'" @stageChanged="stageChanged" @login="login"/>
-    <ToolkitOptions v-if="app === 'TOOLKIT'" @stageChanged="stageChanged" @login="login"/>
+    <QOptions v-if="app === 'AMAZONQ'" @stageChanged="stageChanged" @login="login" @sendTelemetry="sendTelemetry"/>
```
Will-ShaoHua, line 5: we could just reuse @login maybe?

Will-ShaoHua: https://github.com/aws/aws-toolkit-jetbrains/blob/main/plugins/core/webview/src/q-ui/components/login.vue#L83-L104

Will-ShaoHua: so that we don't need to add @sendTelemetry event in qLoginOptions.vue & toolkitLoginOptions.vue

Will-ShaoHua: btw i think you missed toolkitLoginOptions.vue

plugins/core/webview/src/q-ui/components/login.vue
```diff
@@ -81,6 +81,7 @@ export default defineComponent({
             this.mutateStage('START')
         },
         login(type: LoginOption) {
+            window.ideApi.postMessage({command: 'sendTelemetry'})
```
Will-ShaoHua, line 84: since message to java is sent from `@login` so you don't need to add @sendTelemetry actually

Will-ShaoHua: so we could revert `loginOptions.vue`, `qOptions.vue`, `toolkitOptions.vue`

manodnyab: yeah, I had a different command for sendTelemetry earlier

manodnyab:
> does the toolkit panel need changes too?

Yes, this affects both the panels

Will-ShaoHua (approved):

rli (approved):

PR #4678, opened 2024-07-16
Store JVM crash log artifacts in predictable location

rli:
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x000000009c900000, 667942912, 0) failed; error='The paging file is too small for this operation to complete' (DOS error/errno=1455)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 667942912 bytes for G1 virtual space
# An error report file with more information is saved as:
# C:\codebuild\tmp\output\src3225804393\src\github.com\aws\aws-toolkit-jetbrains\plugins\toolkit\jetbrains-core\hs_err_pid1020.log
```

Though we're essentially getting an OOM, it is manifesting as a JVM crash (OnError) rather than a OOM error (OnOutOfMemoryError)
 
## License


gradle.properties
```diff
@@ -18,7 +18,7 @@ kotlin.code.style=official
 # Gradle settings
 org.gradle.parallel=true
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx2048m
+org.gradle.jvmargs=-Xmx2048m -XX:HeapDumpPath="C:\Users\ContainerAdministrator\AppData\Local\Temp\testArtifacts" -XX:OnOutOfMemoryError="powershell.exe dump.ps1"
```
manodnyab, line 21: Will this path always be the same?

buildSrc/src/main/kotlin/toolkit-testing.gradle.kts
```diff
@@ -94,6 +94,11 @@ tasks.withType<Test>().all {
         // 221+ uses a custom classloader and jacoco fails to find classes
         isIncludeNoLocationClasses = true
     }
+
+    jvmArgs("-XX:HeapDumpPath=${rootProject.file("build/reports").absolutePath}")
+    if (System.getProperty("os.name").contains("indows")) {
```
manodnyab, line 99: nit: Windows

dump.ps1
```diff
@@ -0,0 +1 @@
+Get-WmiObject -Query 'Select * from Win32_PerfFormattedData_PerfProc_Process' | Select-Object -Property IDProcess,Name,PageFileBytes,PoolNonpagedBytes,PoolPagedBytes,PrivateBytes,VirtualBytes,WorkingSet,CmdLine | ConvertTo-Csv -NoTypeInformation; get-wmiobject -query 'select * From Win32_Process' | Select-Object -Property ProcessId,Name,CommandLine | ConvertTo-Csv -NoTypeInformation
```
manodnyab, line 1: Is this file required?

rli: we're trying to print process information and there is not a straightforward way of doing this in windows

manodnyab (approved):

PR #4694, opened 2024-07-18
Fix memory leak due to browser not being disposed

manodnyab:
Fix memory leak by attaching the browser to its parent disposable: in this case the amazon q tool window

 
## License


Will-ShaoHua (approved):

rli (approved):

manodnyab:
Failing tests were due to OOM

PR #4702, opened 2024-07-19
Add test for binary compatibility of methods consumed by internal plugin

rli:
As discussed internally when #4644 was raised, comments are only a good intention and we need a reliable mechanism to enforce this.

Since we have not annotated any of the internally dependent methods with `@JvmDefaultWithCompatibility` (and doing so is a breaking change), the Kotlin compiler is free to generate backwards-compatible bytecode that is resolvable at compile time, but then we get reports of issues at runtime since it is not binary compatible.
 
## License

### DIFF
--- a/plugins/amazonq/src/test/kotlin/PluginAmazonQJvmBinaryCompatabilityTest.kt
+++ b/plugins/amazonq/src/test/kotlin/PluginAmazonQJvmBinaryCompatabilityTest.kt
@@ -10,22 +10,21 @@ class PluginAmazonQJvmBinaryCompatabilityTest {
         // $ javap -c -classpath aws-toolkit-amazonq-2024.1.jar <...>.amazonq.AmazonQConnectionService
         //   public final boolean isAuthenticated();
         //       0: new           #27                 // class software/aws/toolkits/jetbrains/services/amazonq/auth/AuthController
-        //       3: dup
         //       4: invokespecial #28                 // Method software/aws/toolkits/jetbrains/services/amazonq/auth/AuthController."<init>":()V
-        //       7: aload_0
-        //       8: getfield      #21                 // Field project:Lcom/intellij/openapi/project/Project;
         //      11: invokevirtual #32                 // Method software/aws/toolkits/jetbrains/services/amazonq/auth/AuthController.getAuthNeededStates:(Lcom/intellij/openapi/project/Project;)Lsoftware/aws/toolkits/jetbrains/services/amazonq/auth/AuthNeededStates;
-        //      14: astore_1
-        //      15: aload_1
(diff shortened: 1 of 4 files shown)


plugins/amazonq/src/test/kotlin/PluginAmazonQJvmBinaryCompatabilityTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        // type erasure :/
+        assertThat(
+            modelConfigurator.getMethod(
+                "listCustomizations\$default",
+                modelConfigurator,
+                Class.forName("com.intellij.openapi.project.Project"),
+                // can't request primitive type using reflection
+                java.lang.Boolean.TYPE,
+                Integer.TYPE,
+                Class.forName("java.lang.Object")
```
heyrichardshi, line 60: Where are these parameter types coming from? [The method signature](https://github.com/aws/aws-toolkit-jetbrains/blob/dd25f96bb575e24ed71e8d826718ac8a88d68723/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/customization/CodeWhispererModelConfigurator.kt#L89) only shows `Project` and `Boolean`.

plugins/core/src/test/kotlin/software/aws/toolkits/jetbrains/core/PluginCoreJvmBinaryCompatabilityTest.kt
```diff
@@ -0,0 +1,62 @@
...
+            Class.forName("java.lang.String"),
+            Class.forName("java.lang.String"),
+            Class.forName("java.util.List"),
+            Class.forName("kotlin.jvm.functions.Function1"),
+            Class.forName("kotlin.jvm.functions.Function1"),
+            Class.forName("kotlin.jvm.functions.Function0"),
+            Class.forName("software.aws.toolkits.jetbrains.core.credentials.ConnectionMetadata"),
+            // can't request primitive type using reflection
+            Integer.TYPE,
+            Class.forName("java.lang.Object"),
```
heyrichardshi, line 41: Where are these parameter types coming from? [The method signature](https://github.com/aws/aws-toolkit-jetbrains/blob/1d081527aee07dc18b9ad22ab97db6653b64be9d/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt#L112-L121) ends with `ConnectionMetadata`.

rli: `*$default` is a synthetic method generated by the Kotlin compiler. Based on the bytecode extracted from the internal plugin jar, this is what is being invoked.

heyrichardshi (approved):
These tests cover all calls from the internal plugin.

manodnyab (approved):

PR #4760, opened 2024-08-02
Require use of trailing commas at declaration site

rli:
https://kotlinlang.org/docs/coding-conventions.html#trailing-commas
> The Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it at your discretion for the call site.

Requiring trailing commas at call site would be nice as well, but will revisit in the future

## License

### DIFF
--- a/plugins/toolkit/jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
+++ b/plugins/toolkit/jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
@@ -388,7 +388,7 @@ class CawsConnectionProvider : GatewayConnectionProvider {
 
     private fun validateEnvironmentIsRunning(
         indicator: ProgressIndicator,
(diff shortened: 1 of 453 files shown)


plugins/toolkit/jetbrains-gateway/src/software/aws/toolkits/jetbrains/gateway/CawsConnectionProvider.kt
```diff
@@ -436,7 +436,8 @@ class CawsConnectionProvider : GatewayConnectionProvider {
                 }
 
                 is GitSettings.CawsOwnedRepoSettings,
-                is GitSettings.NoRepo -> {
+                is GitSettings.NoRepo,
```
manodnyab, line 439: is this the correct syntax, do we want the comma here?

rli: yes it is multi line so it makes sense

manodnyab (approved):

PR #4766, opened 2024-08-05
fix(amazonq): set Q_WORKER_THREADS only when setting is >0

grant0417:
## Types of changes

## Description
This change only sets the `Q_WORKER_THREADS` for the amazonq lsp when the value is >0 and <100. This should make it have the same logic as the VSCode plugin: https://github.com/aws/aws-toolkit-vscode/blob/4b3804fd6c5b4106dc7178cb46121d95adf7ff0d/packages/core/src/amazonq/lsp/lspClient.ts#L141C1-L143C13 

This caused the default number of threads used by the lps to be incorrect as setting the value of `Q_WORKER_THREADS=0` caused the runtime to use a different heuristic to determine thread count than when it is unset.

## Checklist
 
## License

### DIFF
--- a/.changes/next-release/bugfix-a5c03c3c-3d5c-454b-8d52-1d125ca0a5cc.json
+++ b/.changes/next-release/bugfix-a5c03c3c-3d5c-454b-8d52-1d125ca0a5cc.json
@@ -1,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "fix(amazonq): Amazon Q chat `@workspace` uses more than 20% cpu"
+}
--- a/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/cwc/editor/context/project/EncoderServer.kt
+++ b/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/cwc/editor/context/project/EncoderServer.kt
@@ -118,10 +118,12 @@ class EncoderServer(val project: Project) : Disposable {
         val map = mutableMapOf<String, String>(
             "PORT" to port.toString(),
             "START_AMAZONQ_LSP" to "true",
-            "Q_WORKER_THREADS" to threadCount.toString(),
             "CACHE_DIR" to cachePath.toString(),
             "MODEL_DIR" to cachePath.resolve("qserver").toString()
         )
+        if (threadCount > 0) {
+            map["Q_WORKER_THREADS"] = threadCount.toString()
+        }
         if (isGpuEnabled) {
             map["Q_ENABLE_GPU"] = "true"
         }


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/cwc/editor/context/project/EncoderServer.kt
```diff
@@ -118,10 +118,12 @@ class EncoderServer(val project: Project) : Disposable {
         val map = mutableMapOf<String, String>(
             "PORT" to port.toString(),
             "START_AMAZONQ_LSP" to "true",
-            "Q_WORKER_THREADS" to threadCount.toString(),
             "CACHE_DIR" to cachePath.toString(),
             "MODEL_DIR" to cachePath.resolve("qserver").toString()
         )
+        if (threadCount > 0 && threadCount < 100) {
```
rli, line 124: it looks like we already have input validation in the settings pane and i would rather have the constraints be defined in one place
```suggestion
        if (threadCount > 0) {
```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/cwc/editor/context/project/EncoderServer.kt
```diff
@@ -118,10 +118,12 @@ class EncoderServer(val project: Project) : Disposable {
...
+            map["Q_WORKER_THREADS"] = threadCount.toString()
+        }
         if (isGpuEnabled) {
```
rli, line 127: unrelated, but why doesn't `Q_ENABLE_GPU = false` work?

rli (approved):

zixlin7 (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/cwc/editor/context/project/EncoderServer.kt
grant0417: The lsp does not handle strings well, we are going to improve support on that side but at the moment this is the best fix we think.

.changes/next-release/bugfix-a5c03c3c-3d5c-454b-8d52-1d125ca0a5cc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "fix(amazonq): set Q_WORKER_THREADS only when setting is >0"
```
zixlin7, line 3: should we change this to "Amazon Q chat `@workspace` uses more than 20% cpu?

grant0417: Fixed

PR #4785, opened 2024-08-12
Drop support for IDE 232 / Gateway 241

rli:
## License


buildSrc/src/main/kotlin/software/aws/toolkits/gradle/intellij/IdeVersions.kt
```diff
@@ -18,43 +18,6 @@ object IdeVersions {
     )
 
     private val ideProfiles = listOf(
-        Profile(
```
manodnyab, line 21: we also need to remove GW v241 from here!

manodnyab (approved):
We also need to delete the run config for GW 241

PR #4790, opened 2024-08-13
fix(amazonq): fix UI bugs by bumping mynah-ui version

ege0zcan:
## Problem
- Tab selection scrollbar visibility which causes tabs half visible if there are several tabs open.
- Header items in card bodies don't wrap if they don't contain spaces.
- Related link previews sometimes remain on screen and block the whole Chat UI.
- `@` context selector conflicts with some use cases where the user wants use `@` character for a word in the prompt itself.
- `/` command selector doesn't work if user pastes the command to prompt and submits

### Header item wrapping
image

### `/` Command selector
https://github.com/user-attachments/assets/b5ed474d-03d9-4eff-afed-4d0d9e9365a8

### `@` Context selector
https://github.com/user-attachments/assets/68ed9202-7110-4c9f-bc28-8b988ab9182e
## Solution
Updating @aws/mynah-ui to 4.15.6

## License

### DIFF
--- a/.changes/next-release/bugfix-687fa4bc-8244-4b64-873f-c82a0baeec67.json
+++ b/.changes/next-release/bugfix-687fa4bc-8244-4b64-873f-c82a0baeec67.json
@@ -1,0 +1,4 @@
+{
+  "type" : "bugfix",
+  "description" : "Amazon Q Chat: Fix Tab selection scrollbar visibility which causes tabs half visible if there are several tabs open"
+}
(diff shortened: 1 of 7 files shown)


willyyhuang (approved):

willyyhuang:
Remove the diff to update mynah ui dependency in https://github.com/aws/aws-toolkit-jetbrains/pull/4777, will rely on this pull request to bump mynah version.

PR #4819, opened 2024-08-20
Add flag to indicate reauth in metrics

manodnyab:
## Description

Add isReauth flag to indicate reauth in Auth_addConnection and loginWithBrowser metrics

 
## License


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -243,15 +232,46 @@ fun reauthConnectionIfNeeded(
     project: Project?,
     connection: ToolkitConnection,
     onPendingToken: (InteractiveBearerTokenProvider) -> Unit = {},
+    isReAuth: Boolean = false
```
rli, line 235: flip the condition because it's only false for the initial login case

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -26,6 +26,7 @@ import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.BearerTokenPr
 import software.aws.toolkits.jetbrains.core.credentials.sso.bearer.InteractiveBearerTokenProvider
 import software.aws.toolkits.jetbrains.utils.runUnderProgressIfNeeded
 import software.aws.toolkits.resources.AwsCoreBundle
+import software.aws.toolkits.resources.AwsCoreBundle.message
```
rli, line 29:
```suggestion

```

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -243,15 +232,46 @@ fun reauthConnectionIfNeeded(
...
+                        credentialStartUrl = tokenProvider.currentToken()?.ssoUrl,
+                        credentialSourceId = CredentialSourceId.AwsId,
+                        isReAuth = true,
+                        result = Result.Succeeded
+                    )
+                    recordAddConnection(
+                        credentialSourceId = CredentialSourceId.AwsId,
+                        isReAuth = true,
+                        result = Result.Succeeded
+                    )
```
rli, line 257: i still dont understand why these are separate metrics

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -243,15 +232,46 @@ fun reauthConnectionIfNeeded(
...
+                }
+            } catch (e: Exception) {
+                if (isReAuth) {
+                    recordLoginWithBrowser(
+                        credentialStartUrl = tokenProvider.currentToken()?.ssoUrl,
+                        credentialSourceId = CredentialSourceId.AwsId,
```
rli, line 263: source id is based on the start url

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -243,15 +232,46 @@ fun reauthConnectionIfNeeded(
...
+                    recordAddConnection(
+                        credentialSourceId = CredentialSourceId.AwsId,
+                        isReAuth = true,
+                        result = Result.Succeeded
+                    )
+                }
+            } catch (e: Exception) {
+                if (isReAuth) {
+                    recordLoginWithBrowser(
+                        credentialStartUrl = tokenProvider.currentToken()?.ssoUrl,
```
rli, line 262: but if we're passing along the start url, why do we need to also determine the source id?

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/explorerActions/SignInToQAction.kt
```diff
@@ -39,7 +39,7 @@ abstract class SignInToQActionBase(actionName: String) : DumbAwareAction(actionN
         UiTelemetry.click(project, "auth_start_Q")
         val connectionManager = ToolkitConnectionManager.getInstance(project)
         connectionManager.activeConnectionForFeature(QConnection.getInstance())?.let {
-            reauthConnectionIfNeeded(project, it)
+            reauthConnectionIfNeeded(project, it, isReAuth = true)
```
rli, line 42: is this not an initial login?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
manodnyab: Changing the source ID

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
manodnyab: addConnection is to be used in other places too where the call fails (eg: prior to the browser step)

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
manodnyab: startUrl is required for locating internal users

rli: im saying that the source id is an implicit property of the start url, so why can't the BI team make this transform for us

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
```diff
@@ -120,6 +120,7 @@ fun loginSso(
     onSuccess: () -> Unit = {},
     metadata: ConnectionMetadata? = null
 ): AwsBearerTokenConnection? {
+    val source = metadata
```
manodnyab, line 123: This is just to make detekt happy, will look at implementing this in a follow-up PR

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
manodnyab: discussed offline, will continue to keep this as false

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/explorerActions/SignInToQAction.kt
manodnyab: This checks for an existing Q connection: https://github.com/aws/aws-toolkit-jetbrains/pull/4819/files#diff-899f1e03be46308c47e08c0407c1a1711e9e94c5b9afd38b649231041079bd44R41 and if not, it starts with a new login, hence marked this as reauth

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
manodnyab: oh the other way around, got it. Yeah, doesn't add much value, but will let @dhruvigajjar comment

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/explorerActions/SignInToQAction.kt
dhruvigajjar: no we would not consider exiting creds option as reauth. only when creds expire or user is clicking "reconnect". for existing we can emit a ui_click that would be enough to filter. because if user had sso on AWS toolkit, ideally this is their first login on Amazon Q.

dhruvigajjar: Think of `reauth` as flag that will be used to check if users are logging in again on same compute/same product

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
dhruvigajjar: no we dont need `credentialSourceId` if we will always have startUrl.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
dhruvigajjar: AddConnection covers entire flow. 

- Moment users clicks "add new Connection" we record this metric, meaning if they give up only looking at auth options (idc, bId or iam), we emit auth_addConnection with result:Cancelled. 

- if they have chosen one of the auth options and leave when they invalid starturl, profile name,etc we emit auth_addConnection {invalidInputFields}. 

- After all this when user enters auth flow we emit loginWithBrowser to capture nuances of browser flow. we also miss users loggin with IAM creds in loginWithBrowser so auth_addConnection helps us give full picture. 

There are some scenarios where user dont go through normal auth flow and skips to brower, such as Black Caimen users, reauth in this case we rely on loginWithBrowser to complete auth picture. 

more details on design: https://quip-amazon.com/M2VdAWUanrLv/New-Auth-telemetry-design

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ToolkitAuthManager.kt
dhruvigajjar: we should only have isReAuth:true if user is re-logging/auth is expired. existing connection should be counted as first login only.

manodnyab: Yes, this is an implementation detail.

rli (approved):

andrewyuq (approved):

PR #4896, opened 2024-09-19
Fix read access without read lock while building supplemental context

rli:
`isTestFile` needs a read lock, but lost it with commit merged in #4765 
 
## License


Will-ShaoHua (approved):

samgst-amazon (approved):

rli:
will try to follow up with a test if possible

PR #4899, opened 2024-09-20
Add 243 build profile

rli:
## License


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/lambda/wizard/SdkSelectionPanel.kt
rli, line 1: there are lots of EDT warnings in 243 but i don't think its worth the time to investigate at right now

gradle/libs.versions.toml
```diff
@@ -5,12 +5,12 @@ assertJ = "3.26.3"
...
 commonmark = "0.22.0"
-detekt = "1.23.6"
+detekt = "1.23.7"
 intellijExt = "1.1.8"
 # match with <root>/settings.gradle.kts
-intellijGradle = "2.0.0"
+intellijGradle = "2.1.0"
 intellijRemoteRobot = "0.11.22"
-jackson = "2.16.1"
+jackson = "2.17.2"
```
manodnyab, line 13: We had reverted this to 2.16.1 in a previous version: https://github.com/aws/aws-toolkit-jetbrains/pull/4732/files because it broke chat, just confirming if we know if that's been fixed?

manodnyab (commented):
Can you please add changelog, and update the readme.md and the GH workflow?

gradle/libs.versions.toml
rli: yes we now remove the conflicting jackson jars from our distribution

manodnyab (approved):

PR #4911, opened 2024-09-24
Remove log spam from CodeCatalystConnection

rli:
`supportsConnectionType` is called very frequently
 
## License


samgst-amazon (approved):
I was seeing this a lot the other day. Is there a reason why CodeCatalystConnection gets checked so much?

PR #4916, opened 2024-09-25
Track SessionContext instead of InvocationContext

andrewyuq:
1. Since the shift from displaying one trigger at a time to displaying multiple triggers at a time, use SessionContext as the disposable for the session states.
2. Introduce the concept of "display session" which can show multiple triggers at a time and sessionContext is to hold states for that session.
3. ongoingRequests will be a map of <jobId, InvocationContext> where InvocationContext is the state of each trigger.
4. Managed these states globally by dynamically adding Trigger states into ongoingRequests and whenever the user made a decision(accept or reject), conclude decisions for all the ongoing triggers and clear ongoingRequest, so that a new display session can start later.

## Types of changes

## Checklist
 
## License

### DIFF
--- a/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
+++ b/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
@@ -17,16 +17,12 @@ import software.amazon.awssdk.services.codewhispererruntime.model.GenerateComple
 import software.aws.toolkits.jetbrains.core.credentials.ToolkitConnection
 import software.aws.toolkits.jetbrains.services.codewhisperer.codescan.sessionconfig.PayloadContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.language.CodeWhispererProgrammingLanguage
-import software.aws.toolkits.jetbrains.services.codewhisperer.popup.CodeWhispererPopupManager
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererAutomatedTriggerType
-import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererIntelliSenseOnHoverListener
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererInvocationStatus
-import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererService
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.RequestContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.ResponseContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.telemetry.CodeWhispererTelemetryService
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererConstants
-import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererUtil.setIntelliSensePopupAlpha
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CrossFileStrategy
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.SupplementalContextStrategy
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.UtgStrategy
@@ -129,7 +125,7 @@ data class SessionContext(
     var isFirstTimeShowingPopup: Boolean = true,
     var toBeRemovedHighlighter: RangeHighlighter? = null,
     var insertEndOffset: Int = -1,
-    var popupDisplayOffset: Int = -1,
+    var popupOffset: Int = -1,
     val latencyContext: LatencyContext,
     var hasAccepted: Boolean = false
 ) : Disposable {
--- a/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
+++ b/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
@@ -4,25 +4,37 @@
 package software.aws.toolkits.jetbrains.services.codewhisperer.model
 
 import com.intellij.openapi.Disposable
+import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.VisualPosition
 import com.intellij.openapi.editor.markup.RangeHighlighter
+import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.popup.JBPopup
+import com.intellij.openapi.util.Disposer
 import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.util.concurrency.annotations.RequiresEdt
 import software.amazon.awssdk.services.codewhispererruntime.model.Completion
 import software.amazon.awssdk.services.codewhispererruntime.model.GenerateCompletionsResponse
 import software.aws.toolkits.jetbrains.core.credentials.ToolkitConnection
 import software.aws.toolkits.jetbrains.services.codewhisperer.codescan.sessionconfig.PayloadContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.language.CodeWhispererProgrammingLanguage
+import software.aws.toolkits.jetbrains.services.codewhisperer.popup.CodeWhispererPopupManager
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererAutomatedTriggerType
+import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererIntelliSenseOnHoverListener
+import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererInvocationStatus
+import software.aws.toolkits.jetbrains.services.codewhisperer.service.CodeWhispererService
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.RequestContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.service.ResponseContext
+import software.aws.toolkits.jetbrains.services.codewhisperer.telemetry.CodeWhispererTelemetryService
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererConstants
+import software.aws.toolkits.jetbrains.services.codewhisperer.util.CodeWhispererUtil.setIntelliSensePopupAlpha
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.CrossFileStrategy
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.SupplementalContextStrategy
 import software.aws.toolkits.jetbrains.services.codewhisperer.util.UtgStrategy
 import software.aws.toolkits.telemetry.CodewhispererCompletionType
 import software.aws.toolkits.telemetry.CodewhispererTriggerType
 import software.aws.toolkits.telemetry.Result
+import java.time.Duration
+import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 data class Chunk(
@@ -81,10 +93,19 @@ data class SupplementalContextInfo(
 }
 
 data class RecommendationContext(
-    val details: List<DetailContext>,
+    val details: MutableList<DetailContext>,
     val userInputOriginal: String,
     val userInputSinceInvocation: String,
-    val position: VisualPosition
+    val position: VisualPosition,
+    val jobId: Int,
+    var typeahead: String = "",
+)
+
+data class PreviewContext(
+    val jobId: Int,
+    val detail: DetailContext,
+    val userInput: String,
+    val typeahead: String,
 )
 
 data class DetailContext(
@@ -95,17 +116,47 @@ data class DetailContext(
     val isTruncatedOnRight: Boolean,
     val rightOverlap: String = "",
     val completionType: CodewhispererCompletionType,
+    var hasSeen: Boolean = false,
+    var isAccepted: Boolean = false
 )
 
 data class SessionContext(
-    val typeahead: String = "",
-    val typeaheadOriginal: String = "",
-    val selectedIndex: Int = 0,
+    val project: Project,
+    val editor: Editor,
+    var popup: JBPopup? = null,
+    var selectedIndex: Int = -1,
     val seen: MutableSet<Int> = mutableSetOf(),
-    val isFirstTimeShowingPopup: Boolean = true,
+    var isFirstTimeShowingPopup: Boolean = true,
     var toBeRemovedHighlighter: RangeHighlighter? = null,
-    var insertEndOffset: Int = -1
-)
+    var insertEndOffset: Int = -1,
+    var popupDisplayOffset: Int = -1,
+    val latencyContext: LatencyContext,
+    var hasAccepted: Boolean = false
+) : Disposable {
+    private var isDisposed = false
+
+    @RequiresEdt
+    override fun dispose() {
+        CodeWhispererTelemetryService.getInstance().sendUserDecisionEventForAll(
+            this,
+            hasAccepted,
+            CodeWhispererInvocationStatus.getInstance().popupStartTimestamp?.let { Duration.between(it, Instant.now()) }
+        )
+        CodeWhispererInvocationStatus.getInstance().setDisplaySessionActive(false)
+
+        if (hasAccepted) {
+            popup?.closeOk(null)
+        } else {
+            popup?.cancel()
+        }
+        popup?.let { Disposer.dispose(it) }
+        popup = null
+        CodeWhispererInvocationStatus.getInstance().finishInvocation()
+        isDisposed = true
+    }
+
+    fun isDisposed() = isDisposed
+}
 
 data class RecommendationChunk(
     val text: String,
@@ -124,16 +175,21 @@ data class InvocationContext(
     val requestContext: RequestContext,
     val responseContext: ResponseContext,
     val recommendationContext: RecommendationContext,
-    val popup: JBPopup
 ) : Disposable {
-    override fun dispose() {}
+    private var isDisposed = false
+
+    @RequiresEdt
+    override fun dispose() {
+        isDisposed = true
+    }
+
+    fun isDisposed() = isDisposed
 }
 
 data class WorkerContext(
     val requestContext: RequestContext,
     val responseContext: ResponseContext,
     val response: GenerateCompletionsResponse,
-    val popup: JBPopup
 )
 
 data class CodeScanTelemetryEvent(
--- a/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
+++ b/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
@@ -27,6 +27,8 @@ import com.intellij.openapi.editor.event.CaretEvent
 import com.intellij.openapi.editor.event.CaretListener
 import com.intellij.openapi.editor.event.DocumentEvent
 import com.intellij.openapi.editor.event.DocumentListener
+import com.intellij.openapi.editor.event.EditorMouseEvent
+import com.intellij.openapi.editor.event.EditorMouseMotionListener
 import com.intellij.openapi.editor.event.SelectionEvent
 import com.intellij.openapi.editor.event.SelectionListener
 import com.intellij.openapi.fileEditor.FileEditorManager
@@ -39,6 +41,7 @@ import com.intellij.ui.ComponentUtil
 import com.intellij.ui.awt.RelativePoint
 import com.intellij.ui.popup.AbstractPopup
 import com.intellij.ui.popup.PopupFactoryImpl
+import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.messages.Topic
 import com.intellij.util.ui.UIUtil
 import software.amazon.awssdk.services.codewhispererruntime.model.Import
@@ -51,6 +54,7 @@ import software.aws.toolkits.jetbrains.services.codewhisperer.layout.CodeWhisper
 import software.aws.toolkits.jetbrains.services.codewhisperer.layout.CodeWhispererLayoutConfig.inlineLabelConstraints
 import software.aws.toolkits.jetbrains.services.codewhisperer.model.DetailContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.model.InvocationContext
+import software.aws.toolkits.jetbrains.services.codewhisperer.model.PreviewContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.model.SessionContext
 import software.aws.toolkits.jetbrains.services.codewhisperer.popup.handlers.CodeWhispererEditorActionHandler
 import software.aws.toolkits.jetbrains.services.codewhisperer.popup.handlers.CodeWhispererPopupBackspaceHandler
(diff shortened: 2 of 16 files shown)


plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -94,6 +96,10 @@ import java.util.concurrent.TimeUnit
 class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
     private val codeInsightSettingsFacade = CodeInsightsSettingsFacade()
     private var refreshFailure: Int = 0
+    private val ongoingRequests = mutableMapOf<Int, InvocationContext?>()
+    val ongoingRequestsContext = mutableMapOf<Int, RequestContext>()
```
rli, line 100: not sure if this is the best pattern if the exact jobId is only relevant for "which is the last request", which we can determine in other ways

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -161,18 +167,27 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
...
         val requestContext = try {
-            getRequestContext(triggerTypeInfo, editor, project, psiFile, latencyContext)
+            getRequestContext(triggerTypeInfo, editor, project, psiFile)
         } catch (e: Exception) {
             LOG.debug { e.message.toString() }
             CodeWhispererTelemetryService.getInstance().sendFailedServiceInvocationEvent(project, e::class.simpleName)
             return
         }
+        val caretContext = requestContext.fileContextInfo.caretContext
+        ongoingRequestsContext.forEach { (k, v) ->
```
rli, line 179: for example, this is guaranteed to iterate in-order from first request inserted in the map to last

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -183,34 +198,38 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
...
+        }
+        this.sessionContext = sessionContext
 
         val workerContexts = mutableListOf<WorkerContext>()
         // When popup is disposed we will cancel this coroutine. The only places popup can get disposed should be
         // from CodeWhispererPopupManager.cancelPopup() and CodeWhispererPopupManager.closePopup().
         // It's possible and ok that coroutine will keep running until the next time we check it's state.
         // As long as we don't show to the user extra info we are good.
-        val coroutineScope = disposableCoroutineScope(popup)
+        val coroutineScope = projectCoroutineScope(requestContext.project)
```
rli, line 231: `CodeWhispererService` has a service-level scope that should be used instead

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
```diff
@@ -81,10 +93,19 @@ data class SupplementalContextInfo(
 }
 
 data class RecommendationContext(
-    val details: List<DetailContext>,
+    val details: MutableList<DetailContext>,
```
rli, line 96: not seeing why need mutable

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
```diff
@@ -95,17 +116,47 @@ data class DetailContext(
...
+        )
+        CodeWhispererInvocationStatus.getInstance().setDisplaySessionActive(false)
+
+        if (hasAccepted) {
+            popup?.closeOk(null)
+        } else {
+            popup?.cancel()
+        }
+        popup?.let { Disposer.dispose(it) }
+        popup = null
```
rli, line 153: popup manager and sessioncontext are so tied together that it feels like it should just be handled together

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -257,84 +244,36 @@ class CodeWhispererPopupManager {
...
-        popup.closeOk(null)
-        Disposer.dispose(popup)
-    }
-
-    fun closePopup() {
-        myPopup?.let {
-            it.closeOk(null)
-            Disposer.dispose(it)
+    fun showPopup(sessionContext: SessionContext, force: Boolean = false) {
+        val p = sessionContext.editor.offsetToXY(sessionContext.popupDisplayOffset)
```
rli, line 248: `popupOffset`?

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
```diff
@@ -622,60 +561,45 @@ class CodeWhispererPopupManager {
...
-        userInput: String,
-        selectedIndex: Int,
-        typeahead: String
-    ): Int {
-        var currIndexIgnoreInvalid = 0
-        detailContexts.forEachIndexed { index, value ->
+    private fun getValidCount(): Int =
+        CodeWhispererService.getInstance().getAllSuggestionsPreviewInfo().filter { isValidRecommendation(it) }.size
+
+    private fun getValidSelectedIndex(selectedIndex: Int): Int {
```
rli, line 585: feels like this could be cleaner but i see that it's the same as the old logic

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -94,6 +96,10 @@ import java.util.concurrent.TimeUnit
 class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
     private val codeInsightSettingsFacade = CodeInsightsSettingsFacade()
     private var refreshFailure: Int = 0
+    private val ongoingRequests = mutableMapOf<Int, InvocationContext?>()
+    val ongoingRequestsContext = mutableMapOf<Int, RequestContext>()
+    private var jobId = 0
```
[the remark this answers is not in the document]
andrewyuq: We have a monthly trigger limit which is 20000, so jobId will increment 20000 each month if they don't restart IDE.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: jobId is only to separate different triggers and is a local identifier only, (same as sessionId except we know the jobId as soon as a trigger happens before receiving the response)

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: yea, the purpose is not just to identify which the last request is -- but more to separate states of different triggers.

Later on when we send UserTriggerDecisions we will need to send one for each jobId(trigger), having a jobId here and logging it keeps track of what happened E2E for this specific trigger.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/CodeWhispererPopupManager.kt
andrewyuq: changed

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
andrewyuq: popup now becomes a state of a session (1 session 1 popup) so I put it as a field of `SessionContext`, and then I let disposing SessionContext handles everything so I don't have to call 
```
// CWService.getInstance().disposeDisplaySession()
// popup.cancel()/close()
```
separately.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
andrewyuq: Now each trigger will only have one state `InvocationContext` and is stored in `ongoingRequests`, when subsequent suggestions sent back to client, they will be added to this mutable details, and other parts of `InvocationContext` remains the same.

```
private fun updateStates(
        states: InvocationContext,
        response: GenerateCompletionsResponse
    ): InvocationContext {
        val recommendationContext = states.recommendationContext
        val newDetailContexts = CodeWhispererRecommendationManager.getInstance().buildDetailContext(
            states.requestContext,
            recommendationContext.userInputSinceInvocation,
            response.completions(),
            response.responseMetadata().requestId()
        )

        recommendationContext.details.addAll(newDetailContexts)
        return states
    }
```

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: fixed

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
```diff
@@ -81,10 +89,19 @@ data class SupplementalContextInfo(
...
 
 data class RecommendationContext(
-    val details: List<DetailContext>,
+    val details: MutableList<DetailContext>,
     val userInputOriginal: String,
     val userInputSinceInvocation: String,
-    val position: VisualPosition
+    val position: VisualPosition,
+    val jobId: Int,
+    var typeahead: String = "",
```
Will-ShaoHua, line 97: why we need two `typeahead`

Will-ShaoHua:
can we add some doc strings for different models

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
```diff
@@ -81,10 +89,19 @@ data class SupplementalContextInfo(
...
+)
+
+data class PreviewContext(
+    val jobId: Int,
+    val detail: DetailContext,
+    val userInput: String,
+    val typeahead: String,
```
Will-ShaoHua, line 104: dumb q, what's the difference between these 2 typeaheads

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -88,12 +89,17 @@ import software.aws.toolkits.resources.message
...
 import java.util.concurrent.TimeUnit
 
 @Service
 class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
     private val codeInsightSettingsFacade = CodeInsightsSettingsFacade()
     private var refreshFailure: Int = 0
+    private val ongoingRequests = mutableMapOf<Int, InvocationContext?>()
+    val ongoingRequestsContext = mutableMapOf<Int, RequestContext>()
+    private var jobId = 0
+    private var sessionContext: SessionContext? = null
```
Will-ShaoHua, line 102: not sure do we need this to be thread safe

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -88,12 +89,17 @@ import software.aws.toolkits.resources.message
...
 import software.aws.toolkits.telemetry.CodewhispererSuggestionState
 import software.aws.toolkits.telemetry.CodewhispererTriggerType
+import java.util.concurrent.CancellationException
 import java.util.concurrent.TimeUnit
 
 @Service
 class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
     private val codeInsightSettingsFacade = CodeInsightsSettingsFacade()
     private var refreshFailure: Int = 0
+    private val ongoingRequests = mutableMapOf<Int, InvocationContext?>()
```
Will-ShaoHua, line 99: what's the case when InvocationContext being null

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -285,22 +300,42 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
...
+                                            processCodeWhispererUI(
+                                                sessionContext,
+                                                it,
+                                                ongoingRequests[currentJobId],
+                                                cs,
+                                                currentJobId
+                                            )
+                                            if (!ongoingRequests.contains(currentJobId)) {
+                                                cs.coroutineContext.cancel()
+                                            }
```
Will-ShaoHua, line 312: andrew could you help me understand why isn't this in the format of
```
if (ongoingRequests.contains(currentJobId)) {
     processCodeWhispererUi()
}
```
but execute processCodeWhispererUi and cancel the coroutine later, what's the expected behavior here?

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -285,22 +300,42 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
...
                             } else {
                                 // Have enough delay before showing for the first response, or it's subsequent responses
-                                states = processCodeWhispererUI(workerContext, states)
+                                processCodeWhispererUI(
+                                    sessionContext,
+                                    workerContext,
+                                    ongoingRequests[currentJobId],
+                                    cs,
+                                    currentJobId
+                                )
```
Will-ShaoHua, line 326: why here doesn't need the 
```
if (!ongoingRequests.contains(currentJobId)) {
                                                cs.coroutineContext.cancel()
                                            }
```

check

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -321,6 +356,7 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
                     val responseContext = ResponseContext(sessionId)
 
                     CodeWhispererTelemetryService.getInstance().sendServiceInvocationEvent(
+                        currentJobId,
```
Will-ShaoHua, line 359: let's say there're 2 sessions with 4 triggers and 3 triggers respectively, are we sending something like jobid = [1, 2, 3, 4] for the first session and [5, 6, 7] the 2nd session?

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -350,23 +386,22 @@ class CodeWhispererService(private val cs: CoroutineScope) : Disposable {
...
+                            latencyContext
                         )
                     }
                     return@launch
                 } else if (e is CodeWhispererException) {
                     requestId = e.requestId() ?: ""
                     sessionId = e.awsErrorDetails().sdkHttpResponse().headers().getOrDefault(KET_SESSION_ID, listOf(requestId))[0]
                     displayMessage = e.awsErrorDetails().errorMessage() ?: message("codewhisperer.trigger.error.server_side")
-                } else if (e is software.amazon.awssdk.services.codewhispererruntime.model.CodeWhispererRuntimeException) {
+                } else if (e is CodeWhispererRuntimeException) {
```
Will-ShaoHua, line 404: thanks for the change!

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
```diff
@@ -15,16 +16,17 @@ import com.intellij.openapi.components.service
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.VisualPosition
 import com.intellij.openapi.project.Project
-import com.intellij.openapi.ui.popup.JBPopup
 import com.intellij.openapi.util.Disposer
 import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiFile
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.messages.Topic
+import io.ktor.utils.io.CancellationException
```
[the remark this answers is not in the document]
andrewyuq: fixed.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
andrewyuq: previously we have typeahead and typeaheadoriginal for the purpose of accounting leading spaces in the typeahead, 
for example, typeaheadoriginal will be `     System` and typeahead will be `System`, and then we use typeahead to do the suggestion matching logic. This is to preserve the suggestions as much as possible.

Now we trigger more so we have more display oppotunities, I feel like this becomes a bit unnecessary and makes the logic complicated, so remove typeahead original to make it no longer accomodate leading spaces.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererInvocationStatus.kt
```diff
@@ -69,9 +59,9 @@ class CodeWhispererInvocationStatus {
         return timeCanShowCodeWhisperer.isBefore(Instant.now())
     }
 
-    fun isPopupActive(): Boolean = isPopupActive
+    fun isDisplaySessionActive(): Boolean = isPopupActive
```
Will-ShaoHua, line 62: is it different from "Session.isDisposed()"?

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: Technically it needs, so the current rule is to modify this only in EDT so I believe in various places I have `@RequiresEDT` annotation including `disposeDisplaySession`, and initialing SessionContext also is in EDT.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererInvocationStatus.kt
Will-ShaoHua: asking because i feel they kinda serve similar purpose

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: yes, jobId only increments. It's a local field that only helps logging and not sent to telemetry.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererInvocationStatus.kt
andrewyuq: It's still different because it's a valid case that a session is active but displaySession is not active. 
We start the session as soon as a trigger happens to track the current ongoing requests, but all suggestions can be discarded so a displaySession can never be active but we still need to dispose the current valid session.

Will-ShaoHua: thanks for clarifying it

Will-ShaoHua (approved):

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/model/CodeWhispererModel.kt
```diff
@@ -95,17 +112,47 @@ data class DetailContext(
...
-    val isFirstTimeShowingPopup: Boolean = true,
+    var isFirstTimeShowingPopup: Boolean = true,
     var toBeRemovedHighlighter: RangeHighlighter? = null,
-    var insertEndOffset: Int = -1
-)
+    var insertEndOffset: Int = -1,
+    var popupOffset: Int = -1,
+    val latencyContext: LatencyContext,
+    var hasAccepted: Boolean = false
+) : Disposable {
```
rli, line 131: some of these can be private / don't need to be initialized in the primary constructor since the initial value is always expected to be the same

rli (approved):

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: before the InvocationContext for the specific jobId is added, it can be null, we did some null check to know if it's first request/response.

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/service/CodeWhispererService.kt
andrewyuq: so intially there's no InvocationContext in ongoingRequests, and then we contruct it in `initState` in `processCodeWhispererUi` and put it into the map. If after this function it's still null, then it means state creation failed due to e.g. all discarded suggestions or all empty suggestions.

PR #4922, opened 2024-09-26
Fix region selector drop-down getting cut off by page

samgst-amazon:
## Types of changes

## Description

Changed the behavior of the select element in the WebView for SSO login pages-- this fix applies both to the Q SSO and the Toolkit SSO pages.

Previously the dropdown could render the bottom of the list offscreen where users could not select the bottom options with their mouse. This fix allows users to now be able to see and scroll to the bottom of the list.

## Checklist
 
## License

### DIFF
--- a/plugins/core/webview/src/q-ui/components/ssoLoginForm.vue
+++ b/plugins/core/webview/src/q-ui/components/ssoLoginForm.vue
@@ -39,6 +39,10 @@
                 name="regions"
                 v-model="selectedRegion"
                 @change="handleRegionInput"
+                size="1"
+                onfocus='this.size=10;'
+                onblur='this.size=1;'
+                onchange='this.size=1; this.blur();'
                 tabindex="0"
             >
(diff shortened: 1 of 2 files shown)


rli (approved):

plugins/core/webview/src/q-ui/components/ssoLoginForm.vue
```diff
@@ -39,6 +39,10 @@
                 name="regions"
                 v-model="selectedRegion"
                 @change="handleRegionInput"
+                size="1"
+                onfocus='this.size=10;'
+                onblur='this.size=1;'
```
manodnyab, line 44: do we need to explicitly set onBlur?

samgst-amazon: You're right we don't need to explicity set the Onblur if we set it onChange.

samgst-amazon:
Setting this back to draft, as keyboard navigation reverts the dropdown to its previous behavior.

PR #4923, opened 2024-09-26
Fix case where user gets stuck while logging into Builder ID

rli:
If user ends up in state where they have a connection, but extension doesn't think it's valid for Q, then the sign-in flow will never complete and they need to restart the IDE for the new token to be picked up
 
## License

### DIFF
--- a/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
+++ b/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
@@ -220,8 +220,6 @@ abstract class LoginBrowser(
             Login
                 .BuilderId(scopes, onPendingToken, onError, onSuccess)
                 .login(project)
-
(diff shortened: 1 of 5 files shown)


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/webview/LoginBrowser.kt
```diff
@@ -220,6 +220,8 @@ abstract class LoginBrowser(
             Login
                 .BuilderId(scopes, onPendingToken, onError, onSuccess)
                 .login(project)
+
+            // TODO refresh the pane here for case when provider is no-op (i.e. provider exists and has a valid token), to fix issue where user is stuck waiting for browser
```
rli, line 224: maybe solved by abandoning idempotency semantics as current diff;
write test to confirm scenario when activeConnection() has delta from activeConnectionForFeature()

samgst-amazon (approved):

PR #4933, opened 2024-10-03
Fix toolkit connection dropdown hiding when panel is small

samgst-amazon:
## Types of changes

## Description

Updated ActionToolbar's layout strategy to ensure no wrapping or auto-hide occurs when components don't fit into the toolbar. This change also replaces the Deprecated layoutPolicy method. 

OLD BEHAVIOR:
oldDropdowngif

NEW BEHAVIOR:
newDropdowngif

## Checklist
 
## License


samgst-amazon:
Okay, this should be a 1 line fix but I am running into some versioning issues. The current line that we have that sets the layout policy uses a deprecated function that as of 2024.1 does not work(the documentation says this does nothing)

Replacing that with the new LayoutStrategy (instead of policy) works for the new versions, but does not compile for 2023.3 since it cannot resolve the references to the new ToolbarLayout Class in the import statement. 


layoutPolicy -- old and only works for 2023.3 <=. No effect for new versions, and the default behavior seems to have changed from NOWRAP to AUTO 

layoutStrategy -- new method for layout. seems to default to AUTO for instead of NOWRAP, works for new versions, doesn't compile for old versions since it imports a new package only available 2024.1 >=

Added in some logic to check the version, since importing and referencing com.intellij.openapi.actionSystem.toolbarLayout in old versions seems to break it.

I think once 2024.3 releases and we stop supporting 2023.3 we can change this to one single line:

layoutStrategy = ToolbarLayoutStrategy.NOWRAP_STRATEGY
Screenshot 2024-10-03 at 10 54 57 AM
Screenshot 2024-10-03 at 10 55 20 AM

plugins/toolkit/jetbrains-core/src-241+/software/aws/toolkits/jetbrains/core/explorer/ConnectionActionToolbarBuilder.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.intellij.openapi.actionSystem.toolbarLayout.ToolbarLayoutStrategy
+import com.intellij.util.ui.components.BorderLayoutPanel
+
+object ConnectionActionToolbarBuilder {
+    fun createToolbar(toolWindow: AwsToolkitExplorerToolWindow, group: DefaultActionGroup): BorderLayoutPanel {
+        return BorderLayoutPanel().apply {
+            addToCenter(
+                ActionManager.getInstance().createActionToolbar(ActionPlaces.TOOLBAR, group, true).apply {
+                    layoutStrategy = ToolbarLayoutStrategy.NOWRAP_STRATEGY
+                    setTargetComponent(toolWindow)
```
manodnyab, line 18: Since only this section differs across the versions, can we move the creation of this component to the util file and retain the creation for the panel in the init block of AWSToolkitExplorerToolWindow?

manodnyab (commented):
Can you please add a screenshot of the new dropdown to the description of the PR?

plugins/toolkit/jetbrains-core/src-241+/software/aws/toolkits/jetbrains/core/explorer/ConnectionActionToolbarBuilder.kt
samgst-amazon: Done & done, Thanks for the feedback :)

plugins/toolkit/jetbrains-core/src-233/software/aws/toolkits/jetbrains/core/explorer/ActionToolbarLayoutUtil.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.core.explorer
+
+import com.intellij.openapi.actionSystem.ActionToolbar
+
+object ActionToolbarLayoutUtil {
```
manodnyab, line 7: We probably dont need this object, it can be a standalone function

samgst-amazon: got it

manodnyab (approved):

plugins/toolkit/jetbrains-core/src-233/software/aws/toolkits/jetbrains/core/explorer/ActionToolbarLayoutUtil.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.aws.toolkits.jetbrains.core.explorer
+
+import com.intellij.openapi.actionSystem.ActionToolbar
+
+fun setToolbarLayout(toolbar: ActionToolbar) {
+    toolbar.layoutPolicy = ActionToolbar.NOWRAP_LAYOUT_POLICY
+}
```
rli, line 9: `ToolbarLayoutStrategy` doesn't exist in 233, so it would be less overall work if we just aliased
`ToolbarLayoutStrategy.NOWRAP_STRATEGY to be ActionToolbar.NOWRAP_LAYOUT_POLICY` i n233

PR #4944, opened 2024-10-07
Update telemetry: emit auth scopes in user state

samgst-amazon:
## Types of changes

## Description

Changed the emitUserState function to add authScopes to the telemetry record. Scopes were previously not recorded with this metric. 

## Checklist
 
## License


samgst-amazon:
The userState() function in AuthTelemetry did not have a parameter for authScopes-- hence just calling the TelemetryService function directly to add that datum in place. [That's how VSCode was doing it](https://github.com/aws/aws-toolkit-vscode/blob/c529794436e8a48f415e3a5e4012b787816804fe/packages/core/src/extension.ts#L260)

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
```diff
@@ -4,10 +4,13 @@
 package software.aws.toolkits.jetbrains.core.gettingstarted
 
 import com.intellij.openapi.project.Project
+import migration.software.aws.toolkits.jetbrains.services.telemetry.TelemetryService
```
rli, line 7:
```suggestion
import software.aws.toolkits.jetbrains.services.telemetry.TelemetryService
```

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
```diff
@@ -229,13 +233,27 @@ fun reauthenticateWithQ(project: Project) {
...
+            is LegacyManagedBearerSsoConnection -> connection.scopes
+            else -> emptyList()
+        }
+    }.toSet()
+
+    TelemetryService.getInstance().record(project) {
+        datum("auth_userState"){
+            createTime(Instant.now())
+            unit(Unit.NONE)
+            value(1.0)
```
rli, line 249: these are implicit

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
```diff
@@ -229,13 +233,27 @@ fun reauthenticateWithQ(project: Project) {
...
-        passive = true
-    )
+
+    val scopes = ToolkitAuthManager.getInstance().listConnections().flatMap { connection ->
+        when (connection) {
+            is ProfileSsoManagedBearerSsoConnection -> connection.scopes
+            is LegacyManagedBearerSsoConnection -> connection.scopes
+            else -> emptyList()
+        }
+    }.toSet()
```
rli, line 243: honestly it seems odd that we inspect every connection instead of what the extension is actively using, which seems more useful

rli (approved):

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
```diff
@@ -229,13 +234,26 @@ fun reauthenticateWithQ(project: Project) {
...
 
 fun emitUserState(project: Project) {
-    AuthTelemetry.userState(
-        project,
-        source = getStartupState().toString(),
-        authEnabledConnections = getEnabledConnections(project),
-        authStatus = getAuthStatus(project),
-        passive = true
-    )
+    val explorerConnection = ToolkitConnectionManager.getInstance(project).activeConnection()
```
samgst-amazon, line 237: Right now, I am just grabbing the connection being used by the explorer. I see that the activeConnection() function is marked as deprecated, but there is no alternative for getting the explorer connection from what I've seen. The getActiveConnectionForFeature function seems to work only for Q, CodeCatalyst, and CodeWhisperer.

samgst-amazon: If we want to look at all the active connections, I think it would be easy enough to use a function like:
https://github.com/aws/aws-toolkit-jetbrains/blob/80abea5c6be906bfa767400e5dcb19fdd2d9863a/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedTelemetryUtils.kt#L23
to get the scopes from each active connection. If we are looking at enabled connections in this Telemetry metric, why not the scopes for those enabled connections?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
manodnyab: This metric is currently in telemetryOverride. We need to use the one from common.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedTelemetryUtils.kt
```diff
@@ -70,6 +70,37 @@ fun getEnabledConnectionsForTelemetry(project: Project?): Set<AuthFormId> {
...
+        }
+    }
+
+    val explorerConnection = checkIamConnectionValidity(project)
+    addScopes(explorerConnection)
+
+    val codeCatalystConnection = checkBearerConnectionValidity(project, BearerTokenFeatureSet.CODECATALYST)
+    addScopes(codeCatalystConnection)
+
+    val codeWhispererConnection = checkBearerConnectionValidity(project, BearerTokenFeatureSet.CODEWHISPERER)
```
manodnyab, line 92: I wonder if we still need this since Q or we should just record Q

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedTelemetryUtils.kt
```diff
@@ -70,6 +70,37 @@ fun getEnabledConnectionsForTelemetry(project: Project?): Set<AuthFormId> {
...
+    fun addScopes(connection: ActiveConnection) {
+        if (connection !is ActiveConnection.NotConnected) {
+            val connectionScopes = connection.activeConnectionBearer?.scopes
+            if (connectionScopes != null) {
+                scopes.addAll(connectionScopes)
+            }
+        }
+    }
+
+    val explorerConnection = checkIamConnectionValidity(project)
```
manodnyab, line 86: We could also probably check the connection type and use the addScopes function only if the connection type is IDC. We can probably just hardcode it anyway

samgst-amazon: Check the connection type of the explorer only? Would that not ignore the other connections/scopes available if those other features are using the IdC connection type while the explorer isnt?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/GettingStartedAuthUtils.kt
samgst-amazon: I understand that our AuthTelemetry is being generated. no version of AuthTelemetry.userState() accepts authScopes as a parameter. I think this might be why? :

https://github.com/aws/aws-toolkit-common/blob/49de773b65efdeaadea5c30859cf286ccd74ec5b/telemetry/definitions/commonDefinitions.json#L2708C1-L2710C19

I'll try making some changes to common locally to see if I can generate a version of the function that does.

plugins/core/jetbrains-community/resources/telemetryOverride.json
samgst-amazon, line 1: Was literally so confused as to why the version from common with authScopes was not working... DOH!!

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedPanelUtils.kt
```diff
@@ -115,6 +116,20 @@ fun checkIamConnectionValidity(project: Project): ActiveConnection {
...
+        CredentialType.SsoProfile -> ActiveConnectionType.IAM_IDC
+        else -> ActiveConnectionType.IAM
+    }
+    return if (invalidConnection) {
+        ActiveConnection.ExpiredIam(connectionType = connectionType, activeConnectionIam = currConn)
+    } else {
+        ActiveConnection.ValidIam(connectionType = connectionType, activeConnectionIam = currConn)
+    }
+}
+
```
samgst-amazon, line 132: Wrote this because the other function was using isCredentialSSO would always return IAM, even in cases where it should be IAM_IDC.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedTelemetryUtils.kt
samgst-amazon: From Dhruvi: " let's not be concerned about old customers since most of them have migrated" 

I'll remove it.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedTelemetryUtils.kt
```diff
@@ -70,6 +71,36 @@ fun getEnabledConnectionsForTelemetry(project: Project?): Set<AuthFormId> {
 fun getEnabledConnections(project: Project?): String =
     getEnabledConnectionsForTelemetry(project).joinToString(",")
 
+fun getAuthScopesForTelemetry(project: Project?): Set<String> {
```
manodnyab, line 74: The function above repeats some of the logic used here, can we move the common logic to a function?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/gettingstarted/editor/GettingStartedPanelUtils.kt
manodnyab: The other function is for the getting started auth which contains a dialog that accepts a session name.
Where does this function determine the credential type?

samgst-amazon: I saw that there was a CredentialType enum attached to the selected credentials so I thought I could use that. If the credentialType is an SsoProfile, does that not mean it is part of an IDC session?

PR #4949, opened 2024-10-08
fix(dev): add error messages for upload URL expiry

chengoramazon:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### Problem

- Upload url would expire in a given time. The original error message is not clear for users what to do next.

### Solution

- Handle URL expiry explicitly.
- Add a clear error message.

JBExpired

## Checklist
- [x] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [x] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [ ] I have added metrics for my changes (if required)
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4938, opened 2024-10-04
fix(dev): Include "no change required exception" for EmptyPatch

tverney:
## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Description
Problem

Q dev won't be able to generate code for unrelated prompts. (e.g. explain the code)
It would be a bad CX if Q fail to generate code without guiding users the proper use cases.

Solution

Handle no code change required explicitly.
Allow users to provide more details

## Checklist
- [X] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [ ] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [ ] I have added metrics for my changes (if required)
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4953, opened 2024-10-09
fix(amazonq): Update exception handling for feature dev to improve observability

ctidd:
## Types of changes

## Description
**Problem:** Feature dev instrumentation 1) records all failures as `FeatureDevException`, losing information about the specific exception type, and 2) lacks annotation of `reasonDesc` and the operation related to the error. (By operation, although we do capture `amazonq_codeGenerationInvoke`,  we're missing the subprocess name such as `StartCodeGeneration`.) As a result, we do not have insight into feature dev failures in JetBrains.

**Solution:** 1) Update exception modeling pattern for feature dev functionality to derive subclasses of `FeatureDevException`, providing a meaningful class name when inspected as error `reason`. 2) On `FeatureDevException`, accept an `operation` and `desc` to provide further annotation (merged when logged as `reasonDesc`).

This change will have conflicts with https://github.com/aws/aws-toolkit-jetbrains/pull/4938 and https://github.com/aws/aws-toolkit-jetbrains/pull/4949. I'll rebase on top of those and reconcile any drift in a subsequent revision.

## Checklist
- [N/A] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [N/A] I have added metrics for my changes (if required)
 
## License

### DIFF
--- a/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
+++ b/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
@@ -6,40 +6,69 @@ package software.aws.toolkits.jetbrains.services.amazonqFeatureDev
 import software.aws.toolkits.jetbrains.services.amazonq.RepoSizeError
 import software.aws.toolkits.resources.message
 
-open class FeatureDevException(override val message: String?, override val cause: Throwable? = null) : RuntimeException()
+/**
+ * FeatureDevException models failures from feature dev operations.
+ *
+ * - Each failure is annotated based on className, operation, and a short desc. Use the `reason()` and `reasonDesc()` members for instrumentation.
+ * - To throw an exception without modeling, throw FeatureDevException directly.
+ */
+open class FeatureDevException(override val message: String?, val operation: String, val desc: String?, override val cause: Throwable? = null) :
+    RuntimeException() {
+    fun reason(): String = this.javaClass.simpleName
 
-class ContentLengthError(override val message: String, override val cause: Throwable?) : RepoSizeError, RuntimeException()
+    fun reasonDesc(): String =
+        when (desc) {
+            desc -> "$operation | Description: $desc"
+            else -> operation
+        }
+}
 
-class ZipFileError(override val message: String, override val cause: Throwable?) : RuntimeException()
+class NoChangeRequiredException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.exception.no_change_required_exception"), operation, desc, cause)
 
-class CodeIterationLimitError(override val message: String, override val cause: Throwable?) : RuntimeException()
+class EmptyPatchException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.exception.guardrails"), operation, desc, cause)
 
-class MonthlyConversationLimitError(override val message: String, override val cause: Throwable?) : RuntimeException()
+class ContentLengthException(
+    override val message: String = message("amazonqFeatureDev.content_length.error_text"),
+    operation: String,
+    desc: String?,
+    cause: Throwable? = null,
+) :
+    RepoSizeError, FeatureDevException(message, operation, desc, cause)
 
-class UploadURLExpired(
-    override val message: String = message(
-        "amazonqFeatureDev.exception.upload_url_expiry"
-    ),
-    override val cause: Throwable? = null,
-) : FeatureDevException(message, cause)
+class ZipFileCorruptedException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException("The zip file is corrupted", operation, desc, cause)
 
-internal fun featureDevServiceError(message: String?): Nothing =
-    throw FeatureDevException(message)
+class UploadURLExpired(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.exception.upload_url_expiry"), operation, desc, cause)
 
-internal fun codeGenerationFailedError(): Nothing =
-    throw FeatureDevException(message("amazonqFeatureDev.code_generation.failed_generation"))
+class CodeIterationLimitException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.code_generation.iteration_limit.error_text"), operation, desc, cause)
 
-internal fun uploadCodeError(): Nothing =
-    throw FeatureDevException(message("amazonqFeatureDev.exception.upload_code"))
+class MonthlyConversationLimitError(message: String, operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message, operation, desc, cause)
 
-internal fun conversationIdNotFound(): Nothing =
-    throw FeatureDevException(message("amazonqFeatureDev.exception.conversation_not_found"))
+class GuardrailsException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.exception.guardrails"), operation, desc, cause)
 
-internal fun apiError(message: String?, cause: Throwable?): Nothing =
-    throw FeatureDevException(message, cause)
+class PromptRefusalException(operation: String, desc: String?, cause: Throwable? = null) :
+    FeatureDevException(message("amazonqFeatureDev.exception.prompt_refusal"), operation, desc, cause)
 
-internal fun exportParseError(): Nothing =
-    throw FeatureDevException(message("amazonqFeatureDev.exception.export_parsing_error"))
+class ThrottlingException(operation: String, desc: String?, cause: Throwable? = null) :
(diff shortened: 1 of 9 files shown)


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/CodeGenerationState.kt
```diff
@@ -86,6 +91,10 @@ class CodeGenerationState(
             logger.warn(e) { "$FEATURE_NAME: Code generation failed: ${e.message}" }
             result = Result.Failed
             failureReason = e.javaClass.simpleName
+            if (e is FeatureDevException) {
+                failureReason = e.reason()
```
chengoramazon, line 95: it is the same with L93. Why don't we use the error className as reason and message as description?

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
```diff
@@ -6,33 +6,63 @@ package software.aws.toolkits.jetbrains.services.amazonqFeatureDev
...
-open class FeatureDevException(override val message: String?, override val cause: Throwable? = null) : RuntimeException()
+/**
+ * FeatureDevException models failures from feature dev operations.
+ *
+ * - Each failure is annotated based on className, operation, and a short desc. Use the `reason()` and `reasonDesc()` members for instrumentation.
+ * - To throw an exception without modeling, throw FeatureDevException directly.
+ */
+open class FeatureDevException(override val message: String, open val operation: String, open val desc: String?, override val cause: Throwable? = null)
+    : RuntimeException() {
+    fun reason(): String {
```
chengoramazon, line 17: Not seeing this necessary. We use class name as reason. Why don't create explicit Error class?

```
class UploadCodeError(override val message: String? = message("amazonqFeatureDev.exception.upload_code"), override val cause: Throwable? = null) : FeatureDevException(message, cause)

```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/CodeGenerationState.kt
```diff
@@ -100,6 +109,7 @@ class CodeGenerationState(
                 amazonqRepositorySize = repositorySize,
                 result = result,
                 reason = failureReason,
+                reasonDesc = failureReasonDesc,
```
chengoramazon, line 112: I see we pass `operation` around, but not seeing where we emit it.

ctidd: It's exposed inside `FeatureDevException#reasonDesc()`. Rather than have the instrumentation try to build `reasonDesc` consistently, `FeatureDevException` exposes how to get it, including both the operation and additional desc (if present).

```
fun reasonDesc(): String {
        return when (desc) {
            desc -> String.format("%s: %s", this.operation, desc) // <-- Here
            else -> String.format("%s", this.operation)
        }
}
```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/session/CodeGenerationState.kt
ctidd: > error className as reason

We do, but encoding it explicitly in `FeatureDevException.reason()` which contains the class name lookup. Here, 95 and 93 are redundant, but I figured being explicit for `FeatureDevException` would be most clear to a maintainer.

> message as description

`desc` (and `operation`) are fed into `reasonDesc` instead of `message`. I don't want to naively start logging message directly, as the cardinality may be too high. We can think of `desc` as a simplified or lower-cardinality message, explicitly for logging.

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
ctidd: > Why don't create explicit Error class, and replace those unclear FeatureDevException?

There are several catch-alls currently throwing `FeatureDevException` using`apiError()`. In this changeset, we retain the existing behavior of these not being modeled more explicitly. In other words, 1) enumerated error cases should be as you describe with explicit subclasses, 2) non-enumerated error cases (i.e. catch-alls) throw `FeatureDevException`. I agree we can/should enumerate additional errors, but I think that's beyond the scope of this changeset.

ctidd: > Not seeing this necessary. We use class name as reason. 

I'm happy to remove if you prefer. I expanded on this in another comment, so I'll resolve this conversation. It's a tradeoff between explicit expectations for how to use the class (this is what I biased toward in this revision) vs reducing redundancy at the call sites, which call this directly in addition to the lookup for other classes.

What do you think?

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
```diff
@@ -6,40 +6,69 @@ package software.aws.toolkits.jetbrains.services.amazonqFeatureDev
...
+        return this.javaClass.simpleName
+    }
 
-class ContentLengthError(override val message: String, override val cause: Throwable?) : RepoSizeError, RuntimeException()
+    fun reasonDesc(): String {
+        return when (desc) {
+            desc -> String.format("%s: %s", this.operation, desc)
+            else -> String.format("%s", this.operation)
```
rli, line 24:
```suggestion
            desc -> "$description: $desc"
            else -> operation
```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
```diff
@@ -6,40 +6,69 @@ package software.aws.toolkits.jetbrains.services.amazonqFeatureDev
...
 import software.aws.toolkits.resources.message
 
-open class FeatureDevException(override val message: String?, override val cause: Throwable? = null) : RuntimeException()
+/**
+ * FeatureDevException models failures from feature dev operations.
+ *
+ * - Each failure is annotated based on className, operation, and a short desc. Use the `reason()` and `reasonDesc()` members for instrumentation.
+ * - To throw an exception without modeling, throw FeatureDevException directly.
+ */
+open class FeatureDevException(override val message: String?, open val operation: String, open val desc: String?, override val cause: Throwable? = null)
```
rli, line 15:
```suggestion
open class FeatureDevException(override val message: String?, val operation: String, val desc: String?, override val cause: Throwable? = null)
```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
rli, line 1: thanks for cleaning this up

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/controller/FeatureDevController.kt
```diff
@@ -479,24 +471,36 @@ class FeatureDevController(
...
+                     is FeatureDevException -> {
+                        messenger.sendError(
+                            tabId = tabId,
+                            errMessage = err.message,
+                            retries = retriesRemaining(session),
+                            conversationId = session?.conversationIdUnsafe
+                        )
+                    }
+                    else -> {
+                        var msg = createUserFacingErrorMessage("$FEATURE_NAME request failed: ${err.message ?: err.cause?.message}")
```
rli, line 484: var can be val

rli (approved):

neilk-aws (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevExceptions.kt
ctidd: Incorporated as:

```
desc -> "$operation | Description: $desc"
else -> operation
```

chengoramazon (approved):
Just FYI: feels like metadata.metricName already play the role of operation, so have operation in the description maybe redundant. Sharing some metrics links offline.

PR #4968, opened 2024-10-15
SSO logout now keeps profile in config file, popup removed

samgst-amazon:
…onfig instead

## Types of changes

## Checklist
 
## License


rli (commented):
let's think about this change a bit more because customers don't like being asked to reauthenticate

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/ConfigFilesFacade.kt
```diff
@@ -197,6 +198,31 @@ class DefaultConfigFilesFacade(
         }
     }
 
+    override fun deleteSsoProfileScopesFromConfig(sessionName: String) {
```
rli, line 201: what if a user is depending on the session to have scopes outside of the extension?

PR #5040, opened 2024-11-01
fix supplemental context log latency always says 0ms

Will-ShaoHua:
## Description

logs say it's 0ms everytime
 
## License


plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/CodeWhispererFileContextProvider.kt
```diff
@@ -166,8 +167,7 @@ class DefaultCodeWhispererFileContextProvider(private val project: Project) : Fi
                     LOG.warn { "Failed to fetch supplemental context, empty list." }
                 }
 
-                // TODO: fix this latency
```
Will-ShaoHua, line 169: it was indeed correct, the incorrect part is the above log entry

rli (approved):

leigaol (approved):

PR #5055, opened 2024-11-06
Bump 243 EAP targets to IJ / RD RC1

rli:
## License


samgst-amazon (approved):
Thanks for updating this. Looking at git history-- aside form when there is a new version, is this something we have to manually update periodically when issues like this arise?

rli:
yes, but it seems like the platform doesn't like these EAPs so we can't build

PR #5056, opened 2024-11-06
Session retains credentials when disk full

samgst-amazon:
## Types of changes

## Description

Implemented an in-memory caching mechanism as a fallback when disk space is full

-Add InMemoryCache object to store credentials as ByteArrays
-Update writeKey to store data in memory if disk write fails due to space constraints
-Modify load functions to check in-memory cache is file read fails
-Maintain existing error handling and telemetry patterns

## Checklist
 
## License

### DIFF
--- a/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
+++ b/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
@@ -35,6 +35,9 @@ import software.aws.toolkits.core.utils.tryOrNull
 import software.aws.toolkits.core.utils.warn
 import software.aws.toolkits.telemetry.AuthTelemetry
 import software.aws.toolkits.telemetry.Result
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.IOException
 import java.io.InputStream
 import java.io.OutputStream
 import java.nio.file.Path
@@ -46,6 +49,7 @@ import java.time.Instant
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter.ISO_INSTANT
 import java.util.TimeZone
+import java.util.concurrent.ConcurrentHashMap
 
 /**
  * Caches the [AccessToken] to disk to allow it to be re-used with other tools such as the CLI.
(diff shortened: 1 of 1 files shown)


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
```diff
@@ -294,6 +322,27 @@ class DiskCache(
...
     private fun AccessToken.isDefinitelyExpired(): Boolean = refreshToken == null && !expiresAt.isNotExpired()
 
+    private fun storeInMemory(path: Path, consumer: (OutputStream) -> Unit) {
+        val byteArrayOutputStream = ByteArrayOutputStream()
+        consumer(byteArrayOutputStream)
+        val data = byteArrayOutputStream.toByteArray()
+        InMemoryCache.put(path.toString(), data)
+    }
+
+    private object InMemoryCache {
```
rli, line 332: tests?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
```diff
@@ -162,11 +174,18 @@ class DiskCache(
...
-        val inputStream = cacheFile.tryInputStreamIfExists() ?: return null
-
-        val token = loadAccessToken(inputStream)
-
-        return token
+        // If file exists, returns InputStream, if not returns null
+        return cacheFile.tryInputStreamIfExists()
+            //try to load and parse access token, returns AccessToken or null if expired
+            ?.let { loadAccessToken(it) }
+            // If file doesn't exist or loadAccessToken failed, try in-memory cache
```
rli, line 181: what happens if the state on disk does not match what is in memory?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
samgst-amazon: working on em

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
samgst-amazon: should check memory first I think for the load functions. Because if this is the case that there exists both a file and in-memory token(ClientReg too), then that means the most recent state that attempted to save would not fit on disk, so that should be checked first.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/credentials/sso/DiskCache.kt
```diff
@@ -98,12 +103,21 @@ class DiskCache(
...
         LOG.debug { "invalidateClientRegistration for $ssoRegion" }
+        InMemoryCache.remove(clientRegistrationCache(ssoRegion).toString())
         clientRegistrationCache(ssoRegion).tryDeleteIfExists()
     }
 
     override fun loadClientRegistration(cacheKey: ClientRegistrationCacheKey): ClientRegistration? {
         LOG.debug { "loadClientRegistration for $cacheKey" }
-        val inputStream = clientRegistrationCache(cacheKey).tryInputStreamIfExists()
+        val cacheFile = clientRegistrationCache(cacheKey)
+        // try InMemoryCacheFirst in case of stale registration on full disk
```
rli, line 113: shouldn't the disk be the source of truth?
consider the case where multiple different JetBrains IDE instances are running

PR #5134, opened 2024-11-26
fix(amazonq): download each job's zip to its own directory

dhasani23:
## Types of changes

## Description

When users run consecutive transformations (ex. a Java upgrade with multiple diffs requested, then a SQL conversion), the summary/diff from the 1st transformation would show up when trying to view the summary/diff for the 2nd transformation, because we were re-using a temp directory to store all artifacts.

Solution is to create a new and unique temp directory to hold each transformation job's results.

## Checklist
 
## License


plugins/amazonq/codetransform/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/CodeModernizerArtifact.kt
```diff
@@ -52,6 +53,7 @@ open class CodeModernizerArtifact(
          * If anything goes wrong during this process an exception is thrown.
          */
         fun create(zipPath: String): CodeModernizerArtifact {
+            tempDir = createTempDirectory("codeTransformArtifacts-", UUID.randomUUID().toString())
```
rli, line 56: leaking state through the companion object. please have backlog for better encapsulation

Issue #5150, opened 2024-12-03
Plugin Exception: notifications.json not found


rli:
v3.43

```
com.intellij.diagnostic.PluginException: D:\Users\...\AppData\Local\JetBrains\IntelliJIdea2024.3\aws-static-resources\notifications.json [Plugin: aws.toolkit.core]
	at com.intellij.ide.startup.impl.StartupManagerImplKt$launchActivity$1.invokeSuspend(StartupManagerImpl.kt:502)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:608)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:873)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:763)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:750)
Caused by: java.nio.file.NoSuchFileException: D:\Users\...\AppData\Local\JetBrains\IntelliJIdea2024.3\aws-static-resources\notifications.json
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:234)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:379)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:431)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:420)
	at com.intellij.platform.core.nio.fs.DelegatingFileSystemProvider.newInputStream(DelegatingFileSystemProvider.java:72)
	at java.base/java.nio.file.Files.newInputStream(Files.java:159)
	at software.aws.toolkits.core.utils.PathUtilsKt.inputStream(PathUtils.kt:27)
[...]

PR #5157, opened 2024-12-03
Fix: catch error when notification file is not found

samgst-amazon:
NoSuchFileException handling during notification processing fails gracefully instead of bubbling up to an IDE error. 

## Types of changes

## Description

Fix an issue where if there is no notification file in the cache, notification retrieval fails silently instead of bubbling up an exception

## Checklist
 
## License


samgst-amazon:
Working on a better implementation of the Resource Resolver for notification polling but this should handle #5150

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/ProcessNotificationsBase.kt
```diff
@@ -38,12 +39,17 @@ class ProcessNotificationsBase(
...
-        if (content.isEmpty()) {
+        try {
+            val path = Paths.get(PathManager.getSystemPath(), NOTIFICATIONS_PATH)
+            val content = path.inputStream().bufferedReader().use { it.readText() }
+            if (content.isEmpty()) {
+                return null
+            }
+            return NotificationMapperUtil.mapper.readValue(content)
+        } catch (e: Exception) {
+            LOG.warn { "Error reading notifications file: $e" }
```
rli, line 50: would prefer to fix the underlying issue

rli (approved):

PR #5186, opened 2024-12-11
Set expiry for dismissed notifications

samgst-amazon:
## Types of changes

## Description

We store the list of dismissed notif ids in aws.xml, but we need to limit the size of this saved list.
To do this, dismissed notification IDs stored in the aws.xml are timestamped. Each time the persistent state component is loaded, any notifications older than 60 days will be removed from the set. 

## Checklist
 
## License


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
```diff
@@ -9,37 +9,68 @@ import com.intellij.openapi.components.Service
...
+import java.time.Instant
+
+class InstantConverter : Converter<Instant>() {
+    override fun toString(value: Instant): String = value.toEpochMilli().toString()
+
+    override fun fromString(value: String): Instant = Instant.ofEpochMilli(value.toLong())
+}
+
+data class DismissedNotification(
+    @Attribute
```
manodnyab, line 26: are the attribute tags required here?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
```diff
@@ -9,37 +9,68 @@ import com.intellij.openapi.components.Service
...
+    val id: String = "",
+    @Attribute(converter = InstantConverter::class)
+    val dismissedAt: Instant = Instant.now(),
+)
+
+data class NotificationDismissalConfiguration(
+    @Property
```
manodnyab, line 33: Is this tag required?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
```diff
@@ -9,37 +9,68 @@ import com.intellij.openapi.components.Service
...
+    var dismissedNotifications: MutableSet<DismissedNotification> = mutableSetOf(),
```
manodnyab, line 34:
```suggestion
    val dismissedNotifications: MutableSet<DismissedNotification> = mutableSetOf(),
```

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
```diff
@@ -9,37 +9,68 @@ import com.intellij.openapi.components.Service
...
 class NotificationDismissalState : PersistentStateComponent<NotificationDismissalConfiguration> {
     private val state = NotificationDismissalConfiguration()
+    private val retentionPeriod = Duration.ofDays(60) // 2 months
 
     override fun getState(): NotificationDismissalConfiguration = state
 
     override fun loadState(state: NotificationDismissalConfiguration) {
-        this.state.dismissedNotificationIds.clear()
-        this.state.dismissedNotificationIds.addAll(state.dismissedNotificationIds)
+        this.state.dismissedNotifications.clear()
```
manodnyab, line 46: Why do we need to clear the set? Wont each value anyway be unique since it's a set?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
samgst-amazon: No they were leftover from an earlier implementation, you're right they are not needed

samgst-amazon: I needed these tags so that the eprsstent framework would properly store the data to and from the .xml file. Without these attribute/Property tags the dismissed notification state component is not saved on IDE exit.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
samgst-amazon: These properties need to be var for the serialization to work without the property/attribute tags

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/core/notifications/NotificationStateUtils.kt
samgst-amazon: Setting these fields to var removes the need for these tags to exist. needs to be val/not var otherwise the default accessor will not work

plugins/core/jetbrains-community/tst/software/aws/toolkits/jetbrains/core/notifications/NotificationDismissalStateTest.kt
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Test
+    fun `dismissing new notification retains it`() {
+        state.dismissNotification("new-notification")
+
+        assertTrue(state.isDismissed("new-notification"))
+    }
+
+    @Test
```
manodnyab, line 84: Is this test redundant since the previous 3 tests handle the same scenario?

manodnyab (approved):

PR #5207, opened 2024-12-13
Fix test failures caused by mockk not cleaning up

rli:
~30 tests fail on initial run but succeed on automatic retry due to improper mockk cleanup
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5258, opened 2025-01-14
test(amazon Q): fix test failure caused by MockTelemetryExtension throwing NotAMockException

Will-ShaoHua:
## Description

Now there are 2 ways to use NoOpTelemetryService
1. if devs are not interested in validating the metric contents, don't need do anything as default test implementation is NoOpTelemetryService
2. if devs are interested in the metric contents, can use `MockTelemetryServiceExtension` which will inject a spy of batcher and replace the original test implementation in junit before hook and revert it in after hook so that you can use ArgumentCaptor to capture what's going through the batcher

Relevant #5207 

### root cause:
Originally by doing `class NoOpTelemetryService : TelemetryService(publisher, spy(DefaultTelemetryBatcher(publisher))) {`, it's not 100% guaranteed that the TelemetryBatcher passed in is a mockito spy, and it result in `NotAMockException` thrown from these cases.

### solution
Manually inject spy batcher separately in a different step instead of IDE doing the work when initializing app/project service components

 
## License


plugins/core/jetbrains-community/resources/META-INF/aws.toolkit.core.xml
```diff
@@ -26,8 +26,7 @@
                             serviceImplementation="software.aws.toolkits.jetbrains.core.credentials.DefaultCredentialManager"
                             testServiceImplementation="software.aws.toolkits.jetbrains.core.credentials.MockCredentialsManager"/>
         <applicationService serviceInterface="migration.software.aws.toolkits.jetbrains.services.telemetry.TelemetryService"
-                            serviceImplementation="software.aws.toolkits.jetbrains.services.telemetry.DefaultTelemetryService"
-                            testServiceImplementation="software.aws.toolkits.jetbrains.services.telemetry.NoOpTelemetryService"/>
+                            serviceImplementation="software.aws.toolkits.jetbrains.services.telemetry.DefaultTelemetryService"/>
```
rli, line 29: can we keep it as no-op so that tests dont send metrics by default, but you can add the rule if you want to assert metrics?

Will-ShaoHua: aha that makes sense

plugins/core/jetbrains-community/tstFixtures/software/aws/toolkits/jetbrains/services/telemetry/MockTelemetryService.kt
```diff
@@ -32,8 +30,14 @@ class NoOpPublisher : TelemetryPublisher {
...
 sealed class MockTelemetryServiceBase : ExternalResource() {
-    private val mockTelemetryService: NoOpTelemetryService
-        get() = service<TelemetryService>() as NoOpTelemetryService
+    private val publisher: NoOpPublisher by lazy { NoOpPublisher() }
+    private val batcher: TelemetryBatcher by lazy { spy(DefaultTelemetryBatcher(publisher)) }
+
+    private val mockTelemetryService: NoOpTelemetryService by lazy { NoOpTelemetryService(publisher, batcher) }
+
+    override fun before() {
+        ApplicationManager.getApplication().replaceService(TelemetryService::class.java, mockTelemetryService, mockTelemetryService)
```
rli, line 39: replaceService needs to be reverted in the after()

plugins/core/jetbrains-community/tstFixtures/software/aws/toolkits/jetbrains/services/telemetry/MockTelemetryService.kt
```diff
@@ -32,11 +39,21 @@ class NoOpPublisher : TelemetryPublisher {
...
+
+    override fun before() {
+        // hack because @TestDisposable doesn't work here as it's not a test
+        disposableParent = Disposer.newDisposable()
+        ApplicationManager.getApplication().replaceService(TelemetryService::class.java, mockTelemetryService, disposableParent)
+    }
 
     override fun after() {
         reset(batcher())
+        Disposer.dispose(disposableParent)
```
Will-ShaoHua, line 56: @rli is it sufficient for the purpose of reverting `replaceService`?

rli (approved):

PR #5261, opened 2025-01-14
fix(Amazon Q): fix for test generation target file missing from payload.

ashishrp-aws:
… and fix for /test query for non-supported languages.

## Types of changes

## Description

Target File was getting filtered out due to featureDevSession Regex pattern matching based filtering. For example, a folder called `builder` is getting filtered out and the files in it because regex pattern include `build`.
- fix to include target file regardless of filters.
- fix for regex patterns to match correctly. for example, for `build` to filter out
build
build/
build/something
path/to/build
path/to/build/
path/to/build/something

and not filter out `/builder/`

- fix for query for unsupported languages and external files

## Checklist
 
## License


chungjac (approved):

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -218,9 +206,39 @@ class FeatureDevSessionContext(val project: Project, val maxProjectSizeBytes: Lo
 
     // gitignore patterns are not regex, method update needed.
     private fun convertGitIgnorePatternToRegex(pattern: String): String = pattern
+        // Escape special regex characters except * and ?
```
rli, line 209: tests?

ashishrp-aws: will add

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
```diff
@@ -102,11 +117,73 @@ class FeatureDevSessionContextTest : FeatureDevTestBase(HeavyJavaCodeInsightTest
...
+    fun `test pattern starting with slash`() {
+        val input = "/root/file.txt"
+        val expected = "^root/file\\.txt(?:/.*)?$"
+        assertEquals(expected, featureDevSessionContext.convertGitIgnorePatternToRegex(input))
+    }
+
+    @Test
+    fun `test pattern ending with slash`() {
+        val input = "build/"
+        val expected = "(?:^|.*/?)build/.*"
```
rli, line 158: these patterns have polynomial complexity. should there be a filter on input to prevent DoS of user's machine?

rli: also can we investigate what it would look like to use the same glob logic that git uses to match files? seems like that might be simpler than trying to build these patterns

ashishrp-aws: Makes sense. will take a  look into it. Wanted to drop regex matching entirely for /test. Only reason still being used is to restrict payload size which is leading to payload size errors.

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
```diff
@@ -102,11 +117,73 @@ class FeatureDevSessionContextTest : FeatureDevTestBase(HeavyJavaCodeInsightTest
...
+            "builder/GetTestBuilder.java"
         )
 
         assertTrue(zippedFiles == expectedFiles)
     }
+
+    @Test
+    fun `test basic pattern conversion`() {
+        val input = "*.txt"
+        val expected = "(?:^|.*/?)[^/]*[^/]\\.txt(?:/.*)?\$"
```
neilk-aws, line 130: How do I validate if this is correct? 

This appears to be a snapshot test of what is built versus a functional test that demonstrates that what is an example file that would've been dropped is no longer dropped

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -252,16 +259,59 @@ class FeatureDevSessionContext(val project: Project, val maxProjectSizeBytes: Lo
...
+            .replace("?", "[^/]") // Match single character except path separator
+
+        // Handle start of pattern
+        result = if (result.startsWith("/")) {
+            "^${result.substring(1)}"
+        } else {
+            "(?:^|.*/?)$result"
+        }
+
+        // Handle end of pattern
```
neilk-aws, line 296: What is this doing? How do I know if the behavior is comprehensive?

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
```diff
@@ -252,16 +259,59 @@ class FeatureDevSessionContext(val project: Project, val maxProjectSizeBytes: Lo
...
 
     // gitignore patterns are not regex, method update needed.
-    private fun convertGitIgnorePatternToRegex(pattern: String): String = pattern
-        .replace(".", "\\.")
-        .replace("*", ".*")
-        .let { if (it.endsWith("/")) "$it.*" else "$it/.*" } // Add a trailing /* to all patterns. (we add a trailing / to all files when matching)
+    fun convertGitIgnorePatternToRegex(pattern: String): String? {
+        // Skip invalid patterns for length check
+        if (pattern.length > MAX_PATTERN_LENGTH) {
+            return null
```
neilk-aws, line 265: What happens when this returns null? Why was this value decided to be 256?

neilk-aws (changes requested):

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
ashishrp-aws: we can test for pattern matching as well. That's why i included more files in the project zip to validate it. Including false positives.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
ashishrp-aws: 256 is generally placed limit. If it returns null , nothing happens pattern matching skips incase of null.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
ashishrp-aws: to handle folder structures in patterns. for `/src` and `src/` cases

ashishrp-aws: Its too match all contents inside the directory in case of `src/` and glob matching to check for `.svg` and `.png` and other patterns everywhere and not just in root.

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
```diff
@@ -102,11 +117,73 @@ class FeatureDevSessionContextTest : FeatureDevTestBase(HeavyJavaCodeInsightTest
             "gradlew",
             "gradlew.bat",
             "README.md",
-            "settings.gradle",
```
neilk-aws, line 105: Just noticed that `settings.gradle` and `build.gradle` are being dropped. These should be included as a part of the zip for /dev. 

Is this the result of changing the gitignore used for the test or a change in the actual behavior of the function?

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
neilk-aws: My point is that these tests do little to validate behaviors and protect against regressions. If we need to update the logic, how does one go about constructing the expected value? 

Regardless, outside of the zip test above, I'd like to see more examples of each of these type of files and how they respond to the presence of such inputs in the gitignore.

plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt
neilk-aws: Is a different way of saying this that if the pattern is longer than 256 characters, then we will make no effort to exclude it and will always include it in the zip?

plugins/amazonq/chat/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/FeatureDevSessionContextTest.kt
ashishrp-aws: Can amend this for /dev.

ashishrp-aws:
https://github.com/aws/aws-toolkit-jetbrains/pull/5261#discussion_r1917749200
I have included as many file names i could in the tests.

neilk-aws (approved):

rli (approved):

manodnyab (approved):

PR #5301, opened 2025-01-28
docs(amazonq): Update autobuild setting strings

neilk-aws:
## Types of changes

## Description

Improves the formatting and strings presented to customers when the allow /dev to run code and test commands automatically. Also includes minor variable name changes to improve internal consistency. 

## Checklist
 
## License


osdemah (approved):

manodnyab (approved):

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/settings/CodeWhispererSettings.kt
```diff
@@ -153,7 +153,7 @@ class CodeWhispererConfiguration : BaseState() {
     @get:Property
     val stringValue by map<CodeWhispererStringConfigurationType, String>()
 
-    val projectAutoBuildConfigurationMap by map<String, Boolean>()
+    val autoBuildSetting by map<String, Boolean>()
```
rli, line 156: needs `@get:Property` like the other properties in here
after release, renaming this will be a breaking change for customers so it is probably with a test case

rli: oh great there are no existing tests on this state component
basic ser/deser tests similar to below should be sufficient coverage:

https://github.com/aws/aws-toolkit-jetbrains/blob/2e39dc0ff02b6b76d4c52fc2861130262ec9f9b3/plugins/amazonq/codewhisperer/jetbrains-community/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererModelConfiguratorTest.kt#L297-L440

rli: on another note, doing a map of project -> boolean isnt really the best way to model this.

if you don't want the setting to be application-wide, you should define a new project-level state component so that the setting is saved per-project. right now the implementation looks like it makes a new check box for every project the user opens?

neilk-aws: The setting is application wide and mirrors the structure that we're able to give in VSCode. I think there is opportunity for improvement in modeling, but I don't think I want to block on that right now. I'd love to learn about where project-level state component and what other features are using it. If we do migrate, we'll implement the logic to import the settings accordingly.

neilk-aws: Thanks for the pointer on the tests. I've added them here, with an initial focus on our property. I've kept it generic for other teams to be able to contribute more validation there.

rli: switching between the two is not exactly trivial so it would be better to get that right now.

the configurable is already set up to support project-level settings so the only difference would be creating a new state service, and you can get rid of the weird map layer
```kt
    private val codeWhispererProjectSettings
        get() = CodeWhispererProjectSettings.getInstance(project)
```

```kt
            group(message("aws.settings.codewhisperer.feature_development")) {
                row {
                    text(message("aws.settings.codewhisperer.feature_development.allow_running_code_and_test_commands"))
                }
                row {
                    checkBox(codeWhispererProjectSettings.getAutoBuildSetting()).apply {
                        connect.subscribe(
                            ToolkitConnectionManagerListener.TOPIC,
                            object : ToolkitConnectionManagerListener {
                                override fun activeConnectionChanged(newConnection: ToolkitConnection?) {
                                    enabled(isCodeWhispererEnabled(project))
                                }
                            }
                        )

                        bindSelected(
                            getter = { codeWhispererProjectSettings.isAutoBuildFeatureEnabled() },
                            setter = { newValue -> codeWhispererProjectSettings.toggleAutoBuildFeature(newValue) }
                        )
                    }
                }
            }    
```

```kt
@Service(Level.PROJECT)
@State(name = "codewhispererProejctSettings", storages = [Storage("aws.xml")])
class CodeWhispererProjectSettings(private val project: Project) : PersistentStateComponent<CodeWhispererProjectConfiguration> {
...
}

class CodeWhispererProjectConfiguration : BaseState() {
    @get:Property
    val autoBuildSetting by property(false)
}
```

neilk-aws: We can take this as a follow up. Given we're trying to merge the feature branch into main today, I think we'll accept the risk of needing to implement a pathway to import settings from the storage layer from the existing format into a the new format once we migrate to it.

rli: sure

PR #5290, opened 2025-01-23
feat(amazonq): Prefetch next inline recommendation

evanliu048:
## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Description
This PR introduces the ability for users to see the next recommendation immediately after accepting the current one, provided there is a subsequent recommendation available. The key enhancements include:

1. **Automatic Prefetching:** Introduced a function that calls the CodeWhisperer API to fetch the next recommendation while the current one is being reviewed.
2. **Session Promotion** Added functionality to display the next recommendation instantly upon accepting the current recommendation (if there is one)
3.  **Helper Utilities:** Implemented helper functions to calculate the necessary file information for the next request and to send telemetry events related to the subsequent recommendation.

## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5323, opened 2025-02-04
refactor(inline): Remove Dummy Popup Creation in Prefetch Flow and Ensure Session Validation

evanliu048:
## Description
This change removes the unnecessary creation of a dummy popup (`JBPopup`) in the CodeWhisperer prefetch flow. Instead, the `popup` field in InvocationContext is now nullable (`JBPopup? = null`). This ensures that prefetch logic does not create unused UI elements while maintaining compatibility with existing workflows.
reference: https://github.com/aws/aws-toolkit-jetbrains/pull/5290

Additionally, it introduces session ID validation before prefetching recommendations, ensuring that only relevant recommendations are fetched.

## Checklist
 
## License


andrewyuq (approved):
I feel like it's ok to just create the another popup at nextInvocaationContext creation time and don't use it until when promoteNextInvocationIfAvailable is called

PR #5366, opened 2025-02-12
feat(amazonq): initialize flare lsp on project start

rli:
## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQLspService.kt
```diff
@@ -71,6 +72,32 @@ internal class LSPProcessListener : ProcessListener {
...
+    override fun dispose() {
+    }
+
+    companion object {
+        private val LOG = getLogger<AmazonQLspService>()
+        fun getInstance(project: Project) = project.service<AmazonQLspService>()
+    }
+}
+
+private class AmazonQServerInstance(cs: CoroutineScope) : Disposable {
```
samgst-amazon, line 100: Can we pass in `private val project ` here? I was using it for workspace folders -- unless there is a better way to be doing so

rli: sure

samgst-amazon (approved):

PR #5409, opened 2025-02-26
feat(amazonq): generate scanName from workspace config

ctlai95:
## Types of changes

## Description

All code scans are run with a scanName as a UUID which makes it impossible to track the bug lifecycle of projects. We are not leveraging the bug tracking feature which allows grouping together different scan runs of the same project by passing in the same scanName between runs.

## Checklist
 
## License


zuoyaofu (approved):

plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/codescan/CodeWhispererCodeScanSession.kt
```diff
@@ -366,6 +375,26 @@ class CodeWhispererCodeScanSession(val sessionContext: CodeScanSessionContext) {
     private fun isAutoScan(): Boolean =
         sessionContext.codeAnalysisScope == CodeWhispererConstants.CodeAnalysisScope.FILE && !sessionContext.sessionConfig.isInitiatedByChat()
 
+    private fun generateScanName(): String {
+        val clientId = AwsSettings.getInstance().clientId
```
manodnyab, line 379: Is scan name used anywhere other than metrics?

ctlai95: Yes, in the CreateUploadUrl and CreateCodeScan APIs

manodnyab: Hey, quick clarification: Why aren't we using the original UUIDs, what would be the difference here in using client ids? Since I see we are generating a new one at the end of this file?

ctlai95: Using a UUID for every scan run means that the backend is unable to identify different runs of a scan belong to the same project. So here we want to introduce some way to identify that this scan is for a specific user, for a specific project, and for a specific scan type.

rli: the client id is already part of the user agent? why do we need to attach it again?

ctlai95: @rli I guess technically we don't need to, but since `scanName` is exposed as an API field the client should generate the hash value

PR #5411, opened 2025-02-26
feat(amazonq): Add support for complex workspaces for /dev and /doc.

ctidd:
This provides multi-content-root workspace support for /dev and /doc.

## Types of changes

## Description

**Problem:**

Q does not support multi-root workspaces for /dev, /doc, and related features.

**Solution:**

This change starts by adding a `Workspace` construct for uploads involving the `FeatureDevSessionContext`, which detects the IDE workspace's root directory as the common ancestor of the workspace's "content roots", to serve as the root of the addressable workspace filesystem layout. Then, when choosing what files to upload, files within this tree which are members of the workspace's content roots are included. This broadens /dev and /doc to support a multi-content-root workspace as a single working tree, while continuing to respect content roots of the workspace and not uploading extraneous content.

The change itself is straightforward, identifying the common ancestor and basing further work on that root. From here, a few changes unravel:

1. **Problem:** We need to add support for `.gitignore` in subdirectories (across ProjectContext and FeatureDev). Otherwise, we would confuse users with project-level `.gitignore` configurations, and it makes sense to fully support subdirectory `.gitignore` configurations rather than treat projects as a special case. **Solution:** I've **removed nearly all custom gitignore parsing** for FeatureDev, offloading the work to the IDE's VCS manager for more native gitignore support.
2. **Problem:** We need to require that uploaded files are part of a project "content root" directory, and not simply included because they are under the common ancestor directory (which would otherwise over-include ancillary directories that are not content roots as members of the IDE workspace). **Solution:** Filter uploaded content based on content roots. This matches the handling already performed by `ProjectContextProvider`, moving incrementally in the direction of common handling of workspaces across features. (**Note:** /test and /review appear to use `guessModuleDir` in their own independent context implementations, and so would continue to have minor symptoms for complex workspaces after these changes. I have not touched these areas.)
3. **Problem:** `ProjectContextProvider` had a dependency on `FeatureDevSessionContext` for its own ignore rules. **Solution:** I've reworked this implementation so that `ProjectContextProvider` and `FeatureDevSessionContext` now depend on a simplified shared primitive, which should be easier to audit and understand than the use-case specific rules that were inside `FeatureDevSessionContext`. However, `FeatureDevSessionContext` is kept in a global location to be consumed by /doc. (Removing this /doc -> /dev coupling is deeper than I was willing to go with these changes as broad as they already are.)

Additional granular changes to behavior:

1. **Problem:** We had inconsistent behaviors when auto build for /dev was enabled vs disabled. Overzealous filtering of file types was applied when it was not enabled. **Solution:** Now, filtering is applied based on what was previously the more permissive behavior of the two. Now, in both cases, only `.gitignore` and a few additional artifact-exclusion filters are applied (e.g. `.git`, `node_modules`, etc).

## Checklist
- [TODO] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
 
## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
```diff
@@ -0,0 +1,225 @@
...
+                    }
+                }
+            }
+        }
+        zipFilePath
+    }.toFile()
+
+    private suspend fun createTemporaryZipFileAsync(block: suspend (FileSystem) -> Unit): Path = withContext(getCoroutineBgContext()) {
+        // Don't use Files.createTempFile since the file must not be created for ZipFS to work
+        val tempFilePath: Path = Paths.get(FileUtils.getTempDirectory().absolutePath, "${UUID.randomUUID()}.zip")
```
vikshe, line 182: I think it's better to create specific folder in the temp location and store zip there. 
Also how is the zip file deleted after upload ?

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqDoc/controller/DocController.kt
```diff
@@ -967,7 +966,7 @@ class DocController(
                 message = message("amazonqDoc.prompt.choose_folder_to_continue")
             )
 
-            val selectedFolder = selectFolder(context.project, currentSourceFolder)
+            val selectedFolder = selectFolder(context.project, workspaceRoot)
```
vikshe, line 969: should it be session.context.selectionRoot instead ?

ctidd: It's really a niche case: A user would need to select a folder, have the selected folder be too big, and then trigger selection again.

- Before: The file tree is opened on the current source folder to select new source folder.
- Now: The file tree is opened on the workspace root to select a new source folder.

This is a more workspace-oriented behavior (i.e. "select what you want") instead of nudging the user toward a folder inside their selected working folder.

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/Workspace.kt
```diff
@@ -0,0 +1,120 @@
...
+    ".hg",
+    ".idea",
+    ".project",
+    ".rvm",
+    ".svn",
+    "node_modules",
+    "build",
+    "dist",
+    "*.zip",
+    "*.bin",
```
ctidd, line 80: These are the only file types which will additionally be filtered relative to before, IFF auto build is enabled. Looking into whether this change should be removed based on known use cases.

neilk-aws (approved):

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
```diff
@@ -0,0 +1,224 @@
...
+     */
+    var workspaceRoot = findWorkspaceRoot(workspaceContentRoots) ?: error("Cannot detect base workspace root")
+
+    private val changeListManager = ChangeListManager.getInstance(project)
+
+    private var _selectionRoot = workspaceRoot
+
+    // This function checks for existence of `devfile.yaml` in customer's repository, currently only `devfile.yaml` is supported for this feature.
+    fun checkForDevFile(): Boolean {
+        val devFile = File(addressableRoot.toString(), "devfile.yaml")
```
rli, line 62: `devfile.yml` is also valid according to spec

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
```diff
@@ -0,0 +1,224 @@
...
+        val filesToIncludeFlow = channelFlow {
+            // chunk with some reasonable number because we don't actually need a new job for each file
+            files.chunked(50).forEach { chunk ->
+                launch {
+                    for (file in chunk) {
+                        send(file)
+                    }
+                }
+            }
+        }
```
rli, line 150: i see that this is the existing impl, but the flow seems superfluous for what is being accomplished here

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/Workspace.kt
```diff
@@ -0,0 +1,111 @@
...
+    "dist",
+).map(::regexTestOf)
+
+/**
+ * Provides an extremely strict set of global ignore rules, suitable for purely text-based-sources use cases.
+ */
+val additionalGlobalIgnoreRulesForStrictSources =
+    additionalGlobalIgnoreRules +
+        listOf(
+            // FIXME: It is incredibly brittle that this source of truth is a "telemetry" component
```
rli, line 80: +1

rli (approved):

PR #5459, opened 2025-03-10
fix(amazonq): always send creds on lsp init if available

rli:
Previously lsp server gets token by luck depending on timing of credential events. Instead, client should explicitly send token after server finishes init
 
## License


leigaol (approved):

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/auth/DefaultAuthCredentialsService.kt
```diff
@@ -39,7 +41,8 @@ class DefaultAuthCredentialsService(
             ?: return
 
         provider.currentToken()?.accessToken?.let { token ->
-            updateTokenCredentials(token, false)
+            // assume encryption is always on
```
samgst-amazon, line 44: Does this mean that we are expecting any process triggering onChange is utilizing our instance of the EncryptionManager?

samgst-amazon (approved):

PR #5482, opened 2025-03-19
fix(amazonq): Fix cases where content may be incorrectly excluded from workspace.

ctidd:
## Types of changes

## Description

**Problem 1:** Incorrectly filtering out directories in `additionalGlobalIgnoreRulesForStrictSources` due to missing null check on file extension. Fixed and added tests.

**Problem 2:** Because we start traversal at the content roots, but filter out any content not within the selection root, we were incorrectly stopping traversal when the selection root is nested within any content root. In this case, we were exiting without traversing to find the `selectedRoot`. Fixed and added tests.

**Problem 3:** Add support for DevFile (as normally supported, when at workspace root) when workspace root is not a content root. Fixes excluding the DevFile when it's not in a content root.

## Checklist
- [N/A] I have added metrics for my changes (if required)
 
## License

### DIFF
--- a/plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
+++ b/plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
@@ -18,7 +18,7 @@ import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.io.FileUtils
 import software.aws.toolkits.jetbrains.core.coroutines.getCoroutineBgContext
 import software.aws.toolkits.jetbrains.services.amazonq.QConstants.MAX_FILE_SIZE_BYTES
-import software.aws.toolkits.jetbrains.utils.isDevFile
+import software.aws.toolkits.jetbrains.utils.isWorkspaceDevFile
 import software.aws.toolkits.resources.AwsCoreBundle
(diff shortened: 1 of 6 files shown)


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/project/FeatureDevSessionContext.kt
```diff
@@ -109,10 +83,38 @@ open class FeatureDevSessionContext(val project: Project, val maxProjectSizeByte
...
+                            return false
+                        }
+
+                        // Exclude files and directories outside the selection root when working on a subset of the workspace:
+                        if (!VfsUtil.isAncestor(selectionRoot, file, false)) {
+                            // Because we traverse from the content root, ensure we continue traverse toward the selection root:
+                            // (Handles when selection root is inside a content root)
+                            if (VfsUtil.isAncestor(file, selectionRoot, false)) {
+                                return true
+                            }
```
ctidd, line 109: This is the actual change in this file. The rest of the change is inlining `shouldIncludeInZipFile` logic into the `VirtualFileVisitor`. I started out with a change to expose an enum (i.e. `CONTINUE` in addition to `INCLUDE` and `EXCLUDE`) instead of a boolean, but found it was clearer to inline the logic.

gandhi-21 (approved):

neilk-aws (approved):

rli (approved):

PR #5507, opened 2025-03-25
style: ban org.junit.Assert and org.junit.jupiter.api.Assertions

rli:
Fluent assertions through AssertJ result in more useful test output
 
## License


plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/telemetry/otel/OTelService.kt
rli, line 1: revert

samgst-amazon (approved):

PR #5549, opened 2025-04-10
fix(amazonq): reauth workflow should not prompt profile selection page

evanliu048:
### Problem

On IDE startup, `QRegionProfileManager` loads the active profile from `aws.xml`.  
However, `validateProfile()` is called before SSO reauth completes, causing `listRegionProfiles()` to return empty. This results in the active profile being reset to `null`.

### Fix

Update `getIdcConnectionOrNull()` to return the connection only if it's a valid Q connection **and the token is authorized**.  
Also, after successful reauth, immediately show the Q panel to avoid triggering the profile selection flow.

### Summary

- Skip profile validation when token isn't ready
- Ensure Q panel appears after reauth

### Changes

- Refined `getIdcConnectionOrNull()` to check for `BearerTokenAuthState.AUTHORIZED`

This should improve the IDE startup flow for users who rely on SSO, reducing friction and avoiding false-positive "profile selection required" states.

## Checklist
 
## License


rli (approved):
this is probably worth a test case

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/profile/QRegionProfileManager.kt
```diff
@@ -191,11 +192,15 @@ class QRegionProfileManager : PersistentStateComponent<QProfileState>, Disposabl
     }
 
     private fun getIdcConnectionOrNull(project: Project): AwsBearerTokenConnection? {
-        val connection = ToolkitConnectionManager.getInstance(project).activeConnectionForFeature(QConnection.getInstance())
-        if (connection is AwsBearerTokenConnection && !connection.isSono()) {
-            return connection
+        val manager = ToolkitConnectionManager.getInstance(project)
+        val connection = manager.activeConnectionForFeature(QConnection.getInstance()) as? AwsBearerTokenConnection
```
rli, line 196: i think at this point we need to completely remove CodeWhispererConnection if we are assuming everything is now QConnection

evanliu048: I may be missing part of the context, but I believe I'm using `QConnection` here—let me know if I'm misunderstanding anything

Will-ShaoHua: because there was a old `CodeWhispererConnection.kt` before `QConnection` came in, so ideally we will check both connection for backward compatiblity. You can check `isQConnected`. I am in favor of removing codewhispererConnection as it would be a burden to check 2 connection type everywhere and considering we've been using Q_Scope for roughly 1 year, it shouldn't harm to remove CwsprConnection. I can follow up on the cwspr connectin deprecation if @rli you don't have concern about the deprecation

evanliu048: Thanks for the clarification Will.  We would need a separate or to handle the codewhisperer connection deprecation right

Will-ShaoHua: yes

Will-ShaoHua (approved):

PR #5579, opened 2025-04-15
feat(amazonq): Remove unsupported message for non-java python languages in /test

jkuruba:
## Types of changes

## Description
In /test feature, test generation in languages other than Java and Python display message that they were not supported. Tests are generated for all languages, so removed message to the user that /test only supports Python and Java

JB-Lang-Expansion

## Checklist
 
## License


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqCodeTest/controller/CodeTestChatController.kt
Randall-Jiang, line 1: Why we need add file out of workspace error message here? Is it relevant with removing the unsupported Language use case?

jkuruba: This is an existing check for out of workspace handling that should remain (see check at line 248). It is not related to language support

Randall-Jiang (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqCodeTest/controller/CodeTestChatController.kt
```diff
@@ -244,26 +244,22 @@ class CodeTestChatController(
                     })
                 }
                 .build()
+            if (!fileInfo.fileInWorkspace) {
```
samgst-amazon, line 247: on line 203 in this file there is still a check for isLanguageSupported, and a progress bar is removed in the else block on 221. If we are removing the message about unsupported language cases because /test generates them anyways, why not show the progress bar?

Wouldn't this just look like inconsistent behavior on the user side -- because now they don't know that non-java/python /test cases are handled differntly?

samgst-amazon (approved):

PR #5610, opened 2025-04-21
deps(flare): Add registry entry for flare manifest endpoint

manodnyab:
## License


plugins/amazonq/src/main/resources/META-INF/plugin.xml
```diff
@@ -92,6 +92,9 @@
                      restartRequired="true"/>
         <registryKey key="inline.completion.rem.dev.use.rhizome" description="Defined by IntelliJ. Used for Amazon Q to display suggestions on remote."
                      defaultValue="false" restartRequired="true"/>
+        <registryKey defaultValue="https://aws-toolkit-language-servers.amazonaws.com/remoteWorkspaceContext/0/manifest.json"
```
rli, line 95: default null

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/artifacts/ManifestFetcher.kt
```diff
@@ -28,8 +29,12 @@ class ManifestFetcher(
...
         private val logger = getLogger<ManifestFetcher>()
 
-        private const val DEFAULT_MANIFEST_URL =
-            "https://aws-toolkit-language-servers.amazonaws.com/qAgenticChatServer/0/manifest.json"
+        private val DEFAULT_MANIFEST_URL = getManifestEndpoint()
+
+        private fun getManifestEndpoint(): String {
+            val endpoint = Registry.get("amazon.q.flare.endpoint").asString()
+            return endpoint.ifBlank { "https://aws-toolkit-language-servers.amazonaws.com/qAgenticChatServer/0/manifest.json" }
+        }
```
rli, line 37: value will be read on class init
```suggestion
        private val DEFAULT_MANIFEST_URL
            get() = getManifestEndpoint()

        private fun getManifestEndpoint(): String {
            val endpoint = Registry.get("amazon.q.flare.endpoint").asString()
            return endpoint.ifBlank { "https://aws-toolkit-language-servers.amazonaws.com/qAgenticChatServer/0/manifest.json" }
        }
```
should probably just delete DEFAULT_MANIFEST_URL and only use the method

PR #5627, opened 2025-04-23
fix(amazonq): fix LinkageError when starting AmazonQLspService (attempt 2)

rli:
In certain IDE environments, for example when user has the latest Scala plugin, Amazon Q attempts to load lsp4j through
transitive optional dependencies instead of through the IDE.

Ideally we do not bundle the library, but this is not possible until JetBrains can guarantee stronger isolation between plugins
 
As part of this we optimistically exclude Gson in hopes that it does not have the same issue. SLF4J is additionally packaged due to the same linkage error, though it seems to be pulling in an older version from somewhere else.

We also need to bump the Kotlin version, since there is a bug where Kotlin compiler runs out of memory for some reason if we exclude Gson from the distribution bundle.

## License


buildSrc/src/main/kotlin/toolkit-publishing-conventions.gradle.kts
```diff
@@ -54,7 +54,8 @@ configurations {
 
     // Make sure we exclude stuff we either A) ships with IDE, B) we don't use to cut down on size
     runtimeClasspath {
-        exclude(group = "org.slf4j")
+        exclude(group = "com.google.code.gson")
```
manodnyab, line 57: This package is included only in certain environments right but we exclude it in all?

rli: yeah it does not hurt

manodnyab (approved):

PR #5635, opened 2025-04-24
fix(amazonq): don't reset profile config if validation fails to list profiles

rli:
After user selects a profile, if subsequent validation fails because of networking / etc, then we incorrectly clear their selected profile
 
## License


Will-ShaoHua (approved):

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/profile/QRegionProfileManager.kt
```diff
@@ -59,16 +58,24 @@ class QRegionProfileManager : PersistentStateComponent<QProfileState>, Disposabl
...
+     */
+    @Deprecated("This is a giant hack and we are not handling all the cases")
     @RequiresBackgroundThread
     fun validateProfile(project: Project) {
         val conn = getIdcConnectionOrNull(project)
         val selected = activeProfile(project) ?: return
-        val profiles = tryOrNull {
+        val profiles = try {
             listRegionProfiles(project)
+        } catch (_: Exception) {
```
manodnyab, line 71: Can we catch network exceptions here?

rli: i think the only exception acceptable to not catch is if they are not allowed to use Q

PR #5652, opened 2025-04-29
build(amazonq): remove bundled mynah-ui distribution

rli:
No longer needed since we are fetching from DEXP distribution
 
## License


manodnyab:
we need some files from here

samgst-amazon (approved):

samgst-amazon:
What files are we still using here?

PR #5713, opened 2025-05-13
fix(amazonq): reduce log spam and frequency of token updates

rli:
We do not need to update tokens every 10s

Additionally, if we provide the connection metadata in the token update request, then we reduce overall RPC calls
 
## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/auth/DefaultAuthCredentialsService.kt
```diff
@@ -180,7 +184,7 @@ class DefaultAuthCredentialsService(
         val payload = UpdateConfigurationParams(
             section = "aws.q",
             settings = mapOf(
-                "profileArn" to QRegionProfileManager.getInstance().activeProfile(project)?.arn
+                "profileArn" to "invalidinvalidinvalid"
```
rli, line 187: revert

Will-ShaoHua: imao

samgst-amazon (approved):

PR #5723, opened 2025-05-15
fix(amazonq): cancel chat export doesn't bubble error up to user

samgst-amazon:
Clicking cancel after save popup window would bubble up an error to the user in the IDE, and also print the error to the chat UI. 

This change throws the right error when a user selects cancel on the save window and filters out the error from being printed in the chat.

## Types of changes

## Checklist
 
## License

### DIFF
--- a/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
+++ b/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
@@ -401,7 +401,9 @@ class BrowserConnector(
                                 val cause = if (e is CompletionException) e.cause else e
 
                                 // dont post error to UI if user cancels export
-                                if (!(cause is ResponseErrorException && cause.responseError.message == "Export cancelled by user")) {
+                                if (cause is ResponseErrorException && cause.responseError.code == "ResponseErrorCode.RequestCancelled) {
+                                    return@whenComplete
+}
                                     LOG.error { "Failed to perform chat tab bar action $e" }
                                     params.tabId?.let {
                                         browser.postChat(chatCommunicationManager.getErrorUiMessage(it, e, null))
--- a/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
+++ b/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
@@ -27,6 +27,7 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
 import org.cef.browser.CefBrowser
 import org.eclipse.lsp4j.TextDocumentIdentifier
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException
 import software.aws.toolkits.core.utils.error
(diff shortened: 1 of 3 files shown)


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
```diff
@@ -382,6 +384,7 @@ class BrowserConnector(
 
             CHAT_TAB_BAR_ACTIONS -> {
                 handleChat(AmazonQChatServer.tabBarActions, node) { params, invoke ->
+                    println("handling TabBarActions")
```
rli, line 387:
```suggestion
```

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
```diff
@@ -396,9 +399,14 @@ class BrowserConnector(
                                     )
                                 )
                             } catch (e: Exception) {
-                                LOG.error { "Failed to perform chat tab bar action $e" }
-                                params.tabId?.let {
-                                    browser.postChat(chatCommunicationManager.getErrorUiMessage(it, e, null))
+                                val cause = if (e is CompletionException) e.cause else e
+
+                                // dont post error to UI if user cancels export
+                                if (!(cause is ResponseErrorException && cause.responseError.message == "Export cancelled by user")) {
```
rli, line 405:
```suggestion
                                if (cause is ResponseErrorException && cause.responseError.code == "ResponseErrorCode.RequestCancelled) {
                                    return@whenComplete
}
```

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQLanguageClientImpl.kt
```diff
@@ -230,8 +233,7 @@ class AmazonQLanguageClientImpl(private val project: Project) : AmazonQLanguageC
 
                 chosenFile?.let {
                     ShowSaveFileDialogResult(chosenFile.file.path)
-                    // TODO: Add error state shown in chat ui instead of throwing
-                } ?: throw Error("Export failed")
+                } ?: throw ResponseErrorException(ResponseError(ResponseErrorCode.RequestCancelled, "Export cancelled by user", null))
```
rli, line 236: lol

samgst-amazon: I think we need to throw this here to get the specific `RequestCancelled` error otherwise the ResponseErrorCode is `InternalError` which I think is a little too generic and I'm afraid it may unintentionally catch things. Is there any other reason the chosenFile would be null?

rli: seems like they either choose a file or they dont

manodnyab: The other error condition that I can think of here is if a file is chosen which is not html or md if it exists but with the filtering dont think a user should run into this

PR #5724, opened 2025-05-16
fix(amazonq): Switch profiles if not selected in Q chat

manodnyab:
## Description

If a user gets Select Q Developer Profile card in chat, check if an active profile already exists and update Flare with the information. If an active profile doesn't, the dialog to change profiles is shown to the user.
Ideally the user should never land in this state since profiles are chosen before the chat window is shown.

## Checklist
 
## License


Will-ShaoHua (approved):

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/toolwindow/AmazonQToolWindow.kt
```diff
@@ -25,8 +24,7 @@ class AmazonQToolWindow private constructor(
     val component
         get() = chatPanel.component
 
-    fun disposeAndRecreate() {
-        Disposer.dispose(chatPanel)
```
Will-ShaoHua, line 29: not sure what will happen if we keep creating new panel and they are never disposed, will it cause memory leak

rli: yes we would need to dispose if we are creating new panels

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/toolwindow/AmazonQToolWindow.kt
```diff
@@ -25,8 +24,7 @@ class AmazonQToolWindow private constructor(
...
+    fun recreatePanel() {
         chatPanel = AmazonQPanel(project, scope)
```
Will-ShaoHua, line 28: @manodnyab will it help if we assign the new panel first then dispose the stale one?

Will-ShaoHua:
also #5729

plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/toolwindow/AmazonQToolWindow.kt
Will-ShaoHua: maybe not, just tested

Will-ShaoHua: #5729 would fix the crash issue

manodnyab: Removing this part, if 5729 fixes the crash

rli (approved):

PR #5737, opened 2025-05-20
fix(amazonq): fix code issues tab not appearing in remote

rli:
on remote, each client gets its own instance of the tool window.

logic needs to handle this case, otherwise it will only make changes against the 'main' IDE, and will not be visible to remote clients

## License


Will-ShaoHua:
masterpiece

manodnyab (approved):

PR #5761, opened 2025-05-27
Add logging to project scans in /review

manodnyab:
## License


rli (commented):
take a look at `CodeWhispererCodeScanSession` because they are suppressing all the server exceptions and request ids with generic messages

PR #5770, opened 2025-05-30
feat(amazonq): mcp support

leigaol:
## Types of changes

## Checklist
 
## License

### DIFF
--- a/.changes/next-release/feature-6de3f384-aa3b-47fe-b895-d41ef23ed6da.json
+++ b/.changes/next-release/feature-6de3f384-aa3b-47fe-b895-d41ef23ed6da.json
@@ -1,0 +1,4 @@
+{
+  "type" : "feature",
+  "description" : "Support for Amazon Q Builder ID paid tier"
+}
--- a/plugins/amazonq/codewhisperer/jetbrains-community/resources/META-INF/plugin-codewhisperer.xml
+++ b/plugins/amazonq/codewhisperer/jetbrains-community/resources/META-INF/plugin-codewhisperer.xml
@@ -139,5 +139,7 @@
                 class="software.aws.toolkits.jetbrains.services.cwc.commands.codescan.actions.CodeScanCompleteAction" />
         </group>
 
+        <action id="q.manage.subscription"
+                class="software.aws.toolkits.jetbrains.services.codewhisperer.actions.ManageSubscription"/>
     </actions>
 </idea-plugin>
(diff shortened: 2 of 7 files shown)


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQChatServer.kt
```diff
@@ -174,6 +178,18 @@ object AmazonQChatServer : JsonRpcMethodProvider {
         Any::class.java
     )
 
+    val listMcpServers = JsonRpcRequest(
+        LIST_MCP_SERVERS_REQUEST_METHOD,
+        ListMcpServersParams::class.java,
```
rli, line 183: couldnt these have just been any?

laileni-aws: Sorry, Can you elaborate on this please?

rli: the request/response parameter types can be `LspAny` because they are pure passthrough?

laileni-aws: So I guess @rli is suggesting to have in this format

```
val listMcpServers = JsonRpcRequest(
    LIST_MCP_SERVERS_REQUEST_METHOD,
    LspAny::class.java, // Change from ListMcpServersParams::class.java
    Any::class.java
)

val mcpServerClick = JsonRpcRequest(
    MCP_SERVER_CLICK_REQUEST_METHOD,
    LspAny::class.java, // Change from McpServerClickParams::class.java
    Any::class.java
)
```

rli: `LspAny` for both but yes

laileni-aws: but there are certain values assigned to 
```
// List MCP Servers Parameters
data class ListMcpServersParams(
    val filter: Map<String, FilterValue>? = null
)

// MCP Server Click Parameters
data class McpServerClickParams(
    val id: String,
    val title: String? = null,
    // here Any is sufficient to serialize/deserialze complex objects
    val optionsValues: Map<String, Any>? = null
)
```

We can use LspAny::class.java but I guess we can be more specific in this case ?

manodnyab (commented):
can you please add a changelog?

plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQChatServer.kt
rli: anything that is primarily passthrough between flare <-> ui, even if the extension needs a couple of these fields, should be modeled as LspAny, otherwise we will lose data when fields are added/changed. we do not have the ability to directly import the types like in flare/VSC and do not want to go through every single flare commit to make sure none of these types have changed.

there is an ongoing project to codegen these types so that we can use the actual model.

rli (approved):

Issue #5811, opened 2025-06-11
Amazon Q Plugin freeze IntelliJ after upgrade to 3.74


chaobai:
**Describe the bug**

Since the upgrade of Amazon Q plugin on IntelliJ 3.74 released last Friday, it started to freeze IntelliJ altogether from time to time. Noticeably, there was a background process called “Amazon Q Helper (Workspace Context)” constantly running occupying at least one core of the CPU. It was the first time I saw this. I tried to kill the background process but that didn’t work well, the IntelliJ will come back to life briefly then got frozen again in a matter of minutes. For the time being, we have to disable the plugin altogether.

It might worth mentioning that our project is huge, which might partially contribute to the problem, but it was fine before the latest release of plugin, as long as we didn’t ask it to analyse the whole workspace. If there is a way to disable, bypass or limit the resource usage of the background job, that might help to work around our immediate problem.

**Your Environment**

- OS: macOS Sequoia 15.5
- JetBrains product: IntelliJ Ultimate
- JetBrains product version: 2025.1.2
- Amazon Q Plugin version: 3.74

**Additional context**

Error log attached
```
Freeze for 113 seconds
Sampled time: 40100ms, sampling rate: 100ms, GC time: 1933ms (1%), Class loading: 0%
 
The stack is from the thread that was blocking EDT
 
com.intellij.diagnostic.Freeze
    at [java.base@21.0.6/java.io.FileOutputStream.writeBytes(Native](mailto:java.base@21.0.6/java.io.FileOutputStream.writeBytes(Native) Method)
    at [java.base@21.0.6/java.io.FileOutputStream.write(FileOutputStream.java:367)](mailto:java.base@21.0.6/java.io.FileOutputStream.write(FileOutputStream.java:367))
    at [java.base@21.0.6/java.io.BufferedOutputStream.implWrite(BufferedOutputStream.java:217)](mailto:java.base@21.0.6/java.io.BufferedOutputStream.implWrite(BufferedOutputStream.java:217))
    at [java.base@21.0.6/java.io.BufferedOutputStream.write(BufferedOutputStream.java:206)](mailto:java.base@21.0.6/java.io.BufferedOutputStream.write(BufferedOutputStream.java:206))
    at [java.base@21.0.6/java.io.FilterOutputStream.write(FilterOutputStream.java:110)](mailto:java.base@21.0.6/java.io.FilterOutputStream.write(FilterOutputStream.java:110))
[...]

PR #5816, opened 2025-06-12
fix(amazonq): move workspace lsp messages off blocking thread calls

rli:
we are inappropriately freezing EDT while passing workspace context to the LSP

#5811

## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQLspService.kt
```diff
@@ -286,11 +284,32 @@ class AmazonQLspService(private val project: Project, private val cs: CoroutineS
...
+    suspend fun<T> executeIfRunning(runnable: suspend AmazonQLspService.(AmazonQLanguageServer) -> T): T? {
+        val lsp = try {
+            withTimeout(1.seconds) {
+                val holder = mutex.withLock { instance }.await()
+                holder.initializeResult.join()
+
+                holder.languageServer
+            }
+        } catch (_: Exception) {
+            LOG.debug { "LSP not running" }
```
manodnyab, line 310: nit: should this be an info with exception message?

PR #5880, opened 2025-07-03
fix(amazonq): Disable auto-suggestions when paused in the status bar

manodnyab:
## Types of changes

Inline completion suggestions were being shown even when paused from the status bar. 
Removing the manual call filter that disables the provider only when call is manual

## Checklist
 
## License

### DIFF
--- a/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/QInlineCompletionProvider.kt
+++ b/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/QInlineCompletionProvider.kt
@@ -592,7 +592,7 @@ class QInlineCompletionProvider(private val cs: CoroutineScope) : InlineCompleti
         val project = editor.project ?: return false
 
         if (!isQConnected(project)) return false
-        if (!CodeWhispererExplorerActionManager.getInstance().isAutoEnabled() && event.isManualCall()) return false
+        if (!CodeWhispererExplorerActionManager.getInstance().isAutoEnabled()) return false
         if (QRegionProfileManager.getInstance().hasValidConnectionButNoActiveProfile(project)) return false
         return true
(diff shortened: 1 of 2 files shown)


plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/popup/QInlineCompletionProvider.kt
```diff
@@ -592,7 +592,7 @@ class QInlineCompletionProvider(private val cs: CoroutineScope) : InlineCompleti
         val project = editor.project ?: return false
 
         if (!isQConnected(project)) return false
-        if (!CodeWhispererExplorerActionManager.getInstance().isAutoEnabled() && event.isManualCall()) return false
+        if (!CodeWhispererExplorerActionManager.getInstance().isAutoEnabled()) return false
```
rli: won't this will disable completions completely if auto is not on

rli:
```kt
// not sure if current impl is completely correct either
if (event.isManualCall()) return true
if (!CodeWhispererExplorerActionManager.getInstance().isAutoEnabled()) return false
```

manodnyab: We want it to disable when auto is not on right? Auto is the pause and resume suggestions in the dropdown

manodnyab: Discussed offline, enabling manual call for when auto-suggestions are paused

samgst-amazon (approved):

PR #5906, opened 2025-07-15
fix(amazonq): isolate flare from user aws config files

rli:
due to an AWS JS SDK v2 bug (which is EoL), profile redeclarion, though spec compliant, will result in a flare crash
 
## License


plugins/amazonq/shared/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/lsp/AmazonQLspService.kt
```diff
@@ -476,6 +477,9 @@ private class AmazonQServerInstance(private val project: Project, private val cs
                         put("NODE_EXTRA_CA_CERTS", it)
                     }
 
+                    put("AWS_CONFIG_FILE", emptyFile)
```
manodnyab, line 480: why are they dependent on each other?

rli: because of default SDK behavior?

manodnyab (approved):

Issue #5950, opened 2025-08-05
Error with double slash in WSL path when using Windows and WSL


andrew-styles:
**Describe the bug**
When using any Jetbrains IDE and Amazon Q plugin working on a project in WSL the // in the path throws an error.  

**To reproduce**
1. Open a project in WSL such that the path starts with \\wsl$\
2. Authenticate Amazon Q
3. Write "Hello" in Ask a question box.

**Expected behavior**
Get a response from Amazon Q

**Screenshots**
An error occurred while processing your request.
[UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//"): An unexpected error occurred, check the logs for more information.

**Your Environment**
- OS: Windows 11
- JetBrains product: WebStorm 2025.2
- JetBrains product version: Build #WS-252.23892.411, built on August 1, 2025
- AWS Toolkit version: 3.88.251
- SAM CLI version:
- JVM/Python version:

**Additional context**

andrew-styles:
Duplicate of https://github.com/aws/aws-toolkit-jetbrains/issues/5947

Sorry!

andrew-styles:
Duplicate of https://github.com/aws/aws-toolkit-jetbrains/issues/5947

Issue #5947, opened 2025-08-04
Amazon Q interactions not working after sign in


handmadecodes:
<img width="523" height="297" alt="Image" src="https://github.com/user-attachments/assets/62312f88-f122-4896-9ba2-6f1e39386401" />

**Describe the bug**
Just signed in to Amazon Q in WebStorm and then say hello to it and get this:

An error occurred while processing your request.
[UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//"): An unexpected error occurred, check the logs for more information.

**To reproduce**
1. Sign in
2. Say Hello

**Expected behavior**
Gives a hello/greeting response

**Screenshots**
See attached screenshot

**Your Environment**
- OS: Windows 11 Pro
- JetBrains WebStorm 2025.1.4.1
- AWS Toolkit version: 3.87.251
- SAM CLI version: N/A
- Python version: Py 3.10.11

handmadecodes:
Looks like it has to do with me working on a project in WSL. But this should work as it is working with other AI tools

handmadecodes:
Mapped to a network drive, now it appears to be a working workaround.

andrew-styles:
Mapping to network drive does work around the double slash in path but also breaks references for using node on WSL. Would be great if this could get fixed!

JUGaspar:
I experience the same issue.

aseemxs:
Please update to the latest version and let us know if this is still an issue.

mb-fron:
Am getting it in 3.103.253 in PHPStorm 2025.3.2

gianstraf:
This issue still persists.

- PhpStorm 2025.3.2 or PhpStorm 2025.3.3
- Plugin (https://plugins.jetbrains.com/plugin/24267-amazon-q)  3.105.253

@aseemxs could you reopen the ticket ?

gianstraf:
@aseemxs any update on this issue ?

Issue #5799, opened 2025-06-08
Amazon Q Chat 3.74.251 not working on WSL2 folder


BenoitMercier:
**Describe the bug**
After updating to 3.74.251, chat no longer works when project is on a WSL2 folder like \\wsl$\Ubuntu\home\...

The log shows this:
2025-06-08T15:33:25.638Z: 26676  [WARN] Unknown tokenizer class "CodeSageTokenizer", attempting to construct from base class.
2025-06-08T15:33:26.080Z: 26676  [WARN] Unknown tokenizer class "CodeSageTokenizer", attempting to construct from base class.
2025-06-08T15:33:27.209Z: 26676  [INFO] Using Cuda(Linux) or DirectML(Windows).
2025-06-08T15:33:27.210Z: 26676  [INFO] Using ONNX default thread management
2025-06-08T15:33:27.210Z: 26676  [WARN] removing requested execution provider "cuda" from session options because it is not available: backend not found.
2025-06-08T15:33:27.745Z: 26676  [INFO] Embedding provider initialized.
2025-06-08T15:33:27.753Z: 26676  [INFO] Using Cuda(Linux) or DirectML(Windows).
2025-06-08T15:33:27.753Z: 26676  [INFO] Using ONNX default thread management
2025-06-08T15:33:27.754Z: 26676  [WARN] removing requested execution provider "cuda" from session options because it is not available: backend not found.
2025-06-08T15:33:28.364Z: 26676  [INFO] Embedding provider initialized.
**2025-06-08T15:33:29.837Z: 26676  [INFO] Found 0 symbols in this workspace among folders:** 

**To reproduce**
1. Open a project on a folder like \\wsl$\Ubuntu\home\...
2. Initiate chat with any prompt

**Screenshots**
![Image](https://github.com/user-attachments/assets/2f840a10-9dd3-4fbf-8c71-71c6fd3f0922)

**Your Environment**
- OS: Windows 11 Home 24H2
- JetBrains product: Webstorm
- JetBrains product version: 2025.1.2

**Additional context**
Amazon Q works well if the project is on the local drive or a mapped drive.

iMattPro:
This is happening no matter what because of the mis-match between version naming of the various Amazon/AWS plugins, ie: `3.74.251` and `3.74-251`. Downgrade the 3.73-251 and you'll be back in business again until they fix this.

schandroo:
[...]

PR #5960, opened 2025-08-08
Prevents Q UriError : Validate file URI and fallback to Path.toUri() for WSL

Lee-WonJun:
## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)

### Summary

Fix repeated `UriError` when Amazon Q starts with a **WSL-backed workspace**. We now **validate the constructed `file:` URI** and **fallback to JDK’s `Path.toUri()`** if the URI violates Rule (no authority while the path begins with `//`, maybe RFC 3986). This keeps legacy behavior where it’s safe, and prevents Q from rejecting the workspace folder URI.

### Related issues

* Fixes/addresses: #5799, #5950, #5947 (same failure mode reported across these)

### Root cause

During `InitializeParams` creation, `workspaceFolders` receives URIs derived from `VirtualFile`.
Internally we were manually constructing file URIs like:

```kotlin
URI("file", "", file.absoluteFile.toURI().path, null)
```

On **WSL** paths this can become:

```
file:////wsl.localhost/Ubuntu/...
```

This has **no authority** and a **path that starts with `//`**, which violates RFC 3986. Amazon Q rejects it with:

image

Older versions reportedly worked because we used a fn in Nio:

```
file.toNioPath().toUri().toString()
```

[[#5462](https://github.com/aws/aws-toolkit-jetbrains/pull/5462/files)](https://github.com/aws/aws-toolkit-jetbrains/pull/5462/files), when we switched away from the JDK-generated URI toward manual assembly.

### What changed

* After we build the manual `file:` URI, we **inspect the resulting URI**:

  * If `scheme == "file"` **and** `authority` is empty **and** `path` starts with `//`, it’s considered **non-compliant** → **fallback**.
* **Fallback** uses JDK normalization:

  ```kotlin
  file.toPath().toAbsolutePath().normalize().toUri()
  ```
* `jar:` and `jrt:` handling **unchanged**.

### Why fallback (and not just always `Path.toUri()`)?
I personally think standardizing on `file.toPath().toAbsolutePath().normalize().toUri()` (or another well-defined canonicalization strategy) would be simpler and sufficient. However, to maintain backward compatibility, address this specific case, and respect any implicit intent behind the current manual construction, I implemented a conservative fallback: we keep the existing construction and only fall back to the JDK-generated URI when the constructed value would violate rule

### Tests

https://github.com/user-attachments/assets/bf000941-e0c3-472e-a28b-7c4cc3e5fa85

* Manually verified with the plugin:
  * WSL folder workspaces: Q initializes and runs successfully.
  * File/workspace views load as expected.
  
* add unit test

---

## Checklist

* [x] My code follows the code style of this project
* [x] I have added tests to cover my changes
* [ ] A short description of the change has been added to the **[[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
* [ ] I have added metrics for my changes (if required)

## License


LokeshDogga13 (approved):

Lee-WonJun:
I added the test with @EnabledOnOs(OS.WINDOWS) so that it only runs on Windows,
because UNC paths are non-standard on UNIX systems.

andrewyuq:
@Lee-WonJun  can you run `:plugin-amazonq:shared:jetbrains-community:detektMain` on your end?

Lee-WonJun:
> @Lee-WonJun can you run `:plugin-amazonq:shared:jetbrains-community:detektMain` on your end?

I modified the code in a different environment and later ran the check to verify it, but I think I missed reviewing the lint results. My apologies.

PR #5970, opened 2025-08-14
feat(amazonq): add support for CodeReview and DisplayFindings tools, alter behavior of explain and fix buttons

BlakeLazarine:
Alters JB plugin to support CodeReview and DisplayFindings tools which are already present in flare.

This enables the Q agent to use the Q Code Review functionality and also migrates /review to call the agent.

The IssueDetails panel has been modified to remove the generated fix, and instead just has options for Explain and Fix, which send information about the issue to the Q Chat.

image

image

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [ ] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [ ] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [ ] I have added metrics for my changes (if required)
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5992, opened 2025-08-26
chore: cleanup/reduce de/ser cycles introduced in #5970

rli:
Reduce the amount of Gson calls added by #5970
 
## License


plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/webview/BrowserConnector.kt
```diff
@@ -581,6 +583,21 @@ class BrowserConnector(
...
+    // TODO: move to software.aws.toolkits.jetbrains.services.amazonq.lsp.model
+    data class FlareAdditionalMessages(
+        val additionalMessages: List<FlareAggregatedFindings>?,
+    )
+
+    // TODO: move to software.aws.toolkits.jetbrains.services.amazonq.lsp.model
+    data class FlareAggregatedFindings(
+        val messageId: String,
+        val body: List<AggregatedCodeScanIssue>,
+    )
```
BlakeLazarine: Only findings messages have this format, others do not. I am worried if we force this typing, it will cause errors

rli: this is handled

PR #6134, opened 2025-12-02
fix(amazonq): fix incorrect single object being sent to the ChatParams

aseemxs:
## Types of changes

## Description

Fixes code selection not being detected when users select code in the editor and then type directly in the Amazon Q chat panel.

Root cause: The LSP server's ChatParams interface expects cursorState to be an array (CursorState[]), but JetBrains was sending it as a single object. The server code accesses cursorState?.[0] which returns undefined when given an object instead of an array, causing it to fall back to the default cursor state
(no selection).

Fix: Wrap cursorState in a list before sending to the LSP server.

https://github.com/user-attachments/assets/99bc4662-9c44-490c-a4dd-85dc4450065b

This issue affected users who:
1. Select code in the editor
2. Click on the chat panel and type a message directly (without using right-click context menu)

The chat would respond with "There's no highlighted code" even though code was selected.

## Checklist
- [ ] A short description of the change has been added to the [CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests) if the change is customer-facing in the IDE.

## License


rli (approved):
please fix the PR title

PR #6185, opened 2026-01-15
fix(build): Fix IDE flavor resolution for 2025.3 unified distribution

Zee2413:
## Types of changes

## Description

Starting with 2025.3, JetBrains unified the IntelliJ IDEA distribution - there is no longer a separate Community edition download. The `ideaIC-2025.3` artifact doesn't exist; only `ideaIU-2025.3` is available.

This PR fixes two related issues:

### 1. Subplugin dependency resolution fails for 2025.3

**Problem:** `toolkit-intellij-subplugin.gradle.kts` was trying to resolve `ideaIC-2025.3` which doesn't exist.

**Fix:** Added IC→IU mapping when version starts with `2025.3` and flavor is `IC`, so dependencies resolve against `ideaIU-2025.3` instead.

### 2. `runIdeVariant` property doesn't override default IDE flavor

**Problem:** Running with `-PrunIdeVariant=IU` still launched `ideaIC` instead of `ideaIU`. This is due to Gradle property precedence - `.set()` creates an immutable value that `.convention()` cannot override.

**Fix:**
- Changed `toolkit-intellij-plugin.gradle.kts` to use `.convention()` (sets overridable default)
- Changed `toolkit-publish-root-conventions.gradle.kts` to use `.set()` (actually overrides)

This complements the existing 2025.3 handling in `feadd1bfc` which addressed the root project but not subplugin dependencies or runtime IDE selection.

### Impact

Without these fixes:
- Builds targeting 2025.3 fail to resolve dependencies (ideaIC-2025.3 doesn't exist)
- Run configurations with `-PrunIdeVariant=IU` incorrectly launch IC instead of IU
- Features depending on 2025.3+ modules (e.g., Platform LSP) cannot be developed/tested because the build resolves against the wrong distribution

## Checklist

Note: No tests added as these are build system configuration changes. No changelog entry needed as this is not customer-facing.

## License

### DIFF
--- a/buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
+++ b/buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
@@ -5,6 +5,7 @@ import org.jetbrains.intellij.platform.gradle.IntelliJPlatformType
 import org.jetbrains.intellij.platform.gradle.TestFrameworkType
 import org.jetbrains.intellij.platform.gradle.tasks.PrepareSandboxTask
 import software.aws.toolkits.gradle.findFolders
+import software.aws.toolkits.gradle.intellij.IdeFlavor
 import software.aws.toolkits.gradle.intellij.IdeVersions
 import software.aws.toolkits.gradle.intellij.toolkitIntelliJ
 


buildSrc/src/main/kotlin/toolkit-intellij-subplugin.gradle.kts
```diff
@@ -91,7 +92,14 @@ dependencies {
 
         // annoying resolution issue that we don't want to bother fixing
         if (!project.name.contains("jetbrains-gateway")) {
-            val type = toolkitIntelliJ.ideFlavor.map { IntelliJPlatformType.fromCode(it.toString()) }
+            val type = toolkitIntelliJ.ideFlavor.map { flavor ->
+                // Starting with 2025.3, IntelliJ IDEA is unified (no separate Community edition)
+                if (version.get().startsWith("2025.3") && flavor == IdeFlavor.IC) {
```
manodnyab, line 97: This is jb gateway so we shouldnt be checking for these

Zee2413: Isn't gateway below in the else condition?

PR #6215, opened 2026-02-03
feat(cloudformation): add CloudFormation tool window, stacks tree view

Zee2413:
## Types of changes

## Description

Adds CloudFormation Stacks Explorer as a new tab in the AWS Toolkit tool window (requires IDE 2025.3+).

image

Features:
- View CloudFormation stacks with status indicators
- Browse change sets for each stack
- Pagination support with "load more" for large stack/change set lists
- Sign-in panel when no IAM credentials configured

Architecture:
- ToolkitToolWindowTab extension point for adding tabs to the Explorer tool window
- CfnClientService - centralized wrapper for LSP protocol calls
- StacksManager / ChangeSetsManager - manage stack and change set state with pagination

Key files:
- src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/ - tool window and tree nodes
- src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/stacks/ - state managers
- src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/CfnClientService.kt - LSP client wrapper

## Checklist
- [x] A short description of the change has been added to the CHANGELOG if the change is customer-facing in the IDE.

## License


plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/nodes/ChangeSetNodes.kt
```diff
@@ -0,0 +1,94 @@
...
+    override fun isAlwaysLeaf(): Boolean = true
+}
+
+internal class ChangeSetNode(
+    nodeProject: Project,
+    private val changeSetName: String,
+    private val status: String,
+) : AbstractTreeNode<String>(nodeProject, changeSetName) {
+
+    override fun update(presentation: PresentationData) {
```
satyakigh, line 87: These changes seem immutable, since the values are from constructor parameters, should it be in `init`?

plugins/core/resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
```diff
@@ -514,6 +514,15 @@ cloudformation.delete_stack.timeout=Failed to delete stack {0} in {1} seconds. V
 cloudformation.execute_change_set.failed=Failed to execute change set against {0}
 cloudformation.execute_change_set.success=Successfully executed change set against {0}
 cloudformation.execute_change_set.success.title=Successfully executed change set
+cloudformation.explorer.change_sets=Change Sets
+cloudformation.explorer.load_more=load more...
+cloudformation.explorer.load_more_stacks=Load More Stacks
+cloudformation.explorer.refresh=Refresh
```
chrisqm-dev: The message should be "Refresh Stacks", should also probably change the var name

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
```diff
@@ -0,0 +1,48 @@
+// Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cfnlsp.protocol
+
+import com.google.gson.annotations.SerializedName
+
+data class ListStacksParams(
```
chrisqm-dev: can make the data classes internal and also the typealias in StacksManager

chrisqm-dev (approved):

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/stacks/ChangeSetsManager.kt
```diff
@@ -0,0 +1,98 @@
...
+
+@Service(Service.Level.PROJECT)
+internal class ChangeSetsManager(private val project: Project) {
+    internal var clientServiceProvider: () -> CfnClientService = { CfnClientService.getInstance(project) }
+
+    private val stackChangeSets = ConcurrentHashMap<String, StackChangeSets>()
+    private val loadedStacks = ConcurrentHashMap.newKeySet<String>()
+    private val listeners = mutableListOf<() -> Unit>()
+
+    private data class StackChangeSets(
```
satyakigh: Can be moved out of the class (still in same file)

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/stacks/StacksManager.kt
```diff
@@ -0,0 +1,116 @@
...
+import software.aws.toolkits.jetbrains.services.cfnlsp.protocol.ListStacksParams
+import software.aws.toolkits.jetbrains.services.cfnlsp.protocol.StackSummary
+
+typealias StacksChangeListener = (List<StackSummary>) -> Unit
+
+@Service(Service.Level.PROJECT)
+internal class StacksManager(private val project: Project) : Disposable {
+    internal var clientServiceProvider: () -> CfnClientService = { CfnClientService.getInstance(project) }
+
+    @Volatile private var stacks: List<StackSummary> = emptyList()
```
satyakigh: Why do we need Volatile?

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/CfnClientService.kt
```diff
@@ -0,0 +1,73 @@
...
+            lsp.workspaceService.didChangeConfiguration(DidChangeConfigurationParams(emptyMap<String, Any>()))
+        }
+    }
+
+    /**
+     * Sends a request to the LSP server and returns a future with the result.
+     * Uses sendNotification to safely access the server instance, then calls the
+     * protocol method which returns its own CompletableFuture.
+     */
+    private fun <T> sendRequest(request: (CfnLspServerProtocol) -> CompletableFuture<T>): CompletableFuture<T?> {
```
satyakigh, line 49: Response can be `<T?>` shouldn't there always be data?

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractExplorerTreeToolWindow.kt
```diff
@@ -50,10 +50,13 @@ import javax.swing.tree.TreePath
 
 abstract class AbstractExplorerTreeToolWindow(
     treeStructure: AbstractTreeStructure,
+    initialTreeExpandDepth: Int = 2,
 ) : SimpleToolWindowPanel(true, true), DataProvider, Disposable {
     private val treeModel = StructureTreeModel(treeStructure, null, Invoker.forBackgroundPoolWithoutReadAction(this), this)
     private val tree = Tree(AsyncTreeModel(treeModel, true, this))
 
+    protected fun getTree(): Tree = tree
```
satyakigh: can't the `tree` variable itself be scoped to protected for gets?

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
```diff
@@ -0,0 +1,48 @@
+// Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cfnlsp.protocol
+
+import com.google.gson.annotations.SerializedName
```
satyakigh: gson?

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
```diff
@@ -0,0 +1,48 @@
...
+
+data class ListStacksParams(
+    val statusToExclude: List<String>? = null,
+    val loadMore: Boolean = false,
+)
+
+data class StackSummary(
+    @SerializedName("StackName")
+    val stackName: String? = null,
```
satyakigh: Why don't we just name it `StackName` so you don't need any of the annotations?

plugins/core/resources/resources/software/aws/toolkits/resources/MessagesBundle.properties
Zee2413: Good callout. I changed the prefix for the stacks specific messages to be `cloudformation.explorer.stacks`

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/nodes/ChangeSetNodes.kt
Zee2413: The `update()` method is called by the tree framework whenever the node needs to be rendered/refreshed. It's not a one-time initialization - it's called repeatedly. Moving to init would break the tree rendering.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
Zee2413: Thanks done

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/stacks/ChangeSetsManager.kt
Zee2413: Done

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/stacks/StacksManager.kt
Zee2413: Did some further investigation and it does not seem strictly required. Removed. 
The runInEdt call after writes and the CompletableFuture completion both establish memory barriers that guarantee visibility, making @Volatile redundant.

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/CfnClientService.kt
Zee2413: The nullable handles the no server available case or if the cast to CfnLspServerProtocol fails

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractExplorerTreeToolWindow.kt
Zee2413: Yup, updated.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
Zee2413: the `com.google.gson.annotations.SerializedName` import is not banned, only `com.google.gson.Gson` which recommends `jacksonObjectMapper`

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
Zee2413: It violates the Kotlin style guidelines for casing but I've updated it to remove the need for importing gson

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/protocol/StackProtocol.kt
Zee2413: removed since using PascalCase for field names

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/CfnClientService.kt
satyakigh: wouldn't there be an exception if there are no servers available?

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractExplorerTreeToolWindow.kt
satyakigh: I meant something like 
```
    protected myProperty: String = "initial value"
        private set
```

satyakigh: With this change it is mutable by all child classes instead of just "get"

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/CfnClientService.kt
Zee2413: I'm using `.firstOrNull()` for getting `lspServerProvider` instance above in line 24-27 so there would be no exception.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/core/explorer/AbstractExplorerTreeToolWindow.kt
Zee2413: You might not be looking at the most recent commit. I have it as:
```kotlin
    protected val tree = Tree(AsyncTreeModel(treeModel, true, this))
```
No need for `private set` since it's `val` not `var`.

PR #6262, opened 2026-02-23
feat(cloudformation): add semantic versioning and limit server versio…

Zee2413:
…n for compatibility

## Types of changes

## Description
- use semantic versioning to compare language server versions
- limit the language server version to compatible range
- 
## Checklist
 
## License

### DIFF
--- a/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/CfnLspInstaller.kt
+++ b/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/CfnLspInstaller.kt
@@ -145,24 +145,21 @@ internal class CfnLspInstaller(
     }
 
     /**
-     * Removes old versions, keeping the current version and one fallback
-     * (the highest compatible version below current).
+     * Removes old versions, keeping the current version and one compatible fallback.
      */
     private fun cleanupOldVersions(currentVersion: String) {
         if (!Files.exists(storageDir)) return
 
         try {
-            val currentSemVer = SemVer.parse(currentVersion) ?: return
-
             val dirs = Files.list(storageDir).use { stream ->
                 stream.filter { it.isDirectory() }.toList()
             }
 
-            // Find the best fallback: highest compatible version below current
+            // Keep the highest compatible version other than current as fallback
             val fallbackDir = dirs
                 .filter { it.fileName.toString() != currentVersion }
                 .mapNotNull { dir -> SemVer.parse(dir.fileName.toString())?.let { dir to it } }
-                .filter { (_, ver) -> versionRange.satisfiedBy(ver) && ver < currentSemVer }
+                .filter { (_, ver) -> versionRange.satisfiedBy(ver) }
                 .maxByOrNull { (_, ver) -> ver }
                 ?.first
 
--- a/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/GitHubManifestAdapter.kt
+++ b/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/GitHubManifestAdapter.kt
@@ -4,7 +4,6 @@
 package software.aws.toolkits.jetbrains.services.cfnlsp.server
 
 import com.fasterxml.jackson.databind.DeserializationFeature
-import com.fasterxml.jackson.databind.PropertyNamingStrategies
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.intellij.openapi.util.SystemInfo
@@ -18,9 +17,9 @@ import java.net.http.HttpClient
 import java.net.http.HttpRequest
 import java.net.http.HttpResponse
 
-// Release manifest types (primary source)
 internal data class ManifestVersion(
     val serverVersion: String,
+    val latest: Boolean = false,
     val isDelisted: Boolean = false,
     val targets: List<ManifestTarget>,
 )
@@ -39,20 +38,6 @@ internal data class ManifestContent(
     val bytes: Long,
 )
 
-// GitHub releases types (fallback)
-internal data class GitHubRelease(
-    val tagName: String,
-    val prerelease: Boolean,
-    val assets: List<GitHubAsset>,
-)
-
-internal data class GitHubAsset(
-    val name: String,
-    val browserDownloadUrl: String,
-    val size: Long,
-)
-
-// Unified result type
 internal data class ServerRelease(
     val version: String,
     val downloadUrl: String,
@@ -69,23 +54,13 @@ internal class GitHubManifestAdapter(
     private val httpClient = HttpClient.newBuilder().build()
     private val mapper = jacksonObjectMapper()
         .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-        .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE)
 
-    // Cached manifest JSON for offline fallback
     @Volatile
     private var cachedManifestJson: String? = null
 
-    fun getLatestRelease(): ServerRelease = try {
-        getFromManifest()
-    } catch (e: Exception) {
-        LOG.warn(e) { "Failed to fetch release manifest, falling back to GitHub Releases API" }
-        getFromGitHubReleases()
-    }
-
-    private fun getFromManifest(): ServerRelease {
+    fun getLatestRelease(): ServerRelease {
         val manifestJson = fetchManifestJson()
         cachedManifestJson = manifestJson
-
         return parseManifest(manifestJson)
     }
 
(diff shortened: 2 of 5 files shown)


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
```diff
@@ -0,0 +1,103 @@
...
+            val b = other.prerelease.getOrNull(i)
+            if (a == null) return -1 // fewer fields = lower precedence
+            if (b == null) return 1
+            val aNum = a.toIntOrNull()
+            val bNum = b.toIntOrNull()
+            val cmp = when {
+                aNum != null && bNum != null -> aNum.compareTo(bNum)
+                aNum != null -> -1 // numeric < string
+                bNum != null -> 1
+                else -> a.compareTo(b)
```
chrisqm-dev: nit: can we use more clear variable names here

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
```diff
@@ -0,0 +1,103 @@
...
+            val core = if (hyphenIdx >= 0) cleaned.substring(0, hyphenIdx) else cleaned
+            val prePart = if (hyphenIdx >= 0) cleaned.substring(hyphenIdx + 1) else null
+
+            val parts = core.split('.')
+            if (parts.size != 3) return null
+
+            val major = parts[0].toIntOrNull() ?: return null
+            val minor = parts[1].toIntOrNull() ?: return null
+            val patch = parts[2].toIntOrNull() ?: return null
+            val prerelease = prePart?.split('.')?.flatMap { it.split('-') } ?: emptyList()
```
chrisqm-dev: do we need to split on `.`?

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
```diff
@@ -0,0 +1,103 @@
+// Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package software.aws.toolkits.jetbrains.services.cfnlsp.server
+
+/**
+ * Minimal semver implementation for comparing language server versions.
+ * Handles formats: "1.4.0", "v1.4.0", "1.4.0-beta", "1.4.0-2030-alpha"
```
satyakigh: We don't need the -2030-alpha case, don't think we make those

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
```diff
@@ -0,0 +1,103 @@
...
+            val cleaned = version.removePrefix("v")
+            // Split "1.4.0-beta" into core="1.4.0" and pre="beta"
+            val hyphenIdx = cleaned.indexOf('-')
+            val core = if (hyphenIdx >= 0) cleaned.substring(0, hyphenIdx) else cleaned
+            val prePart = if (hyphenIdx >= 0) cleaned.substring(hyphenIdx + 1) else null
+
+            val parts = core.split('.')
+            if (parts.size != 3) return null
+
+            val major = parts[0].toIntOrNull() ?: return null
```
satyakigh, line 55: These should just throw, same as the parse function itself, if the version is not in a format we expect we don't want to continue

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Minimal semver implementation for comparing language server versions.
+ * Handles formats: "1.4.0", "v1.4.0", "1.4.0-beta", "1.4.0-2030-alpha"
+ */
+internal data class SemVer(
+    val major: Int,
+    val minor: Int,
+    val patch: Int,
+    val prerelease: List<String> = emptyList(),
```
satyakigh, line 14: this probably just be an enum of prod/beta/alpha instead of lists

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/GitHubManifestAdapter.kt
```diff
@@ -130,6 +128,15 @@ internal class GitHubManifestAdapter(
         )
     }
 
+    private fun latestCompatibleVersion(versions: List<ManifestVersion>): ManifestVersion =
+        versions
+            .filter { !it.isDelisted }
```
satyakigh, line 112: lets use the "latest" boolean flag and if that doesn't exist, fallback to getting the latest by compute

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/CfnLspInstaller.kt
```diff
@@ -132,12 +144,31 @@ internal class CfnLspInstaller(
...
         try {
-            Files.list(storageDir)
-                .filter { it.isDirectory() && it.fileName.toString() != currentVersion }
+            val currentSemVer = SemVer.parse(currentVersion) ?: return
+
+            val dirs = Files.list(storageDir).use { stream ->
+                stream.filter { it.isDirectory() }.toList()
+            }
+
+            // Find the best fallback: highest compatible version below current
```
satyakigh: Is this true? It should look for highest compatible version only

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
Zee2413: renamed using `this` and `other`

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
Zee2413: our versions don't. removed

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
Zee2413: removed as not needed by our versions

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
Zee2413: parse returning null is intentional because callers use it as a filter, not a validator. For example:

```kotlin
// In latestCompatibleVersion — skip unparseable versions, don't crash
.mapNotNull { v -> SemVer.parse(v.serverVersion)?.let { v to it } }

// In findCachedServer — directories might not be version strings
.mapNotNull { dir -> SemVer.parse(dir.fileName.toString())?.let { dir to it } }

// In cleanupOldVersions — same, skip non-version dirs
.mapNotNull { dir -> SemVer.parse(dir.fileName.toString())?.let { dir to it } }
```

If parse threw, every one of these would need a try/catch wrapper. The callers that need a guaranteed result already fail explicitly — `latestCompatibleVersion` throws "No compatible version found" if the filtered list is empty.

So the safety net is there — just at the right level. A single bad version string in the manifest doesn't crash the whole flow; it gets skipped, and if nothing remains, we get a clear error.

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/GitHubManifestAdapter.kt
Zee2413: add `isLatest` check first then comparison

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/SemVer.kt
Zee2413: prerelease isn't the environment — it's the semver prerelease segment parsed from the version string itself. When we parse "1.4.0-beta", the prerelease becomes ["beta"]. It's a List<String> because that's what the semver spec defines — the part after the hyphen is a series of identifiers used for comparison ordering.

We already have `CfnLspEnvironment` as an enum for prod/beta/alpha. That's what controls which array we read from the manifest. SemVer.prerelease is a completely different concept — it's just version string metadata needed for correct sorting (e.g., 1.4.0 > 1.4.0-beta).

plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/server/CfnLspInstaller.kt
Zee2413: Changed to keep the highest compatible version.

laileni-aws (approved):

PR #6293, opened 2026-03-09
fix(cloudformation): Start CloudFormation LSP upon toolkit activation

chrisqm-dev:
Wait for the LSP Server to be initialized before creating the tool window.

## Types of changes

## Description
This change starts the LSP Server upon toolkit activation. Previously, the LSP Server would lazy load based on if the CloudFormation panel was selected or if a yaml or json file was opened

## Checklist
 
## License

### DIFF
--- a/plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/CloudFormationToolWindow.kt
+++ b/plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/CloudFormationToolWindow.kt
@@ -9,7 +9,6 @@ import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.components.service
 import com.intellij.openapi.project.Project
-import com.intellij.platform.lsp.api.LspServerManager
 import software.aws.toolkits.jetbrains.ToolkitPlaces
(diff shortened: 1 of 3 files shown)


plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/explorer/CloudFormationToolWindow.kt
```diff
@@ -81,11 +84,33 @@ internal class CloudFormationToolWindow(private val project: Project) : Abstract
         }
     }
 
-    private fun ensureLspServerStarted() {
-        LspServerManager.getInstance(project).ensureServerStarted(
-            CfnLspServerSupportProvider::class.java,
-            CfnLspServerDescriptor.getInstance(project)
-        )
+    private fun ensureLspServerStartedAndLoadData() {
+        val lspManager = LspServerManager.getInstance(project)
```
satyakigh: what does this do/why is it required? Why is the ensure method required at all?

chrisqm-dev: LspServerManager.getInstance(project).ensureServerStarted() starts up the server. Since we are now launching the lsp server on toolkit startup (instead of when the cloudformation panel is selected), I removed this private method altogether.

PR #6356, opened 2026-04-10
fix(261): Handle compilation/test/detekt issues in 2026.1

branrin:
## Types of changes

## Description

Fixing 4 issues for 2026.1
1. `compileIntegrationTestKotlin`
    - `com.intellij.compiler.CompilerTestUtil` was removed from IntelliJ 2026.1's test framework. `JavaAwsConnectionExtensionIntegrationTest.kt `and `JavaLocalLambdaRunConfigurationIntegrationTest.kt` imported it. Moved the original files from `it/` → `it-242-253/` (for older IDE versions that still have `CompilerTestUtil`), and created new `it-261+/` versions that remove all `CompilerTestUtil` usage. In 2026.1, the external compiler is always enabled,  
  so `enableExternalCompiler()` / `disableExternalCompiler()` / `saveApplicationSettings()` calls are no longer needed.
2. `detektMain` for jetbrains-core
    - Detekt flagged `findSamBuildContents(module: Module)` in `Lambda.kt` as unused. This is a false positive — the function IS called by the overloaded `findSamBuildContents(project: Project)`, but detekt doesn't recognize the overloaded call chain. Added Suppress
3. `detektMain` for jetbrains-gateway
    - In 2026.1, `DialogBuilder.getOkAction()` returns `CustomizableAction` (an interface). The code casts this to `DialogBuilder.OkActionDescriptor` inside an `okAction.apply {}` block to call `getAction()`. While the cast works at runtime (because `OkActionDescriptor` extends `BuiltinAction` which implements `CustomizableAction`), detekt's newer type analysis correctly flags it as unsafe since `CustomizableAction` and `OkActionDescriptor` have no direct type relationship at compile time. Added Suppress
4. `JavaLambdaBuilderTest.mavenRootPomBuildDirectoryIsCorrect` failure — 
    - The error was `MavenServerManager.getInstance()` returning null during test teardown in `JavaTestUtils.kt`. In 2026.1, `MavenServerManager` is no longer registered as an application-level service in the test environment. Wrapped the `Disposer.dispose(MavenServerManager.getInstance())` call in a try-catch in `tst-253+/JavaTestUtils.kt` so the teardown doesn't crash when the service isn't available.

## Checklist
 
## License


eliseong (approved):

branrin:
Linux Unit failure is just flakiness, unrelated to these changes

PR #6365, opened 2026-04-20
fix(cloudformation): Make CloudFormation LSP download strategy consistent

satyakigh:
## Summary

These changes make the CloudFormation LSP server download, installation, caching, and cleanup strategy consistent across all client plugins and safe for concurrent multi-IDE usage.

Previously each plugin had its own ad-hoc approach to downloading and managing the LSP binary. Now all clients share the same design:

- **Atomic installs** — download to a PID-stamped temp directory, then rename into place
- **PID-based in-use tracking** — marker files prevent one IDE from deleting a version another is actively using
- **File-based manifest caching** — a plain `manifest.json` on disk replaces IDE-specific storage
- **Unified cache directory** — all plugins write to the same OS-standard cache path
- **Stale temp cleanup** — sweeps orphaned `.tmp.<pid>` directories from crashed processes
- **Local fallback resolution** — when both network and cached manifest fail, scans installed versions for a usable binary

---

## Motivation

Users commonly have both the standalone CloudFormation extension AND the AWS Toolkit installed in VS Code, or use both VS Code and JetBrains. Without a shared cache directory and in-use tracking, these plugins fight over the same LSP binary — downloading redundant copies, deleting each other's active versions, and corrupting installations mid-session.

The LSP download happens over the network from GitHub. In corporate environments with proxies, air-gapped networks, or intermittent connectivity, the download can fail at any point. The previous code had minimal resilience: no atomic writes (corruption on crash), no offline fallback (complete failure), and IDE-specific caching (no cross-plugin benefit).

**`CfnLspInstaller.kt`** — Major rewrite. Atomic download via tmp directory + `Files.move()`. New `findServerFileInDir()` handles nested extraction subdirectories. `cleanupOldVersions()` sweeps stale tmp dirs and checks in-use markers before deleting. File-based manifest cache replaces `PropertiesComponent`. New `tryDevBundlePath()` reads `CFN_LSP_DEV_BUNDLE` env var for dev mode. `cfnLspCacheRoot()` computes OS-standard cache paths. `determineEnvironment()` supports env var/system property override. `cleanupAfterResolve()` separates cleanup from resolution so the caller can write a marker first.

**`CfnLspServerSupportProvider.kt`** — Writes in-use marker after `getServerPath()`. Registers a JVM shutdown hook (`Runtime.getRuntime().addShutdownHook`) to remove markers on exit. Uses `ConcurrentHashMap` + `AtomicBoolean` for thread-safe hook registration. Calls `cleanupAfterResolve()` after the marker is written.

**`CfnLspExtensionConfig.kt`** — `EXTENSION_NAME` hardcoded to `"aws.toolkit.jetbrains"` instead of deriving from `AwsToolkit.TOOLKIT_PLUGIN_ID`.

**`GitHubManifestAdapter.kt`** — Adds null check for environment key in manifest JSON and validates the version list is non-empty.

**`HashUtils.kt`** *(new)* — Extracted `parseHashString()` and `computeHash()` from `CfnLspInstaller` into a standalone utility object.

**`InUseTracker.kt`** *(new)* — Kotlin implementation of PID-based marker tracking using `ProcessHandle`. Same API surface as the TypeScript versions: `writeMarker()`, `removeMarker()`, `isInUse()`, `cleanStaleMarkers()`.

## Types of changes

## License

### DIFF
--- a/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/CfnLspExtensionConfig.kt
+++ b/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/CfnLspExtensionConfig.kt
@@ -7,7 +7,7 @@ import software.aws.toolkit.jetbrains.AwsPlugin
 import software.aws.toolkit.jetbrains.AwsToolkit
 
 object CfnLspExtensionConfig {
-    const val EXTENSION_NAME: String = AwsToolkit.TOOLKIT_PLUGIN_ID
+    const val EXTENSION_NAME: String = "aws.toolkit.jetbrains"
     val EXTENSION_VERSION: String = AwsToolkit.PLUGINS_INFO[AwsPlugin.TOOLKIT]?.version ?: "unknown"
     const val ENCRYPTION_MODE = "JWT"
     const val TELEMETRY_NOTIFICATION_GROUP_ID = "aws.cfn.telemetry"
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Build/
 venv/
 guitest.log
 stale_outputs_checked
+.classpath
 # sam build directory
 .aws-sam/
 
@@ -36,3 +37,6 @@ ui-tests/bin
 #CodeWhispererChat
 node_modules
 !mynah-ui/package-lock.json
+
+# Eclipse
+**/org.eclipse.buildship.core.prefs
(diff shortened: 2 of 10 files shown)


plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/cfnlsp/CfnLspExtensionConfig.kt
```diff
@@ -7,7 +7,7 @@ import software.aws.toolkit.jetbrains.AwsPlugin
 import software.aws.toolkit.jetbrains.AwsToolkit
 
 object CfnLspExtensionConfig {
-    const val EXTENSION_NAME: String = AwsToolkit.TOOLKIT_PLUGIN_ID
+    const val EXTENSION_NAME: String = "aws.toolkit.jetbrains"
```
vzhan00, line 10: Just curious why switch to hardcoded value here?

vzhan00 (approved):

PR #6367, opened 2026-04-21
feat(cloudformation): force large templates and artifact to be upload…

Zee2413:
…ed to s3

## Types of changes

## Description
- enforce s3 upload when templates exceed size limit or contain artifact
- backwards compatible change with prior language server versions <=1.6.0

## Checklist
 
## License


plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/documents/CfnDocumentManager.kt
```diff
@@ -27,13 +29,23 @@ internal class CfnDocumentManager {
     fun getValidTemplates(): List<DocumentMetadata> =
         documents.filter { it.cfnType == "template" }
 
+    fun requiresS3Upload(uri: String): Boolean {
+        val doc = documents.find { it.uri == uri }
+        if (doc == null) {
+            LOG.warn { "Document metadata not found for URI: $uri. Assuming no S3 upload required may lead to deployment failure." }
```
satyakigh, line 35: What does this mean?

satyakigh (approved):

deepfuriya (approved):

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/documents/CfnDocumentManager.kt
chrisqm-dev: Can be reworded to: "Document metadata not found for URI: $uri, unable to determine if S3 upload is required. Not using S3 upload may lead to deployment failure."

chrisqm-dev (approved):

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/documents/CfnDocumentManager.kt
Zee2413: We don't allow deploying templates not found by language servers.

Added this warning log before returning false just for the s3 check part.

The size bytes will be available in metadata upon next language server release. This change is backwards compatible till then with the default.

yuriivv (approved):
Re-running checks after fixing Qodana failures

PR #6383, opened 2026-07-07
fix(cloudformation): show failure reason on resource import/clone

gemammercado:
## Problem

When importing or cloning CloudFormation resource state fails, the notification only shows a count (e.g. `Failed to import 1 resource`), with no indication of *why* it failed. This is not useful for debugging.

## Solution

Surface the per-resource error message the CloudFormation language server now returns (`failureReasons`) in the import/clone result notification, so users see the actual AWS error (e.g. `ResourceNotFoundException`, `AccessDenied`) for each failed identifier.

This is the JetBrains counterpart to the VS Code client change, and depends on the server-side change that adds `failureReasons`:
- VS Code client: aws/aws-toolkit-vscode#8632
- Language server: aws-cloudformation/cloudformation-languageserver#451

The search "not found" path already surfaces the server `error` in the JetBrains toolkit; this change adds the import/clone `failureReasons` half.

## Changes

- Add `failureReasons` (`Map<String, Map<String, String>>?`) to the `ResourceStateResult` protocol type (nullable/default for backward compatibility).
- `ResourceNotificationService`: add `formatFailureReasons` and append the reasons to the partial (warning) and full-failure (error) notifications via new `*.with_reasons` message keys.
- `ResourceStateService`: pass `result.failureReasons` through to the notification.
- New `*.with_reasons` keys in `MessagesBundle.properties` for import and clone.
- Unit tests: `ResourceNotificationServiceTest` (new) and a pass-through test in `ResourceStateServiceTest`.

## Testing

- Unit tests pass (`ResourceNotificationServiceTest`, `ResourceStateServiceTest`).
- Manually verified in a `runIde` (Ultimate) sandbox: importing deleted S3 buckets produces, e.g.
  `Failed to import 3 resources: [cfn-fail-test-a: ResourceNotFoundException: AWS::S3::Bucket Handler returned status FAILED: Bucket not found (...)], ...`

## License


plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
```diff
@@ -55,6 +63,16 @@ internal class ResourceNotificationService(private val project: Project) {
         }
     }
 
+    internal fun formatFailureReasons(failureReasons: Map<String, Map<String, String>>?): String {
```
Zee2413: should be private

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
```diff
@@ -55,6 +63,16 @@ internal class ResourceNotificationService(private val project: Project) {
...
+        if (failureReasons.isNullOrEmpty()) {
+            return ""
+        }
+        val reasons = failureReasons.values
+            .flatMap { it.entries }
+            .map { (id, reason) -> "[$id: $reason]" }
+        return if (reasons.isEmpty()) "" else ": ${reasons.joinToString(", ")}"
```
Zee2413: if there are many identifiers, does this show up in a user friendly way?

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
```diff
@@ -55,6 +63,16 @@ internal class ResourceNotificationService(private val project: Project) {
         }
     }
 
+    internal fun formatFailureReasons(failureReasons: Map<String, Map<String, String>>?): String {
+        if (failureReasons.isNullOrEmpty()) {
+            return ""
+        }
+        val reasons = failureReasons.values
```
chrisqm-dev: nit: would adding the `:` in the message bundle make the formatting logic less complex?

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
chrisqm-dev: Should have a hard limit on how many reasons are shown

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceStateService.kt
```diff
@@ -23,9 +24,14 @@ import software.aws.toolkits.resources.message
...
 ) {
+    @VisibleForTesting
     internal var clientServiceProvider: () -> CfnClientService = { CfnClientService.getInstance(project) }
+
+    @VisibleForTesting
     internal var editor = ResourceStateEditor.getInstance(project)
-    private val notificationService = ResourceNotificationService(project)
+
+    @VisibleForTesting
+    internal var notificationService = ResourceNotificationService(project)
```
Zee2413: Can these dependencies be injected via constructor? It would be easier for testing as well.
Why do some of these need to be mutable?

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
gemammercado: Jetbrains shows a scrollable pop-up failure message block, so long failures are not truncated like in VSCode. The size of the pop-up only fits ~2 exception messages so instead of scrolling I added a link to the message that opens a pop up log of all the error messages. Capped the shown reasons to 2 exceptions
Screenshot 2026-07-12 at 7 14 06 PM


I found this way of displaying the errors especially useful when there are various unique exceptions. For example, if the message contains three ResourceNotFoundExceptions and two AccessDeniedExceptions. With the hard limit, the user would only see the NotFoundExceptions and possibly assume the rest are of the same exception type. The expanded log shows all of them

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceNotificationService.kt
gemammercado: colon now lives in the 4 `*.with_reasons templates: Failed to import {0} resource{1}: {2}` (and the partial/clone variants with `: {3})`.
formatFailureReasons returns the bare reason list, no leading `": ":`

plugins/toolkit/jetbrains-core/src-253+/software/aws/toolkits/jetbrains/services/cfnlsp/resources/ResourceStateService.kt
gemammercado: They dont need to be, the var was there so that tests could swap in mocks after construction. I'll implement the injection via construction

Zee2413 (approved):

chrisqm-dev (approved):

afida23 (approved):

PR #6401, opened 2026-07-21
fix(262): verify against IntelliJ IDEA Ultimate for 2026.x

branrin:
## Types of changes

## Description
Fix plugin verification for 2026.x builds. Per comment, `Starting with 2025.3, IntelliJ IDEA is unified (no separate Community edition)` so we should only test against Ultimate and not Community

## Checklist
 
## License

### DIFF
--- a/buildSrc/src/main/kotlin/toolkit-publish-root-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/toolkit-publish-root-conventions.gradle.kts
@@ -26,7 +26,7 @@ intellijPlatform {
             // recommended() appears to resolve latest EAP for a product?
             // Starting with 2025.3, IntelliJ IDEA is unified (no separate Community edition)
             val version = toolkitIntelliJ.version().get()
-            if (version.startsWith("2025.3")) {
+            if (version.startsWith("2025.3") || version.startsWith("2026.")) {
                 create(IntelliJPlatformType.IntellijIdeaUltimate, version)
             } else {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/toolkit-publish-root-conventions.gradle.kts
```diff
@@ -26,7 +26,7 @@ intellijPlatform {
             // recommended() appears to resolve latest EAP for a product?
             // Starting with 2025.3, IntelliJ IDEA is unified (no separate Community edition)
             val version = toolkitIntelliJ.version().get()
-            if (version.startsWith("2025.3")) {
+            if (version.startsWith("2025.3") || version.startsWith("2026.")) {
```
TrevorM314, line 29: Should we refactor this check so we don't need to make the same change for future versions, such as 2027? i.e. 
```
create(IntelliJPlatformType.IntellijIdeaUltimate, version)
if (version earlier than 2025.3) {
    create(IntelliJPlatformType.IntellijIdeaCommunity, version)
}
```

TrevorM314 (approved):

buildSrc/src/main/kotlin/toolkit-publish-root-conventions.gradle.kts
branrin: Will refactor the logic so there is no condition once 2025.1 and 2025.2 have support removed

PR #6405, opened 2026-07-22
fix(262): Java 25 build tooling, JDK-per-profile CI selection, and bare-app test compatibility

branrin:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Description
Adds Java 25 build tooling and unit/integration **test-source compatibility** for IntelliJ 2026.2 (262), so the 2026.2 profile compiles and packages cleanly.

262 changes the platform test framework in two ways this PR addresses:
1. The Kotlin/JVM target for 2026.2 is Java 25 (class file v69), which requires newer instrumentation tooling.
2. Upstream **removed `com.intellij.testFramework.ApplicationExtension`** and now builds a "bare" test application that no longer loads the plugin's `plugin.xml`. Test sources that referenced the deleted class or relied on auto-registered `testServiceImplementation` services fail to compile / bootstrap.

**Build tooling (Java 25)**
- Bump `jacoco` → 0.8.14 and `test-retry` → 1.6.5 for Java 25 / class file v69 support.
- CI: select the profile-required JDK via `printJvmTarget` and pin `JAVA_HOME` for forked Gradle tasks (Linux buildspecs), so `test`/`instrumentCode` run on a JVM matching the compiled bytecode. This was already completed for Windows in https://github.com/aws/aws-toolkit-jetbrains/pull/6402

**262 bare-app test compatibility**
- Add a local `ApplicationExtension` shim (replacing the removed platform class, without the leak tracker that regresses older profiles) and migrate the affected unit/integration tests to it.
- Add `TestApplicationServices.kt` (`getOrRegisterApplicationService` / `getOrRegisterProjectService`) and make the shared `Mock*` fixtures register their service implementation if the platform didn't (register-if-absent). No-op on ≤2026.1; supplies the mock on 262.
- `FileChooser` fixture cleanup for a removed overload, and suppress a 262 `DEPRECATION_ERROR` in a Python integration test.

## Checklist
- [x] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [ ] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [ ] I have added metrics for my changes (if required)
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6406, opened 2026-07-22
ci(262): gate 2026.2 on buildPlugin/compile

branrin:
## Types of changes

## Description

Adjusts CI gating for the **2026.2** profile only: gate the build on the real Java 25 compile/package, and run the unit and integration tests for reporting **without** gating on them. Non-2026.2 profiles are unchanged.

### Why we are not simply making the tests pass

IntelliJ 2026.2 changed the test framework in a way that is fundamentally incompatible with how a large portion of our test suite is written, and the gap cannot be closed by a mechanical fix.

**1. The 262 test application no longer loads our plugin descriptor.**
Previously, the unit-test `Application` was initialized from our `plugin.xml`, so every `<applicationService>`, `<projectService>`, `<extensionPoint>`, and registry key it declares was registered and resolvable in tests — just like in a running IDE. In 2026.2, upstream removed `com.intellij.testFramework.ApplicationExtension` and now brings up a "bare" application that does **not** load the plugin descriptor. As a result, lookups that used to resolve now fail:
- `service<T>()` → `Cannot find service ...` (e.g. `SdkClientProvider`, `ToolkitClientManager`, `AwsResourceCache`, `ToolManager`, `ExecutableManager`, `ToolkitExperimentManager`, `CloudWatchLogWindow`, and ~10 others)
- extension-point lookups → `Missing extension point: ...` (e.g. `aws.toolkit.executable`, `aws.toolkit.experiment`, `aws.toolkit.credentialProviderFactory`, `aws.toolkit.lambda.runtimeGroup`, `aws.toolkit.sdk.clientCustomizer`, `aws.toolkit.toolWindowTab`)

This surfaces as ~885 failures across ~156 test classes spanning essentially the whole toolkit feature surface (Lambda, CloudFormation, DynamoDB, ECS, S3, SSO/IAM, etc.).

**2. There is no single choke point to fix.**
The failing lookups happen in **production** code paths under test (for example `AwsClientManager.sdkHttpClient()` → `AwsSdkClient.getInstance()` → `service<SdkClientProvider>()`), not only in test fixtures — so there is no small set of test helpers we can patch to cover them. #6405 already re-registers the services that *do* have a mock-fixture choke point (register-if-absent), which is why the core module compiles and its services resolve; but that approach does not extend to the many services and extension points that are resolved directly by product code. Extension points in particular cannot be registered through the test service-replacement APIs at all — they require the descriptor to actually load.

**3. A meaningful subset of tests cannot pass in the bare-app model regardless of effort.**
~48 of our test files rely on **language-plugin** fixtures (Java/Python/Go PSI — `LightJavaCodeInsightFixture`, `JavaSdk`, handler resolvers, line markers, completion, template indexing). In the 262 bare-app model the Java/Python/Go language plugins are not loaded into the test application, so these tests cannot be made to pass by re-registering our own services — the language runtime they depend on simply isn't present. Fully re-scaffolding the remainder is a large, high-risk effort (a sibling plugin that hit the identical platform change needed on the order of ~1,000 lines of test scaffolding and still could not green a comparable language-PSI subset), and that scaffolding runs on all profiles, risking regressions on the currently-green ones.

### What this PR does

Given the above, we gate 2026.2 on what actually validates the release — that the plugin **compiles and packages on Java 25** — and keep the tests running for visibility:

- `linuxTestsForToolkit.yml` — run `:check` non-gating; gate on `:plugin-toolkit:intellij-standalone:buildPlugin`.
- `windowsTestsForToolkit.yml` — run `:check` non-gating; add `buildPlugin` as the gate.
- `linuxIntegrationTests.yml` — gate on `integrationTestClasses` (the Java 25 compile of the integration-test sources); run `integrationTest` non-gating.

The 2026.2 unit/integration test reports are still published, so the failures stay visible and can be chipped away at incrementally (restoring the non-language services/EPs) without blocking the 2026.2 build in the meantime. All other profiles continue to gate on `:check` / the full test run exactly as before.

## Checklist

## License

### DIFF
--- a/buildspec/linuxIntegrationTests.yml
+++ b/buildspec/linuxIntegrationTests.yml
@@ -45,7 +45,13 @@ phases:
         export JAVA_HOME=/usr/lib/jvm/java-${REQUIRED_JDK}-amazon-corretto.x86_64
         echo "Using JAVA_HOME=$JAVA_HOME for profile $ALTERNATIVE_IDE_PROFILE_NAME"
         if [ "$ALTERNATIVE_IDE_PROFILE_NAME" = "2026.2" ]; then
-          DISPLAY=:22 ./gradlew -PideProfileName=$ALTERNATIVE_IDE_PROFILE_NAME integrationTest coverageReport --no-parallel --info --console plain
+          # 2026.2 runs on the bare-app test model (upstream removed ApplicationExtension in 262), so the test
+          # sandbox no longer loads our plugin.xml: service/extension-point/PSI-dependent integration tests
+          # cannot register in-sandbox and cannot pass yet. Gate 2026.2 on the real Java-25 compile of the
+          # integration-test sources, then run them for reporting but do not gate on the run (matches the
+          # unit-test buildspecs). Other profiles keep gating on the run.
+          DISPLAY=:22 ./gradlew -PideProfileName=$ALTERNATIVE_IDE_PROFILE_NAME integrationTestClasses --no-parallel --info --console plain || exit 1
+          DISPLAY=:22 ./gradlew -PideProfileName=$ALTERNATIVE_IDE_PROFILE_NAME integrationTest coverageReport --no-parallel --info --console plain --continue || echo "2026.2 integration tests reported failures (non-gating; see integ-test report)"
         else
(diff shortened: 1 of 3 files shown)


branrin:
The Linux Integration failures on this PR are not related to these changes. Those tests have been failing for all versions since April 24

TrevorM314 (approved):

PR #6410, opened 2026-07-23
test: fix integrations tests, DiskCache permission issues

branrin:
## Types of changes

## Description
Fixes CI test failures that surface in the Linux integration-test report (Ex: https://github.com/aws/aws-toolkit-jetbrains/pull/6408). Both are test-only fixes; no production behavior changes.

**1. `DiskCacheTest` POSIX-permission failures (3 tests)**

The integration job runs Gradle as **root**, whereas the unit job runs as a non-root `codebuild-user`. These three tests validate the self-healing permission logic in `PathUtils` (`tryFileOp`/`tryDirOp` → `tryFixPerms`), which only runs when a file/dir operation throws `AccessDeniedException`. Root bypasses POSIX permission bits, so:
- Reading/writing a `0`-permission file and traversing a `r-xr-xr-x` dir succeed → no `AccessDeniedException` → the perm-fix path under test never fires → permissions stay as-set (`---------` / `r-xr-xr-x` instead of the expected healed `rw-------` / `rwxr-xr-x`).
- Even if an ADE were thrown, `tryFixPerms` deliberately refuses to `chmod` root-owned files.

These are inherently non-root tests. Added `@DisabledIfSystemProperty(named = "user.name", matches = "root")` alongside the existing `@DisabledOnOs(OS.WINDOWS)` guard, mirroring the `ownership.name != "root"` invariant the production code itself relies on. They continue to run (and pass) in the non-root unit job, so coverage is preserved.

**2. `ToolkitTelemetryOTelSpanProcessorTest` "UncaughtExceptionsBeforeTest"**

A misattributed, pre-existing flake. The real cause is in the suppressed exception: `UpdaterTest` constructs `Updater` instances (each holds a 10 ms self-rescheduling `Alarm`) and never disposes them. After the test ends, a leaked alarm callback fires `describeStackResources` against the now-cleared Mockito inline mock (`Mock accessed after inline mocks were cleared`), which is logged as an error and attributed to whichever test JUnit starts next. Fixed by tracking the created `Updater`s and disposing them in `@After` — the `Alarm` is a child disposable, so disposal cancels its pending requests.

### Verification
- Both affected test source sets compile for 2025.3.
- `DiskCacheTest` runs (not skipped) and passes as a non-root user locally, confirming the root guard only disables under root.

## Checklist
## License

### DIFF
--- a/plugins/core/jetbrains-community/tst/software/aws/toolkit/jetbrains/core/credentials/sso/DiskCacheTest.kt
+++ b/plugins/core/jetbrains-community/tst/software/aws/toolkit/jetbrains/core/credentials/sso/DiskCacheTest.kt
@@ -8,6 +8,7 @@ import com.intellij.openapi.util.io.NioFiles
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty
 import org.junit.jupiter.api.condition.DisabledOnOs
 import org.junit.jupiter.api.condition.OS
 import org.junit.jupiter.api.extension.ExtendWith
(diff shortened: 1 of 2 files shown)


branrin:
Linux Unit [2026-1] failure due to test env flakiness, not this change

drevan-aws (approved):

